PR #76, opened 2026-07-26
fix(config): restore embedded/tracked bubble preset order parity

milocaetano:
## What

Reorder the `3d spheres` preset block in the embedded `crates/app/config/bubbles.toml` to match the tracked `config/bubbles.toml` (after `consumption focus`). No values change — only block position.

## Why

Main CI is red: #72 (3d spheres) and #73 (dense tape btc) inserted their presets at different positions in the two files. Each PR was green on its branch, but the merged combination fails `the_embedded_fallback_matches_the_tracked_presets_file`, which asserts the parsed files are identical.

## Verification

- [x] `cargo test --workspace` — the sync test fails without this change, passes with it

## Arch-review

Clean, zero findings: config-only reorder restoring a test-asserted invariant; no code touched, no values changed.

### DIFF
--- a/crates/app/config/bubbles.toml
+++ b/crates/app/config/bubbles.toml
@@ -99,6 +99,36 @@ show_quantity_labels = true
 show_trade_count = true
 label_min_radius = 14.0
 
+# For studying who is eating whom: the bubble shrinks and the consumption mark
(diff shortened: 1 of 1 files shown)


milocaetano:
Superseded by #75, which restored preset-file parity on main first.

PR #77, opened 2026-07-26
feat(app): implement the UI design model shell (phases 1-5)

milocaetano:
## Summary

Implements `docs/ux/ui-design-model.md` migration phases 1–5 — the new visual shell. UI-only: the engine, feeds and every hot path are untouched.

- **Design tokens + theme** (`theme.rs`): the §4 palette as named constants, applied to egui's widgets once at startup. Amber stays reserved for provenance honesty. Parity tests pin `style.rs` defaults to the tokens.
- **Icon system** (`widgets.rs`, phase 3): Phosphor icon font; one four-state icon button (idle / hover / active-tinted / disabled at 40% opacity that explains itself on hover). Emoji leave the chrome.
- **Status bar** (`statusbar.rs`, phase 1): one 28 px line — provenance (state dot green live / amber replay / red stalled, venue, symbol, lag), content (bar spec, `240+61 bars` split, `side: inferred (tick rule)` honesty label for MT5/replay), machinery (trades, fps + frame time gated by View → perf readings, timezone picker). Replaces the perf overlay, the floating timezone pill and the header painted over candles. The transport restyles into a single 30 px row directly above it.
- **Toolbar** (`toolbar.rs`, phase 2): SOURCE · BARS · HISTORY left, LAYERS · LOOK · PANELS right. Heatmap/bubbles become icon toggles (right-click opens the dock tab); history is a `+ older ▾` split button with the page size in its menu. The row never wraps: `collapse_plan` folds groups into `⋯` in the §6 order (unit-tested for order and monotonicity). Capability gating unchanged.
- **Dock** (`dock.rs`, phase 4): tabbed right dock (L2 / Bubbles / Session), collapsible to a 36 px strip, width remembered per tab, `Ctrl+B`. Panel bodies migrated as-is; both left `SidePanel`s retired. Opening a tab never toggles a layer.
- **Tool rail** (`toolrail.rs`, phase 5): Pointer / Crosshair as exclusive tools (`Esc`/`1`, `2`); the hover crosshair is now a mode.
- Menus per §10 (View: panels, dock tabs, perf readings, timezone; Tools: appearance); zone sizes per §5 (menu 28, toolbar 44, status 28, transport 30, time axis 24, price axis 64).
- `config.rs` gains `side_note()` so the status bar can disclose tick-rule side inference — the one place a provider maps to a provenance statement.

## Verification

- `cargo fmt --check`, `clippy -D warnings`, `build`, `test --workspace` all green (258 app tests + workspace).
- Smoke-run on Windows: connects to Binance, opens on BTCUSDT with the "dense tape btc" preset, streams live trades, no panics.
- arch-review run over the diff; all findings fixed in-branch (off-screen layout tests for every new chrome surface, named constants on the status bar, documented action convention). Phase 6 (indicator surfaces) deliberately deferred — it rides the indicator plan per §12.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -104,7 +104,7 @@ dependencies = [
  "objc2-foundation 0.3.2",
  "parking_lot",
  "percent-encoding",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
  "x11rb",
 ]
 
@@ -764,6 +764,15 @@ dependencies = [
  "epaint",
  "log",
  "nohash-hasher",
+]
+
+[[package]]
+name = "egui-phosphor"
+version = "0.7.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "74744068685d20e004d1a4326c48403b6759e325c7fce8c5eeeab93f0a9f67ca"
+dependencies = [
+ "egui",
 ]
 
 [[package]]
@@ -2452,6 +2461,7 @@ name = "quantick-app"
 version = "0.1.0"
 dependencies = [
  "eframe",
+ "egui-phosphor",
  "quantick-engine",
  "quantick-feed-binance",
  "quantick-feed-mt5",
@@ -2915,7 +2925,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.4.15",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
--- a/crates/app/Cargo.toml
+++ b/crates/app/Cargo.toml
@@ -26,6 +26,7 @@ eframe = { version = "0.29", default-features = false, features = [
     "x11",
     "default_fonts",
 ] }
+egui-phosphor = "0.7"
 
 [target.'cfg(windows)'.build-dependencies]
 winresource = "0.1"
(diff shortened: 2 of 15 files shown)


milocaetano:
### Code review

Found 2 issues (both already fixed in-branch by bf1faeb):

1. Replay silently relabeled the session: `draw_toolbar` called `ensure_symbol_valid()` unconditionally every frame, so replaying an instrument the selected live feed does not offer (e.g. a WINJ26/B3 recording while the Binance feed is selected) snapped `self.symbol` back to the feed's first symbol — the status bar and every tracing event named the wrong instrument for the whole session (CLAUDE.md says "Data honesty: inferred or incomplete data is labeled as such, never silently patched")

https://github.com/milocaetano/quantick/blob/38ac23c5c8994398ba1dd3bf6419297688e3e2c8/crates/app/src/app.rs#L408-L412

2. The status bar's perf cell dropped the CPU-per-frame reading that commit 5e423c5 added after a real incident ("real CPU frame time in HUD... separating being slow from waiting for the display"); only vsync-bound fps + mean frame time reached the screen, hiding render cost behind the display wait

https://github.com/milocaetano/quantick/blob/38ac23c5c8994398ba1dd3bf6419297688e3e2c8/crates/app/src/statusbar.rs#L251-L259

Fixes in https://github.com/milocaetano/quantick/commit/bf1faeb6b83f28785fcf6025fd826fb4568ef663: the symbol snap now skips replay (with a regression test that plays a real parsed session through the toolbar frame), and the perf cell reads `fps · ms · cpu ms`.

PR #84, opened 2026-07-27
feat(app): compact live tail — fixed-width footprint column for the forming bar

milocaetano:
## What

Live Edge phase 6 (the structural cleanup): the forming bar's order flow no longer stretches on a wall-clock axis. It renders as a **fixed-width live column** — per-price-bucket footprint bubbles (sells on the left rail, buys on the right, accumulated quantity on the bubbles' own √-area rule and reference) over a dimmed **book-now heat column** on the heatmap's ramp, quantization and reference. History clamps at the "now" divider and never re-compresses. The Live Strip is untouched: it stays the "now" on the price axis; the column is the forming bar's record.

Proposal: [Live Edge — UX do bar em formação](https://claude.ai/code/artifact/7c24c517-c05a-4a42-b9b9-77b55fa4c8d2), section "O live tail compacto, lado a lado".

## Resolved open decisions

- **Fixed width**: `LIVE_COLUMN_SPAN = 1.5` candle-widths — the bar's own 1-wide slot plus a half-width overhang right (the proposal mock's "~1,5 candle"). Rails at `LIVE_COLUMN_RAIL_OFFSET = 0.375` candle-widths from the candle centre.
- **Time inside the column**: nothing. No internal time axis, no progress indicator — intra-bar temporal position is the information the proposal accepts losing.
- **Live-region heat**: one column of current-book cells per bucket (`live_strip::depth_rows` on the published ladder), same ramp/reference as the heatmap, dimmed by `LIVE_COLUMN_HEAT_ALPHA = 0.4`. Current state only — no time lanes, so cells recolour in place and never slide.
- **Timeline mode**: removed outright rather than kept as a toggle (the goal overrode the proposal's keep-for-one-version suggestion). `live_span_for`, the `ProjectedLayout::live_span` stretch math and the app's live-span block are deleted; no dual mode behind a flag.
- **`live_tail_hold` dies**: it existed to decay the variable (up to 18-slot) reservation monotonically. The reservation is now the constant `LIVE_TAIL_SLOTS = 1.0`, so the following right edge advances exactly one slot per close and can never step right — same invariant, no bookkeeping. `last_total_bars` and the published `live_end_ms` mirror died with it.

## One aggregation path

The footprint reads `live_strip::aggression_rows` and the column heat reads `live_strip::depth_rows` — the exact functions the Live Strip already draws from, fed by the projection's clusters and the published ladder. No second copy of bucketing anywhere. History layers skip forming-bar primitives by the same open-time cut, so no mark draws twice and none is dropped; they appear with the bar's history at close.

## Acceptance criteria → evidence

- **A new trade never moves or shrinks drawn marks**: accumulation only grows a bubble in place (per-bucket sums, fixed rails); the layout clamp pins open runs at the divider. Tests: `history_clamps_at_the_live_boundary`, `the_live_column_is_fixed_width_with_sells_left_and_buys_right`.
- **No mark unreadable from time allocation**: the time axis inside the live region is gone; bubbles sit on two rails at the bubbles' normal radii (no 2 px horizontal fanning).
- **Jump-free close**: the column's slot becomes the closed candle's slot (the candle never moves — it always sat at its permanent slot centre); the constant reservation advances the right edge exactly one slot per close. Test: `a_bar_close_advances_the_right_edge_by_exactly_one_slot`. Between close and the next trade there is deliberately no column (no forming bar exists) — an honest beat, not a jump.
- **Layer honesty**: the footprint obeys the aggression family switch and per-side toggles like any bubble. Test: `the_footprint_honors_the_aggression_layer_and_side_toggles`.
- **Live validation**: Binance/BTCUSDT live with the shipped "dense tape btc" preset (Binance-only config via `QUANTICK_CONFIG` in the session scratchpad; root `quantick.toml` untouched), `QUANTICK_BOOK_AUTOSTART=1`, `QUANTICK_LIVE_STRIP_AUTOSTART=1`, plus the new `QUANTICK_BUBBLES_AUTOSTART=1`. 60 fps / 16.7 ms frames with book live. Screenshot close-ups confirm the geometry at max zoom: history ends sharply at the divider, column spans exactly 1.5 candle-widths, sell rail and buy rail land at ±0.375 candle-widths, strip intact.

## Arch-review

Run over `git diff main...HEAD`; no Blockers. One Should-fix (footprint gating untested) fixed in-branch. Two Consider findings deliberately deferred:

- `depth_rows`/`aggression_rows` are each computed twice per frame (strip + column), bounded by the ladder clip (≤128/side) and the aggression cap (700 default). Flat at 60 fps in validation; hoisting to one shared computation is signature plumbing through `draw_chart` best done if the cap is ever raised.
- The two column draw fns take 9 positional params; a params struct would be nicer. Zero runtime consequence.

## Verification

`cargo fmt --all -- --check` ✓ · `cargo clippy --workspace --all-targets -- -D warnings` ✓ · `cargo build --workspace` ✓ · `cargo test --workspace` ✓ (293 app tests, all crates green)

https://claude.ai/code/session_01PkY8ocUoB2TszfTErAqNVt

### DIFF
--- a/crates/app/src/app.rs
+++ b/crates/app/src/app.rs
@@ -303,6 +303,11 @@ impl QuantickApp {
         if std::env::var("QUANTICK_LIVE_STRIP_AUTOSTART").is_ok_and(|value| value == "1") {
             app.live_strip_visible = true;
         }
+        // Same convenience for the aggression layer (bubbles + the live
+        // column's footprint). Same code path as the toolbar toggle.
+        if std::env::var("QUANTICK_BUBBLES_AUTOSTART").is_ok_and(|value| value == "1") {
+            app.orderflow.set_bubbles_enabled(true);
+        }
         // Same convenience for Market Replay: open the folder named by
         // QUANTICK_REPLAY_DIR and play its first session. One env var, the same
         // code path a click takes, so a scripted run and a person get the same
--- a/crates/app/src/app.rs
+++ b/crates/app/src/app.rs
@@ -181,12 +181,6 @@ pub struct QuantickApp {
 
     // Pan/zoom navigation over the bar series.
     viewport: Viewport,
-    // Bar-slots reserved past the newest bar for the live region. Persisted so
-    // a closing bar frees exactly one slot: the chart never steps right — the
-    // freed space stays an empty gap the next bar's live region refills.
-    live_tail_hold: f32,
-    // Bar count seen last frame, to detect closes for the hold above.
-    last_total_bars: usize,
     // Manual price-axis pan/zoom (auto-fit until the user drags vertically).
     price_view: PriceView,
     // Last frame's auto-fit price range and chart height, for pixel↔price maths
@@ -280,8 +274,6 @@ impl QuantickApp {
             imbalance_target,
             viewport: Viewport::new(),
             price_view: PriceView::new(),
-            live_tail_hold: 0.0,
-            last_total_bars: 0,
             last_auto_range: None,
             last_chart_height: 1.0,
             hover_pos: None,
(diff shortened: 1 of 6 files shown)


milocaetano:
Closed unmerged after live validation by the owner: the wall-clock motion of the forming bar's bubbles turned out to be the felt signal that the flow is alive — the fixed column, while geometrically stable, made the chart read as idle. Keeping main's timeline behaviour. Salvaged into a follow-up branch: the dark separator ring, QUANTICK_BUBBLES_AUTOSTART, and a histogram-only Live Strip (depth background removed as redundant with the heatmap edge). This branch stays as the reference implementation should a compact-column toggle ever be wanted.

PR #86, opened 2026-07-27
feat(app): tell the user what the MetaTrader bridge is doing

milocaetano:
Running the B3 mini index with the book heatmap already worked — selecting the
MetaTrader feed starts the Python bridge on its own. What did not work was
*finding out why it hadn't*: terminal closed, package missing, contract not in
Market Watch, server clock unmeasurable were all reported only as a JSON line
on stderr, behind a chart that showed an empty canvas and a faint "connecting"
dot. Honest about *that* it was not connected, silent about *why*.

## What changed

- **`quantick-feed-mt5::bridge_log`** — translates a bridge's stderr vocabulary
  into a headline plus the one next step. It sits beside the protocol it reads,
  so renaming an `event_code` cannot leave a stale copy of these strings next
  to the UI.
- **`app::feed::mt5_bridge`** — owns launching and supervising, out of the feed
  module. Reads the bridge's stderr (every line reaches the log verbatim,
  recognized or not — a Python traceback is exactly when hiding the unknown
  hurts), resolves a relative script path against the executable's folder as
  well as the working directory, and falls back from `python` to the Windows
  `py` launcher for the shipped default only; a configured `bridge_command` is
  an instruction, not a guess.
- **`app::notice_card`** — draws the result over the chart under one rule:
  never cover a working chart. Progress shows only while there is nothing to
  look at; anything needing the user always shows, with a **Try again** button
  so the way back is a click rather than a restart of quantick.

Notices travel on their own channel rather than as a `FeedEvent`: connection
trouble is not market data, and a blocked feed has to speak while the trade
channel is silent — which is exactly when it matters. Feeds with nothing to
report (Binance, replay) hand over a channel closed at birth.

## Found by arch-review, fixed here

`bridge_connected` only ever went to `true`. The supervisor therefore returned
the first time it saw a connected bridge, so a terminal restarting mid-session
was never picked back up — while the card promised "reconnecting". The flag now
means *is one feeding us right now*, the supervisor watches instead of
returning (one relaxed atomic load every 2 s), and its budget is five
**consecutive** failures, so a wrong contract still stops and says so.

## Verified live — B3 WINQ26 through the XP terminal, 2026-07-27

- Bars and the book heatmap stream with the app launched from **outside** the
  repository (`-WorkingDirectory C:\Users\Camillo`): 60 fps, 10+10 DOM levels,
  coverage labelled `Limited`, `side: inferred (tick rule)`.
- A non-existent contract (`WINZ99`) produces the amber card — "MetaTrader does
  not list WINZ99" + "Add the contract to Market Watch…" + Try again — instead
  of a blank chart.
- Killing the bridge mid-session: relaunched 5 s later, streaming 1 s after
  that, with no user action.
- Closing the window leaves no orphan bridge process.

Four checks green: `fmt --check`, `clippy -D warnings`, `build`, `test`
(24 test binaries, 0 failures). `arch-review` run over `git diff main...HEAD`;
its one Blocker and two Should-fix findings are resolved in 94573be. The
remaining Consider (two short `String` allocations per frame while the card is
visible) is accepted and documented: the card only draws when nothing else is
being rendered, measured at 59-60 fps with it on screen.

### DIFF
--- a/bridge/mt5/README.md
+++ b/bridge/mt5/README.md
@@ -21,6 +21,22 @@ chart, is used instead of being fought. All it needs is the official package
 `bridge_autostart = false`, or point `bridge_command` somewhere else, under
 `[metatrader]` in your config.
 
+Two things make that work outside a developer's checkout:
+
+- **The script is found, not assumed.** A relative `bridge_command` path is
+  resolved against the working directory *and* against the folder holding the
+  quantick executable and its ancestors, so a build launched from a shortcut
+  still finds the `bridge/` folder shipped beside it.
+- **`python` is a first guess, not a requirement.** With no `python` on PATH,
+  the Windows `py` launcher is tried next. A `bridge_command` you configured
+  yourself is never second-guessed — it is an instruction, not a hint.
+
+**You should not have to read this file to fix a broken setup.** Whatever stops
+the bridge is reported on the chart, with the one next step: terminal closed,
+package missing, contract not in Market Watch, server clock unmeasurable, no
+Depth of Market for that symbol. The log keeps the full detail; the chart
+carries the part you have to act on.
+
 To run it by hand anyway — to see its log lines up close, or to feed a quantick
 that is not yours:
 
--- a/crates/app/src/app.rs
+++ b/crates/app/src/app.rs
@@ -2019,6 +2019,73 @@ mod tests {
         (app, evt_tx, cmd_rx, book_tx)
     }
 
+    /// The same app, plus the notice sender its feed would hold. The other
+    /// ends come back so the caller keeps the channels open, exactly as a live
+    /// feed thread would.
+    #[allow(clippy::type_complexity)]
+    fn test_app_with_notices() -> (
+        QuantickApp,
+        mpsc::Sender<FeedNotice>,
+        (mpsc::Sender<FeedEvent>, mpsc::Sender<DepthEvent>),
+    ) {
+        let (evt_tx, evt_rx) = mpsc::channel(64);
+        let (book_tx, book_rx) = mpsc::channel(64);
+        let (cmd_tx, _cmd_rx) = mpsc::channel(16);
+        let (notice_tx, notice_rx) = mpsc::channel(8);
+        let app = QuantickApp::new(
+            test_config(),
+            "binance",
+            "TESTUSDT",
+            BarSpec::Tick(50),
+            FeedHandle {
+                events: evt_rx,
+                book_events: book_rx,
+                notices: notice_rx,
+                commands: cmd_tx,
+                replay: None,
+            },
+        );
+        (app, notice_tx, (evt_tx, book_tx))
+    }
+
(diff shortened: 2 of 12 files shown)


milocaetano:
### Code review

Found 6 issues, all fixed in 632cf9a.

1. A survivable warning excused every later crash. `explained` was set by any Attention line, but `BRIDGE_BOOK_SUBSCRIBE_FAILED` is Attention while the bridge keeps streaming — so a symbol with no DOM silenced the "bridge stopped" notice for the rest of the session. Fixed by giving `BridgeReport` an `ends_session` flag that mirrors which codes raise `BridgeExit` in `quantick_bridge.py`.

https://github.com/milocaetano/quantick/blob/94573befe66e675e37b30b58ea0ae7e10bc0ab61/crates/app/src/feed/mt5_bridge.rs#L362-L372

2. Card text wrapped to 420 px while the card itself is clamped to the chart, so a long instruction spilled past its own border on a narrow chart. The existing test used strings too short to wrap.

https://github.com/milocaetano/quantick/blob/94573befe66e675e37b30b58ea0ae7e10bc0ab61/crates/app/src/notice_card.rs#L102-L104

3. `AsyncBufReadExt::lines` buffers an unterminated line without limit. PR #59 removed exactly this from the socket path and added `BoundedLineReader`; the new stderr reader reintroduced it. `BoundedLineReader` is now generic over its source and shared by both.

https://github.com/milocaetano/quantick/blob/94573befe66e675e37b30b58ea0ae7e10bc0ab61/crates/app/src/feed/mt5_bridge.rs#L364-L367

4. `while let Ok(Some(line))` ended the read loop on any error without logging, contradicting this module's own promise that every line reaches the log. Over-long and non-UTF-8 lines are now reported and skipped, and reading continues.

https://github.com/milocaetano/quantick/blob/94573befe66e675e37b30b58ea0ae7e10bc0ab61/crates/app/src/feed/mt5_bridge.rs#L365-L368

5. `AUTOSTART_GRACE` ran once, before the loop. Since this PR relaunches after a lost session, the supervisor could race an Expert Advisor's own 5 s reconnect (`InpRetrySeconds`) that the grace exists to avoid. Losing a session now re-arms it, and the connection is re-checked after waiting.

https://github.com/milocaetano/quantick/blob/94573befe66e675e37b30b58ea0ae7e10bc0ab61/crates/app/src/feed/mt5_bridge.rs#L229-L232

6. The give-up message told the user to "switch feeds to try again" — the flow this PR replaced with the Try again button sitting directly under that text.

https://github.com/milocaetano/quantick/blob/94573befe66e675e37b30b58ea0ae7e10bc0ab61/crates/app/src/feed/mt5_bridge.rs#L404-L409

One finding was dropped as a false positive: `restart_feed` gating on `self.replay.is_some()` rather than `FeedCapabilities`. It is a feed-lifecycle guard, not a UI affordance, and copies `maybe_switch_feed` verbatim one screen above it.

PR #87, opened 2026-07-27
feat(app): record the book always, mark the pre-capture span with one line

milocaetano:
## Summary

Two changes to the L2 depth layer, both from the same complaint: the map's toggle was doing a job it should not have, and the "no book here" region was shouting over data that is perfectly real.

**Recording and drawing are now separate concerns.** `HeatmapConfig` gains `show_depth` (display) beside `enabled` (the recorder), and the toolbar switch drives the first only. The recorder starts with any depth-capable feed through `ensure_book_capture` and stops only when the market changes — a feed/symbol switch, or a replay taking the chart over. Hiding the map therefore no longer punches a `capture_disabled` hole into the recorded book: reopen it and the whole retained window is there.

Hiding stays free. With the map closed `any_layer_enabled()` is false, so the worker builds no projection at all; the published ladder, the live tail, the status badge and the sync overlay all follow visibility rather than capture.

**The pre-capture span drops its fill and diagonal hatch** and is marked by a single dashed boundary where the book begins, with its label hugging that line. Interior gaps keep a faint fill between both boundaries — a few pixels of untinted chart would read as "no resting liquidity" rather than "no data". `draw_gap_hatch` and the `gap_hatch` palette entry are gone.

No linked issue: this came from a direct request in a `/goal` session, not from the board.

## Verification loop

- [x] `cargo test --workspace` — 24 suites, 527 tests

## Notes for the reviewer

**Deliberate departure from "additive by default."** Capture now starts without being asked, which the repo rule says a new capability must not do. That is precisely the requested behaviour — a recorder that only runs while someone is watching is the bug being fixed. The map's own visibility default is unchanged (still hidden at startup), so no pixels move until the user asks. `QUANTICK_BOOK_AUTOSTART=1` now opens the map rather than starting capture; capture needs no flag.

**New cost, measured.** Live on BTCUSDT with the map hidden: `heatmap_projection_builds=0`, `heatmap_projection_ms=0.0`, frame CPU ~0.9 ms, versus 1.3–3.9 ms with the map open — so the drawing side really is free. What is new is the ingestion a hidden map now pays: ~10 depth updates/s and ~15 KB/s of history on BTCUSDT, on the book thread, bounded by the existing budget (`retention_ms` 30 min, `max_history_runs`, `max_history_bytes` 64 MiB). No new ceiling.

**Removed per-frame work.** `draw_gap_hatch` pushed one `line_segment` shape per 18 px of gap width every frame — roughly 60 shapes for a full-width leading span — plus a `rect_filled`. Both are gone for that span.

**Where to look hardest.** `gap_marks` in `orderflow_render.rs` is the whole visual decision as a pure function, with three tests. The case worth a second opinion is the leading span's left edge: when the bars do not fill the chart, that edge lands on the oldest bar rather than on the viewport edge, and it is deliberately *not* drawn — there is nothing on the far side of it to separate from. Caught by pixel-inspecting a live capture, not by a test I had written first.

**Test cover.** `hiding_the_heatmap_never_stops_the_recorder` (app) fails if the toggle goes back to driving capture; `a_hidden_map_keeps_recording_and_costs_nothing_to_draw` (engine) fails if the ladder, live tail or projection gate regress to capture-gating; `a_hidden_depth_map_projects_no_depth_primitives` (projection) fails if `depth_enabled` reverts to `config.enabled`; `a_source_without_depth_starts_no_recorder` guards a replay from spawning a recorder behind the chart. There is no painter harness in this file, so nothing asserts the hatch is *painted* nowhere — `GapMarks.fill == false` is the closest proxy the existing test style allows.

https://claude.ai/code/session_01CCkZphyrbJeaiWvTiP8CVu

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,8 +222,9 @@ evenly across each second for smoother playback and marks the session
 ## Optional L2 liquidity map
 
 The chart can capture Binance Spot level-2 order-book depth and render a
-Bookmap-inspired liquidity map. It is **disabled by default** and must be
-enabled from the chart controls. The **⚙ L2** button docks a settings panel to
+Bookmap-inspired liquidity map. Capture starts with any feed that can stream
+depth; the map itself is **hidden by default** and shown from the chart
+controls. The **⚙ L2** button docks a settings panel to
 the left of the chart — the chart keeps the remaining width instead of being
 covered — including a deterministic preview of every visual state, so themes
 and grouping can be tuned without waiting for a rare live-book event.
@@ -236,15 +237,23 @@ switch and its own docked panel:
 - **bubbles** (**⚙ bubbles**) — confirmed aggressive executions, built from the
   aggregate-trade stream the chart already consumes.
 
-Neither switch touches the other. Bubbles render with L2 capture off (and for
-providers without a depth pipeline at all), and stopping the book leaves the
-bubbles running. Turning L2 capture off hides the map without discarding
-retained L2 history.
+Neither switch touches the other. Bubbles render with the map hidden (and for
+providers without a depth pipeline at all), and hiding the book leaves the
+bubbles running.
+
+Both switches are display-only. Recording and drawing are separate concerns:
+the depth recorder runs from the moment a depth-capable feed starts, so hiding
+the map never punches a hole in the recorded book — reopen it and the whole
+retained window is there, without the gap that stopping capture would leave.
+While the map is hidden no projection is built and no depth geometry is drawn,
+so the only cost it carries is the recording itself, bounded by the same
+retention budget documented below. Capture stops only when the market changes:
+a feed or symbol switch, or a replay taking the chart over.
 
 The visualization follows a few data-honesty rules:
 
-- History begins at the first successfully synchronized live snapshot/update sequence. Binance does not provide historical L2 backfill through this feed, so candles before that point are marked as unavailable instead of being reconstructed.
-- Depth update IDs are checked continuously. A disconnect, sequence gap or resynchronization closes the current liquidity runs, marks the affected interval with subtle shading and dashed vertical boundaries, and starts again from a fresh snapshot. Stale book state is never stretched across a gap.
+- History begins at the first successfully synchronized live snapshot/update sequence. Binance does not provide historical L2 backfill through this feed, so candles before that point are marked as unavailable instead of being reconstructed. That stretch is marked by a single dashed line where the book begins, plus a label: it can span most of the chart, and tinting it would bury candles and bubbles that are perfectly real there.
+- Depth update IDs are checked continuously. A disconnect, sequence gap or resynchronization closes the current liquidity runs, marks the affected interval with a faint fill between dashed vertical boundaries, and starts again from a fresh snapshot. Stale book state is never stretched across a gap. Interior gaps keep their fill precisely because they are narrow — an untinted sliver would read as "no resting liquidity" rather than "no data".
 - Heatmap quantities are resting bid/ask amounts from the snapshot plus absolute depth updates, limited to the configured number of price levels on each side. Liquidity outside that coverage is unknown.
 - A displayed band records the bucket total observed when its run opened; changes smaller than ~10% merge into the open run instead of cutting a new one. This churn-merge tolerance is a disclosed granularity choice — larger moves, appearances and full removals always cut a new run at their exact value.
 - `aggTrade` bubbles are confirmed market aggression. Their area is quantity-proportional and nearby prints can be clustered without changing total volume.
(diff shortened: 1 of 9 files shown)


milocaetano:
## Review

### What it does

Splits one flag that was doing two jobs. `HeatmapConfig.enabled` meant *both* "the depth recorder is running" and "the map is drawn", so the toolbar toggle stopped the recording and left a `capture_disabled` gap that survived in history. The PR adds `show_depth` for display, keeps `enabled` for the recorder, and routes every consumer through `depth_visible() = enabled && show_depth`. Separately, the pre-capture region loses its tinted fill and diagonal hatch in favour of a single dashed boundary.

### Correctness

- **The toggle/frame handoff is clean.** `set_depth_visible` drops the view's local frame *and* sends `ApplyVisualConfig`; the worker applies it, `invalidate_projection()` kills the cache, and — when bubbles keep a layer alive — re-projects in the same batch before publishing. So there is no window where a stale depth-ful frame gets drawn after the map is hidden. With bubbles off, `apply_visual_config` clears `last_frame` outright.
- **Visibility survives a market change.** Neither `reset_for_symbol` (engine or view) touches `show_depth`, so a symbol switch preserves whether the user wanted the map open. Only `enabled` resets, which is what `ensure_book_capture` then restores.
- **`live_end_ms` moved from capture-gating to visibility-gating**, which drives the forming bar's live tail. Worth checking, and it holds: before this PR, hiding the map also stopped capture, so `live_end_ms` went `None` on exactly the same user action. Toggle behaviour is byte-for-byte what it was. Same reasoning for the status badge and the sync overlay.
- **`gap_label` matching on the `BEFORE_CAPTURE` const** is a constant pattern, not a binding, because the name is SCREAMING_SNAKE. Worth stating since the lowercase version of that line would silently become a catch-all.

### One issue found — fixed in e669be7

`request_book_capture` only commits state when its command is queued, and its backpressure branch logged `action = "retry_on_next_user_action"`. That advice died with this PR: the toolbar no longer starts capture, so a start command dropped by a full channel would leave the session **silently unrecorded** with no user action left to retry it. The three lifecycle calls do not cover it — they only fire when the market changes.

Low probability in practice (the command channel is freshly created with capacity 16 at each of the three call sites, so it is empty when they fire), but the failure mode is the exact thing this PR exists to prevent, and it fails quietly. Fixed by calling `ensure_book_capture` from the frame loop as a heartbeat and correcting the log action.

### Performance

- **Removed:** `draw_gap_hatch` pushed one `line_segment` shape per 18 px of gap width *every frame*, plus a `rect_filled` — roughly 60 extra shapes per frame for a full-width leading span. Gone.
- **Removed while hidden:** the projection (`any_layer_enabled()` false ⇒ the worker never calls `project()`), the ≤256-level per-batch ladder copy, and the frame publish. Measured on BTCUSDT: `heatmap_projection_builds=0`, `heatmap_projection_ms=0.0`, frame CPU ~0.9 ms against 1.3–3.9 ms open.
- **Added:** ingestion a hidden map now pays — ~10 depth updates/s and ~15 KB/s of history on BTCUSDT, on the book thread, under the pre-existing retention budget. Deliberate, and the whole point of the change.
- **Added (heartbeat):** one bool read per frame while recording; during a replay one further `Option` check, since `capabilities()` short-circuits on `replay.is_some()`.
- `gap_marks` is four float comparisons over the ~1–3 gaps a frame has.

### Conventions and determinism

Follows the repo's capability-gating rule — `ensure_book_capture` asks `FeedCapabilities::book_capture`, never which provider is playing, so a replay is excluded by the same mechanism that greys out the toolbar button. No `HashMap`, no clock, no randomness added; `show_depth` is one more config bit through a projection that stays a pure function of history plus config. Dependency direction untouched.

The change edits nine existing files and adds none. That is worth a moment: it is the *removal* of an implicit coupling, so every place that assumed "capture ⇒ drawn" had to be told which of the two it actually meant. `depth_visible()` is now the name of that decision, and a future display-gated layer docks by adding a `show_*` flag plus one term in `any_layer_enabled()` — the pattern `show_liquidity` and friends already follow.

### Test coverage

Each behavioural claim has a test that fails without the change: `hiding_the_heatmap_never_stops_the_recorder`, `a_hidden_map_keeps_recording_and_costs_nothing_to_draw`, `a_hidden_depth_map_projects_no_depth_primitives`, `a_source_without_depth_starts_no_recorder`, and three cases over `gap_marks`.

**Gap worth naming:** there is no painter harness in `orderflow_render.rs`, so nothing asserts the hatch is painted *nowhere* — `GapMarks.fill == false` is a proxy for the decision, not for the paint. Re-adding a `draw_gap_hatch` call would pass the suite. Also untested: the heartbeat's retry-after-backpressure path, which would need a saturated command channel to exercise.

### Security

Nothing reachable from here — no network, credential or parsing surface in the diff, and no feed arithmetic touched.

### Verdict

Approve. One real finding, fixed. CI green.

PR #88, opened 2026-07-27
fix(app): the live region follows the tape, not just the book

milocaetano:
## The bug

The forming bar's live region — the space to the right of the newest candle where real-time flow rolls — never opened on MetaTrader. Every print of the forming bar piled up inside the candle's single slot, which is exactly where a tape stops reading as a tape.

The cause is the clock behind it. `live_end_ms` was `latest_book_ms` gated on `depth_visible()`: a timestamp only a *depth* event advances, published only while the heatmap is on screen. A MetaTrader session may publish no Depth of Market at all, and the map is often hidden precisely so the bubbles can be read — either one alone left the chart reporting no present. Binance hid the flaw because its book never stops ticking.

## The fix

`LiquidityHistory::latest_flow_ms()` — the furthest of the book's clock and the trade stream's — and the tail gates on `any_layer_enabled()` rather than on the depth map specifically. Depth and trades are two windows onto the same clock; a venue that streams only one of them still has a present. The same edge feeds the projection's timeline, so a bubble lands on the x the tail reserved for it.

The trade side is monotonic on purpose: a late-delivered print must not pull the edge backwards and shrink the tail for a frame.

What deliberately did not change:

- With no layer visible the clock is still withheld — a recorder nobody watches pays for no pixels.
- Coverage gaps are still marked at the last timestamp the **book** was observed. Using the trade clock there would claim book knowledge nobody has.
- Capture, retention and the book's own state machine are untouched.

## Proof

Two tests, both verified to fail without the change:

- `orderflow::history::tests::the_live_edge_follows_trades_when_no_book_is_streaming` — the edge advances on trades alone, takes the furthest of the two streams when both run, and never regresses on a late print.
- `orderflow_engine::tests::trades_alone_advance_the_live_edge_and_spread_the_forming_bar` — the user's report made executable. Three prints at 1 000 / 3 000 / 5 000 ms in a forming bar with no book land at x = 0.0 / 0.5 / 1.0. Reverting the published clock drops `live_end_ms` to `None`; reverting the timeline clock alone leaves **1 of 3** bubbles on screen — the pile-up, measured.

`a_hidden_map_keeps_recording_and_costs_nothing_to_draw` still passes: hiding every layer still publishes no clock.

## Verification

- [x] `arch-review` — no Blocker, no Should-fix. Performance: one `max` and one `Option<i64>` write per trade on the ingest path (no allocation); the live region now projects on feeds that lacked it, bounded as before by `max_aggression_primitives`.

https://claude.ai/code/session_01V9Ao2QbeY3gcYPy7iqszv7

### DIFF
--- a/crates/app/src/app.rs
+++ b/crates/app/src/app.rs
@@ -1186,7 +1186,8 @@ impl QuantickApp {
             return;
         }
 
-        // Live region: while following a live book, the forming bar grows to
+        // Live region: while following the live edge — the newest timestamp any
+        // recorded stream has reached, book or trades — the forming bar grows to
         // the right on a FIXED time-per-slot scale (the previous bar's
         // duration spread over the full width). Linear growth means an event
         // keeps its exact screen position while the region widens — bubbles
@@ -1200,7 +1201,7 @@ impl QuantickApp {
                 .last()
                 .map(|b| (b.close_time - b.open_time).max(1))
                 .unwrap_or(1_000);
-            // Up to ~55% of the chart for the live book.
+            // Up to ~55% of the chart for the live region.
             let max_span =
                 (chart_rect.width() / self.viewport.candle_width() * 0.55).clamp(8.0, 18.0);
             crate::orderflow_render::live_span_for(elapsed, ref_dur, max_span)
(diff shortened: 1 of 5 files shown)


milocaetano:
Superseded by #89, which rebuilt this region: the live lane is now a pane pinned to the chart's right edge, and the `live_tail` reservation this branch patched no longer exists on `main`.

The one thing it fixed that #89 did not is still open, and will be redone on top of `main` rather than merged from here: the lane still gates on `depth_visible()` + `latest_book_ms()`, so it stays inert in replay and on MT5 without DOM. `any_layer_enabled()` + `latest_flow_ms()` is the fix.

PR #96, opened 2026-08-02
feat(indicators): quantick-indicators crate skeleton (M1, PR 1/4)

milocaetano:
First PR of milestone M1 of [docs/indicator-system-plan.md](https://github.com/milocaetano/quantick/blob/main/docs/indicator-system-plan.md): the `quantick-indicators` domain crate skeleton, test-first.

## What's here

- **`IndicatorBar`** — the one-time `Decimal` → `f64` projection of engine bars at the crate boundary, plus derived flow series (`volume`, `delta`, `hl2/hlc3/ohlc4/hlcc4`).
- **`SeriesStore`** — typed full-history columns (f64/bool/color) with the locked *truncate-don't-clone* staging discipline: commit runs persist one row, preview runs stage + roll back in O(columns).
- **`Indicator` trait + `Ctx` + `EvalError`** — the commit/preview execution contract (plan §2.3). `Ctx` carries the host-maintained shared `cvd` series.
- **Plot & input models** — `PlotSpec/PlotBuffer/PreviewFrame`, `InputSpec/InputValue/SourceId` (settings UI will be generated from these).
- **Golden harness** — mirrors `engine::golden`: trade fixture → bars → plot CSV, run twice, byte-identical required; `{:?}` f64 formatting so bit-exactness is the comparison.

Pure domain crate: no UI, no network, no async, no threads, no wall clock. Dependency edge is `indicators → engine` only.

## Deliberate deviation from the plan, flagged

- The `Indicator` trait gains a fifth method, `plots(&self) -> &PlotBuffer` — the committed-columns read path shared by the renderer, the golden harness and future backtest/bot consumers. The plan's four-method trait had no way for the host to read output. Additive, contradicts no locked decision.

## Verification

- [x] `cargo test --workspace` (30 new tests in the crate: series staging/rollback, bar conversion, golden round-trip, preview-commits-nothing, reset+replay determinism)
- [x] arch-review run; both findings fixed in-branch (panic-message column dump; zero-plot row count)


milocaetano:
## PR #96 — arch-review

### Blockers
- none

### Should fix
- `crates/indicators/src/lib.rs:49` (module at `crates/indicators/src/series.rs:86`) — `SeriesStore` / `SeriesId` / `NA_COLOR` are exported as public API but have **no consumer anywhere, not even at the top of the stack**. Grepping the final worktree (`feat/indicators-docs-perf`): the only hits are this re-export and `series.rs` itself plus its own tests. The Pine evaluator that was supposed to be the caller built its own storage instead — `HistRing` + `staged_histories` + `ScriptState::snapshot/restore` at `crates/pine/src/eval.rs:133-200` — and the `ta` kernels keep their own window rings. So the stack ends with two staged-history disciplines, and the one that ships is not this one. That is the standardisation rule ("one way to do a thing, repo-wide") losing to an abstraction nobody docked to, and 406 lines of public surface that the next author has to decide about. Concrete fix: either make `pine::eval::ScriptState` store its histories in `SeriesStore` (its commit/discard contract is exactly `staged_histories` + `restore`), or make `series` private / drop it until a consumer exists. The choice is fine either way; shipping both is not.
- `crates/indicators/src/output.rs:96` — `PlotSpec::id` duplicates information already carried by the spec's position in `IndicatorDescriptor::plots`, and `PlotId`'s own doc (`output.rs:9-11`) defines it as *"an index into `IndicatorDescriptor::plots`"*. Nothing validates the two agree, and `golden.rs:75-86` depends on them agreeing: the CSV header is written from iteration order while each value is read with `plots.value(spec.id, row)`. The same split shows up downstream in the finished stack — `crates/app/src/indicator_render.rs:255` addresses plots by `enumerate()` index, `crates/app/src/indicator_worker.rs:447` by `spec.id`. Related: `PlotBuffer::new(plot_count)` takes a count kept in sync by hand (`tests/golden_harness.rs:40` declares two specs, `:43` writes `PlotBuffer::new(2)`). Concrete fix: drop the `id` field and let position be the id, or add `PlotBuffer::for_plots(&[PlotSpec])` that sizes the buffer from the specs and asserts `plots[i].id == PlotId::new(i)`.

### Consider
- `crates/indicators/src/indicator.rs:121` and `:135` — `ctx: &mut Ctx<'_>` on both `on_close` and `preview`, but `Ctx` holds only `bar_index: usize` and `cvd: &[f64]`, and it never gains a mutable field across all 14 PRs; every implementer immediately downgrades it (`crates/pine/src/script.rs:100` takes `&Ctx<'_>`). `&Ctx<'_>` would state the contract — the host context is read-only — instead of inviting implementers to look for something to mutate. The counter-argument is real (reserving `&mut` avoids a breaking signature change if a draw-object or alert sink lands in `Ctx` later), which is why this is a Consider and not a Should fix.
- `crates/indicators/src/golden.rs:76-78` — plot titles go into the golden CSV header unescaped. `PlotSpec::title` is free-form UI text (`output.rs:98`), so a title like `"Upper, Band"` silently adds a header column and turns a real mismatch into a confusing diff. Cold path, test tooling only. Fix: emit `plot{i}` as the header and keep titles in a leading `#` comment line, or assert titles are comma-free when the descriptor is built.
- `crates/indicators/src/input.rs:73` — `SourceId::as_str` is a 14-arm match that becomes a persistence key, with no test in this PR (the tests at `input.rs:239` cover `value()` only). A duplicated or typo'd arm ships silently. Later PRs close this downstream via `SourceId::ALL` and the round-trip in `crates/app/src/indicators/state_file.rs:83`, so this is a note rather than a gap: a loop over all variants asserting distinct non-empty names would have proved it here, at zero cost.

### Verdict
- **Docking**: strong. 11 files added, 2 edited (`Cargo.toml`, `Cargo.lock`) — a new crate with a one-line workspace registration, no reverse edge (`indicators → engine` only), no fork of bar-building logic. The `Indicator` trait is a real port and the golden test exercises it through a second implementation (`CloseCvd`), not through the production one. The two Should-fix items are both "two sources of truth" rather than missing ports: an abstraction with no implementer (`SeriesStore`), and an id that is derivable from position (`PlotSpec::id`).
- **Performance**: flat, and correctly budgeted. The only recurring path is `IndicatorBar::from` at `bar.rs:54` — 8 `Decimal::to_f64` conversions **per bar, not per trade**, which is the locked decision documented at `bar.rs:4-10` and the whole reason `f64` crosses the boundary once. `PlotBuffer::push_row` is one `Vec` push per column per bar (amortized, no allocation), `SeriesStore::discard_staged` is O(columns) with nothing copied, and the store uses `Vec` throughout — no `HashMap` iteration order can reach the output. `golden` ships in the release build (`lib.rs:37`), which allocates freely, but that matches `quantick-engine`'s own `pub mod golden` (`crates/engine/src/lib.rs:23`) and is not a new pattern. No magic numbers found: `NA_BOOL`, `NA_COLOR` and every UI default are named or come from the spec.
- **Proof**: `tests/golden_harness.rs:78` (`golden_close_cvd_round_trip`) fails if the trades → bars → `IndicatorBar` → commit-run → CSV pipeline drifts or becomes non-deterministic — it runs the whole thing twice from factories and requires byte-identical `{:?}` f64 output; `tests/golden_harness.rs:88` (`preview_commits_nothing`) fails if a preview ever appends a row; `series.rs:297` (`discard_staged_rolls_back_to_committed`) and `:311` (`unstaged_column_commits_na`) fail if the staging discipline or the `na` honesty rule regresses. CI on this PR is green — verified with `gh pr checks 96`.

PR #97, opened 2026-08-02
feat(indicators): incremental ta kernels + fmath guard (M1, PR 2/4)

milocaetano:
Second PR of milestone M1 of [docs/indicator-system-plan.md](https://github.com/milocaetano/quantick/blob/main/docs/indicator-system-plan.md). **Stacked on #96** — merge that first; GitHub will retarget this to `main`.

## What's here

The full §2.5 v1 kernel table minus pivots (M3), each an incremental `Clone` struct, amortized O(1) per push, never re-scanning its window:

- **smooth**: `Sma`, `Ema`/`Rma` (locked semantics: seed = SMA of first `len`, warmup NaN), `Wma` (no-rescan recurrence), `Vwma` (zero-volume window → na, never ±inf)
- **window**: `Highest`/`Lowest`/`HighestBars`/`LowestBars` on one monotonic-deque core (ties → most recent), population `Stdev`, `WindowSum` (`math.sum`), `ValueWhen`, `BarsSince`
- **flow**: `Cum`, `Change`/`Mom`, `Rsi`, `Tr`/`Atr`, `CrossOver`/`CrossUnder`/`Cross`
- **`fmath`**: the *only* transcendental call site (libm = bit-exact cross-platform), enforced by a grep-guard test that fails the moment any other file in `src/` calls `.powf(`/`.exp(`/`.ln(`/`.log*`
- **NaN policy** (documented in `ta::mod`): windowed kernels gate on a NaN-in-window count and recover exactly when it slides out; recursive kernels skip-and-hold; `cum` poisons honestly

## Tests

- 34 new hand-referenced unit tests (warmup, NaN in/out of window, deque plateaus, RSI edge values falling out of IEEE with no special cases, TR gaps, cross NaN rules)
- `golden_ta`: sma/ema/wma over a dyadic ramp fixture — every expected value exact in f64 and checkable by eye — plus the preview-never-advances-kernels property through the whole pipeline
- `fmath_guard`: the grep test

## Verification

- [x] `cargo fmt --all -- --check` / `clippy -D warnings` / `build` / `test --workspace` all green (70 tests in the crate)
- [x] arch-review run: clean (fully additive, no findings)


milocaetano:
## PR #97 — arch-review

### Blockers
- none

### Should fix

- `crates/indicators/src/ta/smooth.rs:219` — **the zero-volume guard cannot do the job its own doc claims.** `if den == 0.0` tests the *running* SMA sum, which retains a rounding residue after eviction, so an all-zero-volume window generally does **not** compare equal to zero. Exact IEEE walk-through for `Vwma::new(2)` fed volumes `0.1, 0.2, 0.0, 0.0`: `0.1+0.2` → `0.30000000000000004`; evicting `0.1` is exact → `0.2_f64 + 1 ulp`; evicting `0.2` leaves `sum = 2.7755575615628914e-17`, so `den = 1.39e-17` and the kernel returns `num / den` — a ratio of two rounding residues, which on a BTC-scale `src·volume` lands in the thousands and is plotted as a plausible-looking price. `smooth.rs:195-196` promises `na` here precisely because "IEEE would say ±inf, which is a lie on a chart"; the residue path is a quieter lie. Reachable: `crates/replay/src/format.rs:627` parses `Volume` with no positivity check, so a recorded session with zero-volume rows produces zero-volume bars (live feeds cannot — `crates/feed-mt5/src/map.rs:262` drops them). Fix: track a non-zero count in `Window` exactly as `nan_in_window` is tracked (`ta/mod.rs:48`) and gate on that count instead of on a float comparison — same chassis, one branch per push, no allocation.
- `crates/indicators/src/ta/smooth.rs:316` — the test for the above (`vwma_zero_volume_window_is_na_not_inf`) pushes only zeros into a virgin kernel, where `sum` is exactly `0.0`. It would pass against a guard written `if den != den`. It proves nothing about the path that actually occurs. Fix: prime with non-dyadic volumes (`push(10.0, 0.1)`, `push(20.0, 0.2)`) before the zeros — that test fails today.
- `crates/indicators/src/ta/window.rs:217` — `(self.sum_sq / len - mean * mean).max(0.0)` is the naive variance, and the clamp is unbounded while `window.rs:181-183` describes it as absorbing only "the tiny negative variance floating rounding can produce". Two error sources the comment does not cover: catastrophic cancellation (both terms are ~mean², the result is σ², so ~`ulp(mean²)` of absolute error survives) and drift in the running `sum_sq` (each push adds and subtracts at scale `len·x²`; error accumulates as √N). For a 20-bar window on a non-integer-priced instrument at 1e5 the magnitudes are small — estimated variance error ~1e-3 after ~1e5 bars, negligible against a normal variance but comparable to a near-flat one. Unmeasured; the bench that would settle it is `Stdev::new(20)` over a 1e6-bar fixture of realistic closes compared against a per-bar exact re-scan. Fix, and it is free: capture an `offset: f64` on the first non-NaN push and accumulate `x - offset` (variance is shift-invariant), so the sums live at σ-scale instead of mean²-scale and the clamp genuinely only ever eats ulp-level negatives. One extra subtraction per push, strictly better than today's implicit `offset = 0`.
- `crates/indicators/src/ta/window.rs:386` — `stdev_is_population_stdev` uses `3, 6, 9` and `4, 4`, all exactly representable, where the naive formula is exact. It cannot regress. Fix: add a case at realistic magnitude with a tight spread (e.g. closes `104532.15 ± 0.02`, len 20) asserting against a hand-computed value with a stated tolerance — that pins the numerical behaviour the shipping instrument actually hits.
- `crates/indicators/src/ta/mod.rs:8-9` — "`Clone` is what makes preview snapshots cheap (a kernel state is a few f64s or a short deque)". `Window` holds `buf: Vec<f64>` (`ta/mod.rs:43`), so cloning **any** windowed kernel is a heap allocation of `len × 8` bytes; `ta.sma(close, 200)` is 1.6 KB per clone. `tests/golden_ta.rs:84` repeats the claim as "previewing on copies costs nothing", and the whole preview design is licensed by it (the stack's host clones the entire kernel map per preview). The rate is set by the caller and it is off the render thread, so this is not a hot-path regression — but the doc is what the next author will size the preview budget against, and it is wrong. Fix: state the real cost (one `Vec<f64>` of `len` per windowed kernel per preview, plus the extremes' deque), or add an allocation-free `peek(&self, x) -> f64` — every kernel here is a pure function of state + input, including `Highest` (`max(front, x)`) and `Wma` (`(numerator - sum + len·x) / weight_total`).
- `crates/indicators/src/ta/mod.rs:5` and `:17` — the module doc says "**No kernel re-scans its window on push**" and that NaN "recovery is exact", and `lib.rs` promises bit-exact output. Both are true of the *logic*, but the consequence is unstated: a running-sum kernel's value at bar N depends on all N inputs, not only on the `len` inputs in its window, so it will not equal a fresh kernel fed only that window. Concretely, `reset()` + replay from a different starting bar — history paging, a trimmed backlog, a replay seek — produces different values for the same visible bars. That is exactly the kind of thing the data-honesty rule exists for, it costs one sentence in this doc block, and it belongs next to the no-rescan claim it is the price of.

### Consider

- `crates/indicators/tests/fmath_guard.rs:15` — `FORBIDDEN` is a denylist of seven method patterns. It misses every other std float transcendental (`.sin( .cos( .tan( .asin( .acos( .atan( .atan2( .sinh( .cosh( .tanh( .exp2( .exp_m1( .ln_1p( .cbrt( .hypot(`) and all UFCS forms (`f64::powf(a, b)`). `fmath.rs:10-11` claims "every transcendental goes through this module"; the guard enforces four families. The dialect deliberately caps `math.*` at pow/exp/log/log10/sqrt, so nothing is broken today — but the failure mode is silent, and the list is one line to extend.
- `crates/indicators/tests/fmath_guard.rs:50` — the scan root is `CARGO_MANIFEST_DIR/src`, i.e. this crate only. In the finished stack `crates/pine/src/eval.rs:1176-1179` is the code that actually implements `math.*` for user scripts; it uses `fmath::` correctly but nothing enforces that it keeps doing so. Making the root a `const` list of crate `src` roots now costs nothing and lets the later crate opt in with one entry.
- `crates/indicators/src/ta/smooth.rs:74` — `ExpSmooth` encodes "not seeded yet" in `self.value.is_nan()`. If the recursion ever yields NaN from non-NaN inputs (±inf closes: `α·(-inf) + (1-α)·inf`), the kernel silently re-enters the seeding branch, `seed_count` overshoots `len`, the `== self.len` check never fires again and it latches NaN forever. An explicit `seeded: bool` compiles away and makes the state machine impossible to misread.
- `crates/indicators/src/ta/window.rs:26` — `VecDeque::with_capacity(len)`, but the deque transiently holds `len + 1` entries: `push_back` runs before the out-of-window `pop_front`. On a monotone series (a steady trend — `Lowest` in an uptrend) that is every bar, forcing one reallocation the first time. `len + 1` makes the per-bar path allocation-free for good.
- `crates/indicators/tests/golden_ta.rs:14` — the golden covers `Sma`/`Ema`/`Wma` only. The mechanism is already built and the fixture is already dyadic; adding `Rsi`/`Atr`/`Stdev` columns would give the flow and window families the same replay-twice regression cover the smoothers get, and would be the natural home for the `Stdev` magnitude case above.
- `crates/indicators/src/lib.rs:37` — the crate doc six lines above still describes `fmath` as "arriving with the `ta` kernels". It arrived in this hunk.

### Verdict

- **Docking**: clean. `Window` + `zero_if_nan` are a real shared chassis, and `ExtremeCore` / `ExpSmooth` factor the two families that would otherwise be copy-paste; a new kernel is a struct plus one `pub use` line in `ta/mod.rs`. The absence of a common `Kernel` trait is the right call, not a gap — the signatures genuinely differ (`push(x)`, `push(src, volume)`, `push(cond, src)`, `push(h, l, c)`) and a trait object would buy a vtable call per bar per kernel for nothing. Dependency direction is intact (`indicators` → `engine` only), nothing is wired to a consumer, nothing activates itself. `libm` is declared with a comment explaining why it exists, which is the standard this repo already sets.
- **Performance**: per-bar-close (commit) and per-preview, not per trade and not on the render thread. Every kernel is genuinely O(1) per push with no window re-scan — the `Wma` recurrence was traced through the NaN-eviction case and it recovers exactly. Steady-state pushes are allocation-free apart from the one-time `VecDeque` growth noted above. The one unquantified cost is clone-per-preview on windowed kernels; the doc calls it free and it is not, magnitude unmeasured.
- **Proof**: 34 hand-referenced unit tests, and they are real references, not restatements — the dyadic-length choices (`Ema::new(3)` → α = 0.5, `Rma::new(2)` → α = 0.5) make every assertion exact rather than epsilon-fuzzed, and `kernel_preview_never_advances_committed_state` (`golden_ta.rs:114`) is the test that fails if the double-advance bug returns. The gaps are specific, not general: `vwma_zero_volume_window_is_na_not_inf` and `stdev_is_population_stdev` both exercise arithmetic that is exact, so neither can catch the two numerical findings above.

PR #98, opened 2026-08-02
feat(indicators): IndicatorHost + native EMA/CVD + eval bench (M1, PR 3/4)

milocaetano:
Third PR of milestone M1 of [docs/indicator-system-plan.md](https://github.com/milocaetano/quantick/blob/main/docs/indicator-system-plan.md). **Stacked on #97** (which stacks on #96) — merge in order; GitHub retargets automatically.

## What's here

- **`IndicatorHost`** (§2.8) — ordered multi-indicator manager that only sees the `Indicator` trait. One `Bar → IndicatorBar` projection per bar for all instances, shared `cvd` column, per-instance error isolation (a failing indicator is disabled with bar index + message; neighbours never notice).
- **Retained projection** — the host keeps the projected bar history (9 words/bar ≈ 7 MB at 100k), which makes `add`/`replace` self-contained: a late-added indicator or an input change (= construct anew + `replace`) replays the full history with the exact per-bar cvd prefix a live run saw. Scripts never observe inputs changing mid-stream (§2.7).
- **Preview staging truncate-don't-clone end to end** — the forming bar's cvd is pushed onto the committed column and popped right after previews; no allocation on that path.
- **`native::Ema` + `native::Cvd`** — the M1 reference indicators. EMA's `overlay` follows the *source's scale* (an EMA of `delta` renders in a pane, not over price — the scale decides, not the indicator kind).
- **Bot-readiness proof** (`tests/bot_readiness.rs`) — a headless consumer builds a host, pushes engine bars and drives a toy EMA-cross-with-flow-filter strategy off committed plot columns. Zero UI anywhere: this is the exact access path a backtester will use.

## Rollback property tests (plan §6, written against the public API only)

EMA over `close, preview×3, close` ≡ `close, close` · staged cvd rolls back · error isolation (no half-appended row) · late-add catch-up ≡ added-at-bar-zero · replace recomputes full history · rebuild determinism + seek-back.

## Bench (`cargo bench -p quantick-indicators`, §5 budgets)

| Path | Measured | Budget |
|---|---|---|
| `ta::Ema/Sma/Rsi/Highest` push | 5–19 ns | — |
| Host commit (projection + EMA + CVD) | **841 ns/bar** | ≤ 50 µs/indicator |
| Host preview | **921 ns** | commit + ≤ 10 µs |
| Rebuild 100k bars × 2 indicators | **77 ms** | 10×5k ≤ 1.5 s |

## Verification

- [x] fmt --check / clippy -D warnings / build / test --workspace (81 tests in the crate)
- [x] arch-review: clean


milocaetano:
## PR #98 — arch-review

### Blockers

- none. `quantick-indicators` still depends only on `quantick-engine` (+ `rust_decimal`, `libm`) — no reverse edge, no feed-to-feed edge. The host is `Vec`-ordered throughout (`instances`, `bars`, `cvd`); no `HashMap`, no clock, no randomness reaches the output. Nothing activates itself: an empty `IndicatorHost` evaluates nothing until `add` is called.

### Should fix

- `crates/indicators/src/native/mod.rs:1` — both native indicators live inline in `native/mod.rs`, so a third one (RSI, VWAP) has to grow the file that holds EMA's and CVD's behaviour. The sibling module in this same crate already set the standard: `ta/mod.rs:28-34` is `mod flow; mod smooth; mod window;` + `pub use`, with the shared chassis (`Window`, `zero_if_nan`) kept there and every implementation in its own file. Matters at priority 2 (extensibility) and 4 (one way to do a thing, repo-wide). Fix: split into `native/ema.rs` and `native/cvd.rs`, leave `PLOT_WIDTH` plus `mod`/`pub use` lines in `mod.rs` — then a third indicator really is a new file plus one registration line.

- `crates/indicators/src/host.rs:90` (and `:75`) — `replace` and `add` set `preview = None` and the host retains no partial, so an instance added or replaced while a bar is forming has no preview until the next `set_partial`. That contradicts the module doc at `host.rs:14-19`, which sells retained history as what makes `add`/`replace` "self-contained": it is self-contained for committed history and silently is not for the forming bar. Traced the consumer to be sure this is not covered downstream — `crates/app/src/indicator_worker.rs:411` only calls `set_partial` when a `PartialUpdated` arrived in the same batch, and no `last_partial` is retained anywhere in the stack. Live, that is a millisecond gap; on a paused replay the changed indicator's line stops one bar short indefinitely. Fix: keep `partial: Option<IndicatorBar>` in the host (written by `set_partial`, cleared by `commit_bar` and by `set_partial(None)`) and re-run the preview for just the new instance at the end of `add`/`replace`. 72 bytes, zero cost on commit and preview.

- `crates/indicators/src/host.rs:132` — the preview-failure branch (`instance.error = Some(e); instance.preview = None`) is never executed by any test. `FailingAt::preview` (`tests/host.rs:121-127`) always returns `Ok`, and it stays that way through the top of the stack (`feat/indicators-docs-perf`). Error isolation is this PR's headline property and it is proven only on the commit path. The branch also encodes a semantic nobody has pinned: a preview failure disables the instance *permanently* — `commit_bar` never clears `error`. Fix: give `FailingAt` a `fail_preview` flag, assert `error(id)` is set after `set_partial`, that the healthy neighbour still has its preview, and that the instance stays disabled on the next `push_closed_bar`.

- `crates/indicators/src/host.rs:70` — `add`'s doc promises "A replay failure lands in the instance's error state, exactly like a live failure", and the catch-up error path is untested; `replace` (`:88`) has the same gap. `adding_late_catches_up_identically` uses a healthy EMA, and `rebuild_replays_identically_and_clears_errors` adds `FailingAt` to an *empty* host, so catch-up never fails anywhere. Fix: push 3 bars, then `add(FailingAt::new(1))`, and assert `error(id).unwrap().bar_index == 1` while the neighbour is untouched.

- `crates/indicators/src/native/mod.rs:22` — `PLOT_WIDTH` carries no unit, and `PlotSpec::width` documents itself as points; `PLOT_WIDTH_PT` costs nothing and removes the px/pt guess. Same line and `:21` (`EMA_DEFAULT_LEN`) are undocumented — the `///` at `:19` binds only to `EMA_COLOR`, and this crate documents even private constants elsewhere (`series.rs`'s `NA_BOOL`). Free at runtime, so priority 5 has no reason to yield here.

### Consider

- `crates/indicators/src/native/mod.rs:70` and `:79` — `Ema::new(21, SourceId::Delta)` builds a descriptor whose input specs still say `default: 9` and `default: Close`. That is defensible (a spec is a declaration, and `InputSpec::default_value` is documented as seeding a *fresh* instance), but nothing in the trait lets a consumer read an instance's current values, and `tests/bot_readiness.rs:59` advertises the descriptor as "the same source of truth as the chart" for a headless config UI. The app works around it by mirroring values itself (`indicator_worker.rs:287-292` seeds from `default_value`). One doc line on `IndicatorDescriptor::inputs` — "declaration-time schema, not the instance's current values" — closes the ambiguity for whoever writes the next consumer.

- `crates/indicators/tests/host.rs:267` — `remove_forgets_the_instance` runs on a host with no bars and no neighbours. The regression that would actually hurt is removing mid-stream: neighbours must keep their columns and the shared `cvd` must not shift. Two extra lines in the existing test.

### Verdict

- **Docking**: the port itself passes, and the stack proves it — `host.rs` is untouched from here to `feat/indicators-docs-perf` except two additive `objects_*` accessors, and `quantick_pine::ScriptIndicator` docks as a whole new crate with zero host edits (the trait grew `objects()` with a default body, so no existing impl was reopened). The failure is one level down: *native* indicators do not dock, because `native/mod.rs` holds implementations instead of declarations. The port is exercised by a second implementation — `FailingAt` in `tests/host.rs:87` is a real non-native `Indicator` — which is exactly what the skill asks for; it just only covers the commit path.
- **Performance**: hot paths are honest and declared. Commit is one `IndicatorBar` projection per bar shared by all instances plus one `cvd` push — no per-indicator re-derivation. Preview stages and pops the `cvd` cell (`host.rs:123`/`:138`), matching `SeriesStore`'s truncate-don't-clone discipline; the `Vec` regrowth is amortised and the popped capacity is reused by the next commit. The retained `Vec<IndicatorBar>` cost (~7 MB at 100k bars) is stated in the module doc with the reason it buys. The one per-event allocation left is `PreviewFrame { values: Vec<f64> }` in `native/mod.rs:576`/`:645`, but its shape is fixed by PR #96 and `indicator_worker.rs:411` calls `set_partial` once per drained command batch, not per trade or per frame — at 921 ns measured against a 60 µs budget this is not worth reshaping. Accessor lookups (`plots`, `preview`, `error`) are linear scans over ~10 instances, correct at that N. The bench matches the repo's existing convention (`engine/benches/hot_path.rs`: dependency-free, `harness = false`, print-only, no asserted thresholds), so its lack of budget assertions is the standard, not a divergence.
- **Proof**: `previews_never_advance_committed_state` is the load-bearing one — an EMA over `close, preview×3, close` must equal `close, close`, which fails the moment a preview mutates committed state. `adding_late_catches_up_identically` and `replace_recomputes_the_full_history` would fail if catch-up drifted from a from-bar-zero run, and `rebuild_replays_identically_and_clears_errors` covers seek-back. Nothing would fail if the *error* paths regressed — that is the gap the two test findings above name.

PR #99, opened 2026-08-02
feat(app): indicator worker + overlay/pane rendering (M1, PR 4/4)

milocaetano:
Final PR of milestone M1 of [docs/indicator-system-plan.md](https://github.com/milocaetano/quantick/blob/main/docs/indicator-system-plan.md). **Stacked on #98** (← #97 ← #96) — merge in order; GitHub retargets automatically.

## What's here

- **`IndicatorWorker`** on the `BookWorker` pattern (§4.1): unbounded commands in, **delta events** out. `Rebuilt` moves full columns only on a real rebuild; `Appended` is one row per closed bar; `Preview` is latest-wins and `PartialUpdated` is coalesced per batch — preview cost is bounded by worker cadence, not feed rate, so a 50× replay can't melt it.
- **`IndicatorViews`** — the UI's own copy of plot columns, synced by applying deltas each frame (the `FeedEvent` shape; no locks anywhere near the render path). Eye toggle is render-side only.
- **`indicator_render`** — batched polylines (NaN breaks segments: warmup renders as a gap, never an interpolated lie), all seven `PlotStyle`s, panes with auto-fitted value scale, zero line and an honest "warming up" label.
- **Paint order**: overlays ride the candles' clip/scale/x-mapping, after candles and before aggression bubbles — the same slot M3's draw objects will take. Panes stack in a 20%-each band (max 3) carved off the chart bottom, sharing the candles' x-mapping.
- **Lifecycle**: backfill / live close / partial / spec switch / prepend / reset all funnel into 4 worker commands; replay seek inherits correct behavior through the existing `Reset` path, exactly as the plan promised.
- **INDICATORS toolbar menu** (hardcoded add EMA(9)/CVD — M2's script browser replaces these entries) + `QUANTICK_INDICATORS_AUTOSTART=1` for scripted validation runs.

## Visual validation (milestone M1 acceptance)

Live Binance BTCUSDT, tick(50), `QUANTICK_INDICATORS_AUTOSTART=1`, 60 fps / 16.7 ms frames:

- amber EMA(9) overlays the candles, starting only after the 9-bar warmup (the NaN gap is honest);
- CVD pane below with label, live value and zero-anchored line, sharing the candles' x-axis;
- two captures 1.2 s apart: forming bar went 23/50 → 26/50 ticks and the CVD preview value moved 3.7557 → 3.7854 **with zero bars committed** ("20+3 bars" in both) — the rollback/preview semantics visibly working;
- backfill|live divider intact; heatmap/bubbles layers unaffected.

## Verification

- [x] fmt --check / clippy -D warnings / build / test --workspace (462 app tests, 8 new)
- [x] arch-review: clean (edits land on the designated ports: FeedEvent drain, ToolbarAction enum, paint-order slots)


milocaetano:
## PR #99 — arch-review

### Blockers

- `crates/app/src/app.rs:2081` — `split_panes` shrinks the chart rect **only inside `draw_chart`**. The three other `plot_split` call sites (`:1551` in `handle_drawing_placement`, `:1777` and `:1781` in `handle_navigation`) still use the full-height `areas.chart`, so as soon as one pane indicator is visible the input path and the draw path disagree about where prices are. Concretely: `:1797` builds `drawing_scale` from `areas.chart.top()/bottom()` (full height) while `drawing_point_at` at `:1526` builds its scale from `self.last_chart_top` / `self.last_chart_height`, which `draw_chart` sets from the **shrunk** rect at `:2325-2326`. Two scales for the same pixels, in the same feature: a drawing is *placed* where you clicked but *hit-tested* against a scale stretched 20% (or 60%, at three panes) — you can no longer grab the line where you see it. `:1929` compounds it, dividing the translate delta by `areas.chart.height()` (unshrunk) while the candle price pan two blocks down correctly uses `last_chart_height`. And both interact rects (`:1554-1555`, `:1790`) still cover the pane band, so a drag started inside the CVD pane pans the candles and an armed drawing tool places a point there. This is a regression on drawing tools (merged in #94), it activates the moment `QUANTICK_INDICATORS_AUTOSTART=1` adds the CVD pane, and it survives the whole stack — `feat/indicators-docs-perf` still calls `split_panes` from `draw_chart` alone. Fix: give the layout one owner — add a `pane_count` parameter to `plot_split` so all four call sites get the same `chart` rect, and delete the local `split_panes` call from `draw_chart`. Regression test: with one pane visible, place a horizontal line via `click_chart` and assert a second click on the same pixel selects it (it does not today).

### Should fix

- `crates/app/src/indicator_render.rs:223` + `crates/app/src/app.rs:1216` — the columns are drawn as if index-aligned with `state.bars()`, but they are a worker round-trip behind. `update` drains the feed (sending `BarClosed`) at `:3691` and only then drains events at `:3694`, so on the frame after a bar closes `partial_slot == closed.len()` is one past the last committed cell: `draw_line` joins the last committed point straight to the preview point, interpolating a segment across a bar that has no computed value — precisely the lie the NaN-gap design exists to prevent. Worse after `+ older`: `prepend_history` grows the bars at the front and `:1214` calls `drawings.shift_bars(added)` for drawings, but `:1216` only asks the worker to rebuild, so for the round-trip the whole overlay is drawn shifted by `added` slots — every EMA point on the wrong candle. Fix: gate the preview with `.filter(|&slot| slot == column.len())` at `:223`, and mirror the drawings pattern with an `IndicatorViews::shift_rows(added)` at `:1216` that prepends `added` NaNs to every column, which keeps the alignment and renders the unknown region as an honest gap.
- `crates/app/src/indicator_worker.rs:246` — `if rebuilt || mirror.known_rows == 0` uses "no rows yet" as a proxy for "the UI has not been told yet", so while no bar has closed **every batch** re-sends a full `Rebuilt`: a cloned `IndicatorDescriptor` (two `String`s plus a `String` per `PlotSpec`) and a fresh `Vec<Vec<f64>>` per slot, at feed-batch rate. That window is not a blink — it is the whole first bar on a slow tape (MT5 WINQ26, a large tick N), and it reopens on every `reset_market_state` and every replay seek, both of which send `Rebuild(empty, None)` and drive `known_rows` back to 0. Fix: add `synced: bool` to `SlotMirror` and branch on `rebuilt || !mirror.synced`.
- `crates/app/src/indicator_render.rs:311`, `:341`, `:362` — per-element draw calls where this repo already batches. `draw_area` allocates a `vec![…]` of four `Pos2` **per visible bar per frame** and pushes one `Shape::convex_polygon` each; `draw_bars` pushes one `Shape::Rect` per bar; `draw_markers` one circle per bar. With ~400 visible slots at 60 Hz that is ~24k heap allocations/s for `Area` alone, plus one tessellation pass each. `orderflow_render.rs:1194` states the repo's answer in a comment — "hundreds of stroked segments per frame would pay stroke tessellation each; one mesh keeps the per-frame cost flat" — and implements it with `egui::Mesh` + `Shape::mesh`. Neither shipped native reaches these paths (`native/mod.rs:61` and `:149` are both `PlotStyle::Line`, the already-batched `Shape::line`), which is why this is Should-fix and not a Blocker; it is also exactly why it must be fixed before M2 scripts start emitting histogram and area plots onto this port. Fix: build one `egui::Mesh` per plot (two vertices per point for `Area`, one quad per bar for `Histogram`/`Columns`) and `painter.add(Shape::mesh(mesh))` once.
- `crates/app/src/indicator_render.rs:1` — the module has zero tests, and its central honesty claim is the untested part. The doc comment promises "NaN cells break a polyline into segments … warmup renders as a gap rather than a lie interpolated across missing data"; nothing fails if `flush_segment` stops flushing. Five of the seven `PlotStyle`s are also unreachable in M1, so they ship neither tested nor exercised. Name the tests: `value_range` returns `None` for an all-NaN window and ignores NaN otherwise (`:178`); `last_value` prefers the preview and skips trailing NaN (`:163`); and one test over `egui::Context::default()` asserting a column with a NaN in the middle emits two `Shape::line`s, one with a single point emits the fallback dot. Separately, both toolbar tests pass `indicators: Vec::new()` (`toolbar.rs:743`, `:797`), so the entry rows added at `toolbar.rs:531-551` — status dot, eye, trash — are never drawn in any test; passing a non-empty vec in one of them costs nothing and covers the new menu.
- `crates/app/src/indicator_render.rs:114`, `:137`, `:144`, `:152`, `:202-205` — magic numbers in a renderer, against a neighbour file (`orderflow_render.rs:240-339`) that names every single one with a unit suffix. The pane frame's `gamma_multiply(0.4)`, the zero line's `gamma_multiply(0.3)`, the label insets `vec2(6.0, 3.0)` / `vec2(-6.0, 3.0)`, and `format_value`'s `1000.0` switch with its `.1` / `.4` precisions all belong at module top as `PANE_FRAME_ALPHA`, `ZERO_LINE_ALPHA`, `PANE_LABEL_INSET_PX`, `LARGE_VALUE_THRESHOLD`, `LARGE_VALUE_DECIMALS` / `SMALL_VALUE_DECIMALS`. `PANE_LABEL_FONT` at `:33` should be `PANE_LABEL_FONT_PX` — every other constant in the file already carries its unit.

### Consider

- `crates/app/src/indicators/mod.rs:161` — `.take(MAX_PANES)` silently drops the fourth pane indicator, but the toolbar still paints it with the green `CIRCLE` dot that means "drawn" (`toolbar.rs:537`). The doc comment calls this "the honest alternative to shrinking panes into unreadability"; make the UI say it — a muted dot plus a hover "no pane available (max 3)" for entries past the cap.
- `crates/app/src/toolbar.rs:533` — `IndicatorMenuEntry` carries `errored: bool` but never the message, so a failed indicator reads as broken with no reason given while `EvalError` already holds `bar_index` and `message`. Carry `error: Option<String>` and hang it off the dot's hover text.

### Verdict

- **Docking**: good, with one seam. The worker/views/render split lands on real ports — `IndicatorViews` is fed only by `IndicatorEvent`, the toolbar speaks raw `u64` slots and never sees worker types, `IndicatorSource` is a two-variant enum M2 extends with one arm, and paint order and lifecycle funnel into existing slots (`FeedEvent` drain, `ToolbarAction`, `reset_market_state` → `Rebuild`). The seam is the layout: the pane band is carved in one place and ignored in three, which is the Blocker.
- **Performance**: the render path is sound in shape and the M1 styles are batched — `Line`/`StepLine` cost one `Shape::line` per segment per frame, and the per-frame passes (`value_range`, `cells`) are bounded by visible slots, matching what the candles already do. The worker design is the right one and is copied faithfully from `BookWorker`: unbounded queue drained into batches, latest-wins `PartialUpdated`, delta events instead of column clones, `Bar` clones that are pure memcpy with no heap behind them, `BTreeMap` for deterministic emission order. Nothing blocks the render thread. What is not flat: the `known_rows == 0` full-`Rebuilt` stream (sustained, feed-rate, per slot) and the per-element `Area`/`Histogram`/`Columns`/marker draws (~400 shapes and, for `Area`, ~400 allocations per frame) that M2 will be the first to hit. The magnitude of the second is unmeasured here — a frame-time reading from the perf HUD with a CVD plot forced to `PlotStyle::Area` over a full viewport would settle it.
- **Proof**: `delta_stream_replays_to_identical_columns` (`indicator_worker.rs`) is the strong one — it asserts bit-for-bit equality between the delta-replayed views and a directly-driven host, so a broken `Appended`/`Preview` path fails loudly; `rebuild_resyncs_the_views` and `remove_isolates_the_survivor` cover the other two lifecycle shapes, and `pane_split_is_stacked_and_bounded` pins the geometry. Eight new tests, all real. The gaps are `indicator_render.rs` (nothing at all — no test fails if warmup stops rendering as a gap) and the Blocker itself: no test compares the rect the input path uses against the rect the draw path uses, which is why a 20% price-scale divergence ships with green CI.

PR #100, opened 2026-08-02
feat: add Hyperliquid trades and L2 feed

milocaetano:
## What changed

- add the standalone `quantick-feed-hyperliquid` crate for public perpetual trades and visible top-20 L2 data
- recover the first factual WebSocket trade batch, then forward venue batches through the shared `FeedEvent` path instead of sending one channel message per trade
- map venue-reported aggressor sides into the shared deterministic engine contract and convert complete L2 images into provider-neutral snapshots/deltas through `SnapshotDiffer`
- reject malformed or regressive L2 images atomically; a backwards venue timestamp ends that capture generation instead of being clamped to invented time
- add explicit provider-neutral connection lifecycle for Binance, Hyperliquid and MetaTrader so a quiet market or stale latency observation cannot masquerade as a healthy/disconnected socket
- preserve the complete Binance + Hyperliquid + MetaTrader selector while allowing startup-only feed/symbol overrides through `QUANTICK_DEFAULT_FEED` and `QUANTICK_DEFAULT_SYMBOL`
- cache settled heatmap projections by chart/book revision and coalesce dirty rebuilds, while keeping the live edge current
- seed BTC, ETH, HYPE, SOL and ZEC in the built-in Hyperliquid selector and document its limited depth/history capabilities

## Why

Quantick should expose Hyperliquid order flow through the same chart, bar engine, bubble layer and L2 heatmap used by the other sources, without copying aggregation/rendering logic. Startup and rendering also needed to remain responsive, and adding one source must not remove Binance or MetaTrader from the configured catalog.

## User impact

Users can open Hyperliquid without an account or API key, switch among the five seeded perpetual markets, and use the same L2/bubble controls as Binance. The first venue batch appears immediately as factual recovery data; older-history paging stays disabled because the source does not expose pageable trade history. The status bar now distinguishes connection lifecycle from arrival latency, including honest reconnect transitions.

## Architecture review

Final `arch-review`: **CLEAN** — no outstanding Blocker or Should-fix findings.

- **Docking:** Hyperliquid enters through `FeedHandle`, `FeedCapabilities`, `Trade` and `DepthEvent`; no feed depends on another feed, and the engine remains UI/network-free.
- **Data honesty:** no synthetic history, no inferred aggressor side, no timestamp clamp, and limited L2 coverage is declared explicitly.
- **Performance:** trade delivery is batched per WebSocket frame; settled projection work is revision-driven and rate-limited when dirty; connection signals stay off the data hot path.
- **Proof:** deterministic tests cover wire mapping, batching/deduplication, L2 generations, atomic timestamp rejection, catalog preservation, cache invalidation, arrival latency, and connected -> disconnected -> connected lifecycle for Binance, Hyperliquid and MetaTrader. A regression test also proves late MT5 supervisor notices cannot overwrite a healthy connection state.

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo build --workspace`
- `cargo test --workspace` (including 469 app tests)
- `cargo test -p quantick-feed-hyperliquid --test live_market_smoke -- --ignored --nocapture`
  - BTC: 30 trades + L2 OK
  - ETH: 30 trades + L2 OK
  - HYPE: 30 trades + L2 OK
  - SOL: 30 trades + L2 OK
  - ZEC: 30 trades + L2 OK
- native app smoke from the built-in three-feed catalog: Hyperliquid BTC recovered 30 trades, synchronized 20 bids/20 asks, kept the book queue empty and rendered the live heatmap with no dropped primitives


milocaetano:
## PR #100 — arch-review

Read at head `42dbbea` in `F:/src/quantick-worktrees/feat-hyperliquid-feed`. CI is green (`ci pass 2m32s`, run 30750045574); cargo was not re-run. Docking is genuinely clean: no `feed == "hyperliquid"` anywhere, the only `ProviderKind` matches outside `config.rs` are the documented `spawn` dispatch (`feed/mod.rs:270-272`) and the capability declarations; the "load older" affordance gates on `capabilities.history_paging` (`toolbar.rs:390`, `:503`), the heatmap on `capabilities.book_capture` (`app.rs:803`, `:812`). Findings below are mostly about what the lifecycle change took away, not about the new venue.

### Blockers

- `crates/app/src/app.rs:1313` (with `statusbar.rs:153` and `statusbar.rs:207-210`) — `trade_lag_ms()` now returns a stored observation that is never re-derived, and the status bar still renders it as `lag 42 ms` next to a green `live` dot. The number stops ageing the moment prints stop, so a socket that is open but delivering nothing displays a fixed, live-sounding latency indefinitely. This is not hypothetical: neither session loop has a read-idle watchdog and neither verifies that a `pong` came back (`feed-hyperliquid/src/stream.rs:102-174`, `feed-hyperliquid/src/depth/stream.rs:110-202`, and `feed-binance/src/stream.rs` likewise), so a half-open TCP connection produces no error, no `connected.send(false)`, and no state change. Deleting `FeedState::Stalled` (`statusbar.rs:69-77`) removed the only signal that used to catch exactly this, and the surviving WARN branch at `statusbar.rs:209` can no longer be reached by ageing — only by a single print arriving >5 s late. Data honesty comes before everything else in the skill's order, and the app is now making a positive claim it cannot back. Splitting transport from tape is the right call; the missing half is the tape side. Concrete fix: keep `FeedConnectionState` driving the dot, relabel the readout to what it measures (`arrival 42 ms`), and add a separate wall-clock staleness badge computed per frame from the newest event timestamp — `book.last_event_ms` is already carried through `OrderflowHealth` and logged at `app.rs:1351`, so the trade side needs one retained `latest_trade_ms` and one subtraction. A read-idle timeout in the session loops would fix the wedged socket rather than only reporting it.

### Should fix

- `crates/feed-hyperliquid/src/reconnect.rs:11` — `Backoff` is a rename-only copy of `crates/feed-binance/src/reconnect.rs:27`: same xorshift, same equal-jitter formula, same `for_feed(500 ms, 30 s)` defaults, `next_random_unit` vs `next_rand`. Standardisation says one way to do a thing repo-wide, and because feed crates may not depend on each other the fourth venue copies it a third time. The rule forbids `feed-* → feed-*` edges, not a shared support crate: extract `Backoff` (and the `watch::Sender<bool>` connection contract this PR introduces) into a small `crates/feed-core` that depends on nothing, and have both feeds depend on it. The dependency direction stays one-way.
- `crates/app/src/feed/mod.rs:25` — `pub use quantick_feed_binance::depth::DepthEvent;` routes the app's provider-neutral depth type through one specific venue, while `crates/app/Cargo.toml:13` already depends on `quantick-orderbook`. This PR is the one that makes the channel multi-provider — three feeds now publish on it — so the neutral port is named after a competitor of two of its producers. It is a broken port in name only today, but it is what the next feed author copies. Fix: `pub use quantick_orderbook::DepthEvent;`, zero new dependencies, and `feed/hyperliquid.rs:18` stops importing the type from its own crate's re-export.
- `crates/feed-hyperliquid/src/backfill.rs:113` — `fetch_recent_trades`, `HyperliquidHttp` and `RecentTradesSource` have no caller anywhere in the workspace; the only references are the `pub use` at `lib.rs:30-33` and the module's own fake-source test. The commit "fix: improve Hyperliquid startup throughput" moved recovery to the first WebSocket batch (`feed/hyperliquid.rs:100-117`) and left the REST path behind. It is 159 lines of untested-against-reality code that drags `reqwest` + `rustls-tls` into the build for nothing (`crates/feed-hyperliquid/Cargo.toml:15`). The `#[ignore]`d smoke test does not exercise it either. Either delete the module and the dependency, or wire it into `feed_task` as the startup recovery it was written for so it has a caller and can be proven.
- `crates/app/src/metrics.rs:86` — `received_at_ms - timestamp_ms` is a plain `i64` subtraction on a path that now runs for every accepted depth update as well as every trade (`orderflow_engine.rs:774`, `:800`). This repo has an established rule that feed arithmetic saturates rather than panics. Hyperliquid does validate `time >= 0` upstream (`wire.rs:227`, `depth/mapper.rs:178`), but `feed_lag_ms` is a shared app-layer helper fed by three feeds plus replay, so the guard is not local to the arithmetic. Fix: `received_at_ms.saturating_sub(timestamp_ms)` — free at runtime, and the doc's "reported as-is, not clamped" stays true for every realistic value.
- `crates/app/src/app.rs:1436` — the `HEATMAP_HIGH_LAG` message still reads "order-book events are behind wall clock", but `book_lag` (`app.rs:1329`) is now `book.arrival_latency_ms`, a frozen observation from the last accepted event with no clock comparison in it. The `feed_lag_ms` field name at `app.rs:1343` has drifted the same way. These are `schema_version`-stamped structured events that ops and log tooling read literally. Rename the fields and messages to arrival latency, or restore the clock-based measure the text describes (see the blocker — one fix can serve both).

### Consider

- `crates/app/src/orderflow_engine.rs:678` — `record_trade` marks the settled half dirty on every print, so on any live venue the projection is dirty essentially always and the new revision cache collapses back to the old 220 ms cadence. The invalidation itself is correct — every path that can change `project_settled`'s output was traced (`lane` and `on_newest_bar` are in `ProjectionLayout` at `:125-126`; `live_edge`'s `latest_book_ms` only moves on an accepted event, which dirties; `apply_auto_base` is reached only after `invalidate_projection()` at `:752`; config changes invalidate at `:586`, `:607`, `:665`) — so this is a calibration point, not a bug: the win is real only when tape *and* book are both silent (overnight MetaTrader, paused replay, bubbles-only with capture off). Worth saying plainly in the PR body rather than as an unqualified perf claim. `heatmap_projection_builds` vs `heatmap_projection_cache_hits` in `APP_HEALTH_SUMMARY` over a BTC session settles the magnitude.
- `crates/feed-hyperliquid/src/depth/stream.rs:155` — `_ => unreachable!()` destructures a `DepthStatus` built one line earlier just to read back level counts the mapper already holds. Have `synchronized_status()` return the counts alongside the status, or read `mapper` directly for the log fields, and the catch-all disappears.
- `crates/feed-hyperliquid/src/depth/stream.rs:134` — every crossed image emits a `warn!`, while `crates/feed-mt5/src/depth.rs:264` solves the identical problem with `warn_once`. At the ~2 images/s l2Book cadence this is bounded, but an auction or venue incident floods the log at that rate. Copy the `warn_once` shape (feed crates cannot share it) or drop to `debug` after the first.
- `crates/feed-hyperliquid/tests/live_market_smoke.rs:12` — `SEEDED_MARKETS` restates the shipped catalog a third time, after `crates/app/config/feeds.toml:43` and the assertion in `crates/app/src/config.rs`. The `#[ignore]`d test is the one that will silently stop covering what ships. Move the list to a `pub const` in `quantick-feed-hyperliquid` and assert the embedded config against it from the app's config test — the app already depends on the feed crate, so the direction is fine.
- `crates/feed-binance/src/stream.rs:66` vs `crates/feed-hyperliquid/src/stream.rs:151` — the same `watch::Sender<bool>` is set on WS handshake for one feed and on `subscriptionResponse` for the other. Both are defensible (Binance subscribes via the URL path, so there is no separate ack), but the parameter carries no doc comment in either signature and the next venue has no written rule to follow. One line on `connected:` stating "the live transport for this subscription is established" closes it.

### Verdict

- **Docking**: yes. A fourth venue is a new crate plus one arm in `feed/mod.rs:270-272`, one `ProviderKind` variant with its `capabilities()`, and one `[[feeds]]` block. No UI gates on which venue is playing; the seeded symbols are config with a dated, sourced rationale (`feeds.toml:32-38`); the endpoints are named constants, matching the Binance precedent. The `connection_notice` abstraction (`feed/mod.rs:203-216`) is a found one, not an imposed one — the MetaTrader edit is two `FeedNotice` constructor swaps (`metatrader.rs:195`, `:208`) and the Binance edit threads one `watch` sender through an existing signature. Blast radius is proportionate. The two deductions are the venue-named neutral depth re-export and the third copy of `Backoff`.
- **Performance**: flat on the hot paths, better when idle. Trade ingest allocates one `Vec<Trade>` per WebSocket batch, not per trade (`wire.rs:166`), and dedupe is an O(log n) `BTreeSet` probe over a bounded 65 536-key window. Wall-clock reads went from O(trades) to one per UI drain (`app.rs:1147-1183`, `:1272-1288`), proven by `one_ui_drain_uses_one_observation_for_single_and_batched_trades`. The settled-projection cache is a strict improvement — it can now skip a rebuild the old cadence would have forced — but see the Consider above for how often that actually fires on a live tape. `last_projection_ms` also got tighter, now timing only the build rather than the build plus window setup (`orderflow_engine.rs:988-990`).
- **Proof**: strong on the new crate, thin on the thing the blocker is about. `backwards_time_rejects_the_entire_image_and_preserves_state` (`depth/mapper.rs:366`) would catch a regressive-timestamp regression; `clean_projection_cache_survives_cadence_and_dirty_updates_coalesce` and `a_new_bar_timeline_rebuilds_an_otherwise_identical_layout` (`orderflow_engine.rs:1644`, `:1712`) would catch a stale or over-eager cache; `connection_status_tracks_disconnect_and_resubscribe` exists on both feeds against a real local WebSocket server. The port is exercised by two implementations plus the MetaTrader path, which is what makes it a port. What no test covers: a transport that stays open and stops delivering — precisely the case the removed `Stalled` state used to report. That gap is the blocker, and the test that would close it is a status-bar assertion that a connected feed with an ageing newest-event timestamp reads as stale.

PR #101, opened 2026-08-02
feat(pine): lexer + parser + conformance corpus (M2, PR 5/8)

milocaetano:
First PR of milestone M2 of [docs/indicator-system-plan.md](https://github.com/milocaetano/quantick/blob/main/docs/indicator-system-plan.md). **Stacked on #99** (← #98 ← #97 ← #96) — merge in order.

## What's here

The `quantick-pine` crate: the language frontend of the Quantick Pine dialect, hand-rolled with **zero dependencies** (§3 locked decision).

- **`error.rs`** — `PineError` = stable code + byte span + message + notes, rendered `file.pine:12:8: message (PINE_NO_SECURITY)`. Every stage collects *all* errors (a script with three problems reports three). The §3.6 code catalog is in place (`PINE_LEX`, `PINE_SYNTAX`, `PINE_INDENT`, `PINE_NO_SECURITY`, `PINE_STATEFUL_IN_LOOP`, …).
- **`lexer.rs`** — indentation-aware (Indent/Dedent, strictly-deeper blocks), continuation on unclosed brackets or trailing operator/comma. Documented divergences: no continuation after `=>` (that ends a multi-line function header — the plan listed it, but it would swallow every block body); tabs = 4 columns. `//@version` read from comments; `#RRGGBB[AA]` literals.
- **`ast.rs`** — flat arena (`Vec<Node>` + `NodeId`), spans on every node, no `Box` chains.
- **`parser.rs`** — recursive descent + precedence climbing. Handles: `var`/`varip`/plain/typed declarations, `:=`, tuple destructuring *and* tuple expressions (function multi-returns, disambiguated by bounded lookahead), user functions (inline + block bodies), `if`/`for`/`while` as expressions, ternary (right-assoc), named call args, history `[n]`, member chains. `switch` is rejected honestly (`PINE_UNSUPPORTED`, may land M4). Error recovery swallows a failed header's indented block so one bad statement reports once.
- **Conformance corpus** (§6): `tests/corpus/ok/*.pine` must parse (starter-pack shapes); `tests/corpus/err/*.pine` must fail with **exactly** the codes+lines declared in their `// expect: CODE@LINE, …` directive — multi-error collection asserted.

## Verification

- [x] fmt --check / clippy -D warnings / build / test --workspace (29 new tests: 27 unit + 2 corpus walkers over 10 fixture scripts)
- [x] arch-review: 1 finding (declared-but-unused dependency), fixed in-branch


milocaetano:
## PR #101 — arch-review

### Blockers

- `crates/pine/src/lexer.rs:446` — **a single non-ASCII byte panics the lexer instead of producing `PINE_LEX`.** The `_` arm of `token()` does `self.pos += 1` and then slices `&self.source[start..self.pos]`; for any multi-byte character (`é`, `—`, a smart quote, or the U+00A0 non-breaking space that arrives with every copy-paste from a web page) that index is inside the character and `&str` indexing panics with *"byte index N is not a char boundary"*. Priority 0 (correctness): the one function whose job is to convert bad input into an honest error aborts instead, and there is no `catch_unwind` anywhere in the workspace — `crates/app/src/indicators/library.rs:158` feeds `fs::read_to_string` straight into `quantick_pine::compile`, so a user's script file takes the app down with a Rust panic message rather than `file.pine:3:5: unexpected character (PINE_LEX)`. Checked the top of the stack (`feat/indicators-docs-perf`): the line is byte-identical there, so no later PR fixes it. Fix: use the idiom this file already has 100 lines below at `lexer.rs:553-558` — `let ch = self.source[start..].chars().next().expect("in-bounds char"); self.pos = start + ch.len_utf8();` — and report `ch`. Add `tests/corpus/err/non_ascii.pine` with `// expect: PINE_LEX@3` so it stays fixed.

- `crates/pine/src/error.rs:40` (span manufactured at `crates/pine/src/lexer.rs:344`) — **the rendering path itself can panic on a span that is not a char boundary.** `line_col` slices `&source[..self.start]` unguarded, and `end_logical_line` builds the closing `Newline` token as `Span::at(self.pos.saturating_sub(1))`, which at EOF is `len - 1` — the middle of the last character when the file ends in a multi-byte one without a trailing newline. Two-line repro: `x = // ré` (no trailing `\n`) → `primary()` reports *"expected an expression, found end of line"* carrying that span, and `render` panics while printing the error. Narrower trigger than the finding above, same class and same blast radius. Fix: record the newline's own byte offset rather than `pos - 1`, and make `line_col` total — floor `start` to a boundary (`while start > 0 && !source.is_char_boundary(start) { start -= 1 }`) before slicing. Rendering a diagnostic must be infallible; it is the last thing standing between a bad script and a crash.

### Should fix

- `crates/pine/src/parser.rs:714` (and the `-`/`not` recursion at `:599`) — **unbounded recursion on nesting depth.** `primary` → `expression` → `ternary` → `binary` → `unary` → `postfix` → `primary` costs ~5 frames per `(`, with no depth counter anywhere in the stack (grepped the top branch). A file of 100k `(` — corrupt, generated, or truncated — overflows the stack, and a stack overflow aborts the process: no unwind, no `PineError`, nothing for `library.rs` to render. Note the recursion happens on the way *down*, so even unbalanced input that the parser would eventually reject never reaches the rejection. Cold path, so the guard is free: a `depth: usize` field on `Parser`, a `const MAX_NEST_DEPTH: usize = 64;` at module top, and a `PINE_SYNTAX` "expression nests too deeply" when it trips. A test with `"(".repeat(MAX_NEST_DEPTH + 1)` proves it.

- `crates/pine/src/parser.rs:784` — **`describe`'s fallback leaks Rust `Debug` names into script-author-facing messages.** `other => format!("`{other:?}`")` covers every operator and keyword-not-listed, so the shipped corpus fixture `three_problems.pine` (`b = ]`, `c = ,`) renders as *"expected an expression, found `RBracket`"* and *"found `Comma`"*. Nobody wrote `RBracket`; they wrote `]`. The err-corpus walker only compares code+line, never the message, so this is invisible to the suite. Priority 5 but free at runtime (compile-once path): spell the punctuation out in the match (`Tok::RBracket => "`]`"`, `Tok::ColonEq => "`:=`"`, `Tok::KwVarip => "`varip`"`, …) and drop the catch-all so a new token can't silently regress the message — exhaustive `match` over `_ =>` is the repo's stated preference.

- `crates/pine/src/parser.rs:231` (and `:198` for the `var` arm) — **a qualified type annotation is misdiagnosed.** The lookahead accepts exactly one annotation word (`Ident Ident Eq`), so ordinary Pine v5 like `simple int len = 9` or `var series float x = na` falls through to `expression_statement` and reports *"expected end of line, found `int` (one statement per line)"* — a message that points at the wrong thing and names no reason. The dialect reference says annotations are "accepted and ignored", so the honest behaviour is to skip qualifier + type the same way; alternatively reject with `PINE_UNSUPPORTED` naming type qualifiers. Either is a few lines; the current outcome is the one the crate's whole error contract exists to prevent. Cover it with an `ok/` or `err/` fixture.

- `crates/pine/src/ast.rs:244` — **the arena doc states an invariant the parser does not hold.** *"Node 0 is always the `NodeKind::Script` root once parsing succeeds"* — but `script()` (`parser.rs:179`) pushes the root **after** every statement, so node 0 is whatever literal was parsed first and the root is the last id. Nothing depends on it today only because `parse` returns the root explicitly; a later pass that trusts the comment reads a random node. Free to fix: state that `parse` returns the root id and that the arena is in push order (children before parents).

### Consider

- `crates/pine/src/parser.rs:50` — `peek`/`peek_at`/`span` all compute `self.tokens.len() - 1`, which underflows on an empty slice. `lex` always emits `Eof` so the crate's own path is safe, but `parse` is `pub` and takes `&[Token]`; a `debug_assert!(!tokens.is_empty(), "the token stream always ends in Eof")` in `parse` documents the precondition at zero cost.
- `crates/pine/src/lexer.rs:236` — the `&& !self.is_version_comment()` guard has no observable effect: when it fires, `skip_blank_prefix` returns and `run`'s loop immediately calls `comment()` on the same bytes, reading the directive anyway. Dead condition; drop it or comment why it is load-bearing.
- `crates/pine/tests/corpus.rs:30` — `every_ok_script_parses` asserts only *absence of errors*; a regression that builds a wrong-but-well-formed tree (ternary going left-assoc, a named arg swallowed as positional — `delta_histogram.pine` has exactly that shape: `color=delta >= 0 ? … : …`) passes it. The unit tests in `parser.rs` do cover those shapes, and `script_semantics.rs` later in the stack covers values, so this is a note, not a gap: an AST digest (node count + kind histogram) per `ok/` fixture would make the corpus itself load-bearing for the same cost as the walker it already has.

### Verdict

- **Docking**: clean. `quantick-pine` is a new pure domain crate with genuinely zero dependencies, one added workspace member, no reverse edge, and no existing file touched outside `Cargo.toml`/`Cargo.lock` — all adds, no surgery. The arena AST with a span on every node is the right port for the compile passes and interpreter that dock in PRs 6-7; `PineError` (stable code + span + notes, collected not short-circuited) is one error type for every stage, which is the standardisation call I would have asked for. The one thing that will force a later author back into these files is the depth guard and the `describe` table — both cheap now, both awkward once scripts are in the wild.
- **Performance**: flat, and correctly so. Everything here runs once per script load; the `String` per identifier, the `peek().clone()` at statement dispatch and the `format!` per error are all in that budget, and cold-path clarity is spent well (no micro-optimisation to call out in reverse). `Vec<Node>` + `NodeId` is the shape the per-bar walker needs; the real measurement — nodes visited per bar — is owed by PR 7, not this one.
- **Proof**: solid for what it claims. 27 unit tests plus two corpus walkers with `checked >= 5` floors that stop the corpus shrinking silently. A precedence regression fails `precedence_follows_pine`; a ternary associativity regression fails `ternary_is_right_associative`; losing the call-vs-definition lookahead fails `a_call_statement_is_not_mistaken_for_a_definition`; breaking multi-error collection fails `three_problems.pine`. What no test would catch: the two panic paths above (no fixture contains a non-ASCII byte anywhere) and any nesting depth at all.

PR #102, opened 2026-08-02
feat(pine): compile passes — slots, call sites, const folding, reject scan (M2, PR 6/8)

milocaetano:
Second PR of milestone M2 of [docs/indicator-system-plan.md](https://github.com/milocaetano/quantick/blob/main/docs/indicator-system-plan.md). **Stacked on #101** (← #99 ← #98 ← #97 ← #96) — merge in order.

## What's here

The §3.4 compile passes, arena AST → `CompiledScript`, every pass collecting errors so a script with three problems reports three:

- **`builtins.rs`** — the one registry: every accepted name (including the flow series Pine doesn't have: `delta`, `cvd`, `buy_volume`…), statefulness (call-site policy), top-level-only rules, per-family rejection reasons, and the did-you-mean vocabulary (Levenshtein with a budget that hints on transpositions and stays silent on garbage).
- **Dense per-node tables** — resolutions, call-site ids, plot/input indices, implicit history-series ids, all indexed by `NodeId`: **zero string lookups at eval time**, the precondition of the §3.5 interpreter budget.
- **Pass 1** — `indicator()` header (unknown args accepted-and-ignored with warnings), the fixed top-level plot registry (style/color/linewidth const-folded, `color.new/rgb` transparency math included), `input.*` → `InputSpec` extraction.
- **Pass 2** — scope-stack resolution: globals vs function-frame locals vs user functions; unknown names carry did-you-mean notes.
- **Pass 3** — `CallSiteId` per stateful kernel call; `PINE_STATEFUL_IN_LOOP`; `PINE_RECURSION`.
- **Pass 4** — kernel lengths must const-fold to a positive integer, through never-reassigned globals and input defaults (`PINE_SERIES_LENGTH`); a `:=` anywhere honestly kills foldability.
- **Pass 5** — `max_bars_back` from constant offsets; dynamic offsets capped at 500; every `History` node gets its own implicit series id.
- **Pass 6** — the §3.1 reject list with honest reasons (`PINE_NO_SECURITY` explains that activity-sampled charts have no timeframes, etc.).

## Verification

- [x] fmt --check / clippy -D warnings / build / test --workspace (17 new tests: 4 registry + 13 pass-level through the public `compile()`)
- [x] arch-review: 1 finding — `color.*` constants outside const positions resolved to nothing, which would have forced a per-bar name lookup; fixed in-branch (`Resolution::ColorConst`)


milocaetano:
## PR #102 — arch-review

Verified by reading `crates/pine/src/compile.rs`, `builtins.rs`, `tests/compile_passes.rs` on `feat/indicators-pine-compile`, plus the stack tip (`feat/indicators-docs-perf`) to check what later PRs already fix. `cargo` was not run (parallel reviews); CI on #102 is green.

### Blockers
- `crates/pine/src/compile.rs:964` — the top-level-only rule is `ctx.frame.is_some() || ctx.in_loop`, which misses `if` blocks. A `plot()` inside a top-level `if` body walks with `frame: None, in_loop: false`, so no error fires; and because the plot registry is built in `collect_declarations` (`compile.rs:401`), which only scans top-level `ExprStmt` nodes, that `plot()` is never registered either — `plot_of_call[id]` stays `None` and the script draws *nothing*, silently. The mirror image is worse: an `input.*` inside an `if` body *is* registered (`input_call` runs from the walk, not from pass 1), so the same nesting is refused for neither and behaves inconsistently for both. TradingView errors on both. Correctness/data honesty is priority 0: a script that cannot do a thing must be told so. Fix: add `top_level: bool` to `WalkCtx`, clear it when entering any `Block`/`If`/`For`/`While` body, and gate the `PINE_SYNTAX` check on `!ctx.top_level`; add a test that `if close > open\n    plot(close)` fails. Still present at the stack tip.
- `crates/pine/src/compile.rs:470` — `plot_call` const-folds `title`/`style`/`color`/`linewidth`, but it runs from pass 1 (`compile.rs:152`), *before* `resolve_and_scan` fills `resolutions`. `fold()` on a `Name` reads `self.resolutions[id.index()]` (`compile.rs:561`), which is still `Resolution::None` for every node at that point, so **no name ever folds in the plot registry**. `c = color.red` + `plot(close, color=c)` silently plots the amber default; `w = 2` + `linewidth=w` silently plots 1.5; `title=myTitle` silently becomes `"plot 0"`. The same applies to the idiom the dialect genuinely cannot support — `plot(x, color = up ? color.green : color.red)` — which is dropped without a word (`PlotSpec` has one `base_color`, and `eval.rs` never re-derives it per bar). Two fixes, both needed: (a) register plots after `resolve_and_scan`, using the same deferred shape the crate later adopts for `fill` (`pending_fills`); (b) when a *supplied* `color`/`style`/`linewidth`/`title` argument fails to fold, push a `PINE_UNSUPPORTED` warning naming the fallback instead of substituting in silence. Note the asymmetry this creates today: `input.*` defaults *do* fold through names, because `input_call` runs inside the walk. Still present at the stack tip.

### Should fix
- `crates/pine/src/compile.rs:837` — dynamic offsets **assign** `MAX_BARS_BACK_CAP` instead of raising the floor. `plot(close[700] + close[i])` ends with `max_bars_back = 500` and the 700-bar read silently returns `na`; write the two reads in the other order and you get 701. The order-dependence proves it unintended, and it contradicts the doc comment three lines up ("constant offsets size storage exactly"). Fix: `self.max_bars_back = self.max_bars_back.max(MAX_BARS_BACK_CAP);`. While there: a folded *negative* constant offset (`close[-1]`) also lands in this branch and is treated as dynamic — it should be a load-time error.
- `crates/pine/src/compile.rs:686` — `reassigned[slot]` is set by the same walk that performs the folding, so a `:=` that appears *after* the call is invisible to `check_kernel_length` (`compile.rs:962`). This compiles clean: `var len = 9` / `plot(ta.sma(close, len))` / `len := len + 1` — a kernel length that changes every bar, which is exactly what `PINE_SERIES_LENGTH` exists to prevent. It is not caught downstream either: `eval.rs` builds the kernel with `or_insert_with(|| Atr::new(len))`, so the first bar's length wins forever and the plot is quietly wrong. The existing test only covers reassign-*before*-use. Fix: pre-scan the arena for `Reassign` names to fill `reassigned` before any fold runs (or run pass 4 as a separate pass after the walk, as §3.4 numbers it).
- `crates/pine/src/compile.rs:554` — the folder's float division has no zero guard, so `1.0 / 0` folds to `Const::Float(inf)`, while the interpreter maps division by zero to `na` as a documented, deliberate divergence from IEEE (`eval.rs:700`). A folded expression must evaluate identically to the interpreted one; here it does not, and the `inf` reaches `InputSpec::Float { default }` and `PlotSpec::width`. (The int/int path is safe — `*b != 0` guards both arms and the fallthrough returns `None`.) Fix: mirror the interpreter — return `Const::Na` when the divisor is zero.
- `crates/pine/src/compile.rs:924` — `rejection_of` is consulted only from the `Call` arm (`compile.rs:891`) and the `Member` arm (`compile.rs:858`), never from the bare-`Name` arm. But the whole calendar family in the reject list (`builtins.rs:437`: `year`, `month`, `dayofmonth`, `dayofweek`, `hour`, `minute`, `second`) and `max_bars_back` are *variables* in Pine, so their common usage reports `PINE_UNKNOWN_NAME` — "unknown name `dayofweek`" — instead of the `PINE_NO_CALENDAR` reason that `docs/pine-dialect.md:176` promises. That branch of the reject list is effectively unreachable, and the doc-drift test only checks the code string appears in the doc, not that a script produces it. Fix: consult `rejection_of` at the top of `unknown_name()` (`compile.rs:1146`) so every path gets the honest family reason; add a compile-level test for a bare calendar name.
- `crates/pine/src/builtins.rs:329` — the accepted-name list is written twice: once as the `lookup` match (`builtins.rs:243`) and once as the `all_names` array. Adding one builtin means editing the enum, `lookup`, `all_names`, and possibly `is_stateful_call` / `is_top_level_only` / `kernel_length_arg` in the other file — the docking test fails here. Worse, the guard at `builtins.rs:495` is named `every_listed_name_resolves_and_back` but only checks `all_names → lookup`; a variant added to the enum and `lookup` but forgotten in `all_names` silently loses its did-you-mean entry *and* escapes the PR 8 doc-drift guard, with no test failing. Fix: one `const BUILTINS: &[(&str, Builtin)]` that both `lookup` and `all_names` derive from — a linear scan over 76 entries at load time costs nothing.
- `crates/pine/tests/compile_passes.rs:252` — two behaviours this PR implements have no test that fails without them: the top-level-only rule (`is_top_level_only`, `builtins.rs:222`) and `PINE_INPUT_NOT_CONST` (`compile.rs:1054`). Add the four cases the findings above name: `plot()` in an `if` body, `plot()` in a user function, `input.int(close)`, and a plot whose `color`/`linewidth` comes from a variable.

### Consider
- `crates/pine/src/compile.rs:467` — `let _ = call;` at the end of `indicator_header`: the parameter is unused. Drop it from the signature rather than suppressing it.
- `crates/pine/src/compile.rs:1041` — the input `name` is derived from the title (`title.to_lowercase().replace(' ', "_")`), but `InputSpec::name` is documented as "Stable identifier (persistence key, script argument name)". Two inputs sharing a title collide, an untitled one becomes `input_0`, and punctuation leaks in (`ema_(fast)`). The declared variable (`len = input.int(...)`) is the identifier Pine users expect. Harmless today only because the app saves input values by index.
- `crates/pine/src/compile.rs:531` — `Const::Int(-v)` and `a % b` (`:543`) are unchecked, sitting between `checked_add` / `checked_sub` / `checked_mul`. `-(-9223372036854775807 - 1)` folds to a debug panic on a user-supplied script file. `v.checked_neg()?` / `checked_rem` / `checked_div` keeps the function uniformly panic-free, matching the repo's overflow-safe-arithmetic precedent.
- `crates/pine/src/compile.rs:63` — `Resolution::EnumConst(&'static str)` is free here, but it hands the interpreter a string tag: the later `eval.rs` turns each reference into `Value::Str(Rc::new(tag.to_owned()))`, an allocation on the per-bar path, in a module whose header promises "zero string lookups at eval time". A small `EnumTag` enum in this file would close that off before PR 7 depends on the string shape.
- Note, no action needed: `check_kernel_length` (`compile.rs:962`) also runs inside function bodies at this commit, where a length parameter cannot fold, so `f(len) => ta.ema(close, len)` is rejected. The stack tip already guards it with `if ctx.frame.is_none()` and re-validates at bar time — flagged only because #102 merges before that fix.

### Verdict
- **Docking**: strong. Two new files, `lib.rs` +3 lines, one workspace dependency added in the declared direction (`pine → indicators`); nothing existing was reshaped to make room. The builtin registry is the right port — one table that name resolution, call-site numbering and the reject scan all consult. The seam is inside it: a new builtin costs 3-6 edits across two files with no compile-time link between the enum, `lookup` and `all_names`.
- **Performance**: load-time only; this PR adds no per-bar code. It allocates five dense `Vec`s of `ast.len()` (~64 B/node, so ~320 KB for a 5 000-node script) once per script load so that the interpreter indexes by `NodeId` instead of hashing strings each bar — the right trade, and the reason `history_series` / `call_sites` / `plot_of_call` exist. `self.kind(id).clone()` per node during the walk and the Levenshtein sweep over 76 names are both cold-path and correctly left readable. Determinism holds where it matters: no `HashMap` anywhere, slot / call-site / history ids come from a single AST-order walk over `Vec`s, and `did_you_mean` breaks ties on the fixed `all_names()` order, so the tables are byte-identical across runs.
- **Proof**: 13 integration tests drive every pass through the public `compile()`. `a_dynamic_kernel_length_is_refused`, `stateful_kernels_inside_loops_are_refused`, `recursion_is_refused` and `history_offsets_size_max_bars_back` would each fail if their pass regressed. Nothing today fails if the top-level-only rule breaks, if a plot's color/width silently defaults, or if the reject list stops reaching bare names — which is precisely where the two Blockers live.

PR #103, opened 2026-08-02
feat(pine): tree-walking interpreter + ScriptIndicator (M2, PR 7/8)

milocaetano:
Third PR of milestone M2 of [docs/indicator-system-plan.md](https://github.com/milocaetano/quantick/blob/main/docs/indicator-system-plan.md). **Stacked on #102** (← #101 ← #99 ← #98 ← #97 ← #96) — merge in order.

## What's here

A compiled `.pine` script now implements the `Indicator` trait — the host (and therefore chart, backtest and bot) cannot tell it from a native indicator.

- **`eval.rs`** — slot-indexed environments, `Value` ≤ 16 bytes (unit-tested), kernels keyed by **(call site, interned call path)**: each textual call owns its state; a user function called from two places owns two (§2.5 locked decision, proven by test: `f() => ta.cum(1.0)` called twice yields `2,4,6…`). Paths intern to integers, so the per-bar walk allocates nothing after its first bar.
- **Expression history** — every `History` node owns a ring of its subject's committed values; previews stage-and-drop, commits flush. `close[1]` can never see a forming bar (tested with a wild 999.0 preview).
- **Preview discipline** — snapshot globals/var-flags/kernels, run, restore; **`varip` slots are exempt** (kept both preview bumps in the test); histories are committed-only so their rollback is free.
- **Semantics locked by 14 tests** — scripted `ta.ema` bit-identical to the native kernel; `var`/plain/`varip`; tuples through functions; `if`/`for` as expressions; `PINE_LOOP_BUDGET` instead of a worker freeze; division by zero = na (an infinite line on a chart is a lie); runtime errors render `line:col … (CODE)`.
- **`ta::PivotHigh/PivotLow`** — pulled forward from M3 (the registry and corpus already accept `ta.pivothigh`); confirmation semantics per plan §2.5, strict inequality, NaN-deferred.
- **Compile adjustments** — `hline` registers a plot column; `plotshape`/`fill`/`bgcolor`/`barcolor` accepted-but-inert with honest warnings until M3; `fixnan` is stateful; kernel lengths inside function bodies validate at bar time (parameters cannot fold at load).

## Bench (`cargo bench -p quantick-pine`, §5 budget ≤ 50 µs/commit)

| Script | AST nodes | Measured |
|---|---|---|
| `ema.pine` (input + 1 kernel + plot) | 19 | **1.83 µs/bar** |
| flow script (3 kernels, if-expr, 4 plots) | 61 | **5.92 µs/bar** |

The escalation path (constant-fold harder → bytecode) stays unused, as the plan hoped.

## Verification

- [x] fmt --check / clippy -D warnings / build / test --workspace (61 pine tests + 75 indicators tests)
- [x] arch-review: clean (pivot pull-forward declared above)


milocaetano:
## PR #103 — arch-review

### Blockers

- `crates/pine/src/eval.rs:366` (and `:288`) — `match self.kind(id).clone()` clones the whole `NodeKind` **once per evaluated node, per bar**, then throws it away. The variants that carry heap data are exactly the common ones: `Name(String)`, `Declare{name}`, `Reassign{name}`, `Member{field}`, `Str(String)`, `Block{Vec<NodeId>}`, `Tuple{Vec<NodeId>}`, `Call{Vec<Arg>}` (and `Arg.name: Option<String>` inside it). `:481` makes it worse: the `Call` arm binds `args: _` — discarding the `Vec<Arg>` the line-366 clone just allocated — and then re-clones `args` at `:483`, so every call node pays **two** `Vec<Arg>` allocations plus a `String` per named argument. `:369` compounds it again (`Value::Str(Rc::new(v))` = the cloned `String` plus a fresh `Rc<String>` per string literal per bar). This directly contradicts the module doc at `eval.rs:7` and the PR body ("the per-bar walk allocates nothing after its first bar") — the walk allocates on most nodes. Counting the node kinds in the PR's own `FLOW_SCRIPT` (5 `Declare`, ~12 `Name`, 3 `Member`, 4 `Call`, 1 `Block`) puts it on the order of 25 malloc/free pairs per bar; that is a count off the AST, not an instrumented run, and `cargo bench -p quantick-pine` before/after is the measurement that settles it — the reported 5.92 µs / 61 nodes ≈ 97 ns per node is consistent with a malloc dominating each node visit. Priority order puts performance first and this is the one path the plan gives a budget. **Fix**: `self.script` is `&'a CompiledScript` (Copy), so change `fn kind(&self, id) -> &NodeKind` to `fn kind(&self, id: NodeId) -> &'a NodeKind { &self.script.ast.node(id).kind }`. The returned borrow is independent of `&mut self`, so `match self.kind(id) { ... }` compiles with `self.eval(...)` inside the arms, and every `.clone()` at `:245`, `:288`, `:366`, `:483` disappears. Pre-intern string literals into a compile-side `Vec<Rc<String>>` indexed by node so `:369` becomes a refcount bump.

- `crates/pine/src/eval.rs:861` — the `length` closure accepts **any** finite integral `v >= 1.0`; `Const::as_positive_len` in `compile.rs:185` has no upper bound either. So `ta.sma(close, 2000000000)` compiles clean and then runs `Window::new(2_000_000_000)` → `vec![0.0; 2e9]` = 16 GB. A `Vec` allocation failure calls `handle_alloc_error`, which **aborts** — it does not unwind, and there is no `catch_unwind` anywhere in the workspace. That breaks the contract `indicators/src/indicator.rs` promises: "a runtime failure … puts this indicator in an error state; the host stops evaluating it". A user typing a wrong length kills the process, not the indicator. `:947` is the sharper case — `ta.pivothigh`/`ta.pivotlow` are **absent from `Compiler::kernel_length_arg`**, so they get no load-time check at all, and `PivotCore::new` (`pivot.rs:28`) computes `left + 1 + right`; with `ta.pivothigh(high, 1e19, 1e19)` the `v as usize` casts saturate to `usize::MAX` and the sum overflows — panic in debug, silent wrap in release (the workspace `Cargo.toml` sets no `[profile.release] overflow-checks`). This is the repo's own "feed arithmetic must saturate, never panic" rule applied to the script path. **Fix**: name a `pub const MAX_KERNEL_LENGTH: usize` next to the existing `MAX_BARS_BACK_CAP: usize = 500` precedent, enforce it in both `check_kernel_length` (load-time `PINE_SERIES_LENGTH` error) and the `length` closure, and add `TaPivotHigh`/`TaPivotLow` to `kernel_length_arg` with their two length positions.

### Should fix

- `crates/pine/src/eval.rs:966` / `:972` — the length is read and validated every bar but only *used* on the bar that first creates the kernel: `.entry(key).or_insert_with(|| Kernel::Sma(ta::Sma::new(len)))` drops `len` on every later bar. So `f(n) => ta.sma(close, n)` called with a changing `n` silently keeps bar 0's window forever, and the plot's title says one thing while the maths does another. Pine rejects a series length outright; here it is silently pinned. That is a data-honesty violation, the rule that outranks everything but determinism. **Fix**: store the length alongside the kernel in the `Kernel` entry and return the existing `ErrorCode::PineSeriesLength` when a later bar's length differs from the instantiated one.

- `crates/pine/src/eval.rs:179` — `ring.values.remove(0)` shifts the whole ring left on every commit once it is full. `max_bars_back` is capped at `MAX_BARS_BACK_CAP = 500`, so that is up to 500 × 16 bytes = 8 KB memmoved **per `History` node per closed bar**; a script with five `[n]` reads moves 40 KB per bar, and tick/volume bars close many times a second under a dense tape. The repo already has the answer one crate over: `ta::Window` (`indicators/src/ta/mod.rs`) is a fixed `Vec` with a `head` index and O(1) push — inventing a second, worse bounded ring is also a standardisation finding. **Fix**: give `HistRing` a `head: usize` and overwrite in place, or use `VecDeque::pop_front`.

- `crates/pine/src/eval.rs:234`, `:245`, `:672` — three more per-bar allocations under the same "allocates nothing" claim. `locals: Vec::new()` builds a fresh frame stack in `Eval::new`, i.e. once per bar: the first `user_call` `resize`s it, so any script with a user function pays a malloc/free every bar and the doc's "one flat reused frame stack" is only true *within* a bar. `:245` clones the top-level `Vec<NodeId>` every bar (dies with the Blocker's `kind()` fix). `:672` clones the whole `FunctionInfo` — including `name: String`, which the struct's own doc says is "diagnostics only; calls are index-resolved" — on **every user-function call, every bar**. **Fix**: hoist `locals` into `ScriptState` so its capacity persists across bars (`self.locals.clear()` in `run`); at `:672` read the three `Copy` fields (`param_count`, `local_slots`, `body`) out of `&self.script.functions[index]` instead of cloning, and take the name by reference only on the arity-error path.

- `crates/indicators/src/ta/pivot.rs:51` — `PivotCore::push` scans the entire `left + 1 + right` window on every push. `ta/mod.rs:5` — a file this PR edits — states as the module's contract: "**No kernel re-scans its window on push** … `highest`/`lowest` use a monotonic deque. This is what makes the per-bar commit budget (§5) achievable." The new kernel is the first to break that sentence, and the cost is user-controlled: `ta.pivothigh(high, 250, 250)` is 501 comparisons per bar per script, and combined with the unbounded length above it is arbitrarily worse. **Fix**: either build it on the monotonic-deque machinery `Highest`/`Lowest` already use (a strict pivot is "the candidate is the unique max of the window"), or amend `ta/mod.rs:5` to name pivot as the documented exception with its O(left+right) cost — the second option is free at runtime and is the minimum this PR owes.

- `crates/pine/src/compile.rs:434` — the warning string reads `"…(shape plots and bar                              colors land with the drawing milestone)"`: ~30 spaces mid-sentence, from a wrapped line that lost its `\`. This is user-visible — `ScriptIndicator::warnings()` surfaces it in the indicator panel. It also survives to the top of the stack in mutated form (`"per-bar color                              columns…"` at the same site), so no later PR cleans it up. **Fix**: collapse to a single space.

- `crates/pine/benches/interp.rs:54` — the bench loop calls only `on_close`, so the declared performance evidence covers the *lower*-rate path. `preview` runs once per drained partial-update batch in the worker (`app/src/indicator_worker.rs:412`, fed per trade from `app.rs:1421`) — well above bar-close rate — and it is the path that additionally pays `ScriptState::snapshot` (`script.rs:143`): two `Vec` clones plus a deep clone of the whole `kernels` `BTreeMap`, each windowed kernel dragging its ring `Vec` with it. Neither benched script has a user function or a `[n]` history read either, so the `locals` and `FunctionInfo` allocations and the `remove(0)` shift above are all unmeasured. **Fix**: add a third bench arm that interleaves N previews per commit (the `kernel_previews_roll_back_like_natives` pattern) over a script with a function and a history read, and put that number in the PR body next to the commit number.

- `crates/pine/src/eval.rs:1077-1090` — `math.max`/`math.min`/`math.avg` allocate a `Vec` per call per bar; `math.max(volume, 1)` is in this PR's own `FLOW_SCRIPT`, so it is on the benched hot path. The fold identities are also wrong at the edges: `f64::MIN` / `f64::MAX` are the extreme *finite* values, so `math.max(math.log(0))` returns `-1.797e308` instead of `-inf`. **Fix**: single streaming pass — track min/max/sum and a `saw_nan` flag while evaluating the args, no `Vec`; use `f64::NEG_INFINITY` / `f64::INFINITY` as identities.

### Consider

- `crates/pine/src/eval.rs:779` — every kernel call does a `BTreeMap<(u32,u32), _>` lookup keyed by `(site, path)`. `BTreeMap` is the right call for determinism, but the overwhelmingly common case is `path == 0` (no user function), where the compile-assigned `site` index is already dense. A `Vec<Option<Kernel>>` indexed by `site` for path 0, falling back to the `BTreeMap` for deeper paths, removes the key comparisons from the hot path without touching iteration-order determinism. Same shape applies to `histories` / `staged_histories`.
- `crates/pine/src/script.rs:143` — `snapshot()` allocates three fresh containers per preview (`globals`, `var_done`, and the `kernels` map with a deep clone of every kernel ring). Cloning kernels for preview is the repo's established discipline (`native/mod.rs` does it too), so this is not a new pattern — but a snapshot buffer owned by `ScriptIndicator` and refilled with `clone_from` would reuse the two `Vec` allocations for free.
- `crates/pine/src/eval.rs:557` — `barstate.islast` is implemented as `!is_commit`, i.e. "we are inside a preview". In Pine it is also true on the last *historical* bar. A script gating on it behaves differently during a `rebuild` replay than on the live bar. The line right above documents the `LastBarIndex` divergence honestly; this one gets no comment. One comment line, zero runtime cost.
- `crates/pine/src/compile.rs:489` — `plot_call` is reused for `hline` via `is_hline`, but the two builtins have different positional signatures: `plot(series, title, color, linewidth, style, …)` vs `hline(price, title, color, linestyle, linewidth, …)`. Positions 0-2 line up; position 3 does not, so `hline(50, "mid", color.red, hline.style_dashed, 2)` reads the linestyle enum as a width (folds to `None`, defaults) and drops the user's `2` entirely. Named arguments work. **Fix**: read width from position 4 when `is_hline`.
- No test replays the same bars through `reset()` and diffs the output, which is the exact path `IndicatorHost::rebuild` takes on every spec switch and replay seek. `kernel_previews_roll_back_like_natives` covers cross-instance determinism and would catch most regressions, so this is a gap rather than an absence — a five-line addition to `script_semantics.rs` closes it.

### Verdict

- **Docking**: clean. The interpreter enters the system entirely through `Indicator`, `ScriptIndicator` adds one file, and the host, worker and future backtest/bot see no new type. Adding a `ta.*` builtin means one `Kernel` variant plus one dispatch arm — the `match` over `Builtin` in `kernel_call` grows, but that is the compiler's own dispatch table, not a feature type-switch. The one existing-file edit outside the crate is `ta/mod.rs` (+2 registration lines) — the docking test passes. The dependency direction holds: `pine → indicators`, no reverse edge.
- **Performance**: the commit path is inside budget as measured (1.83 / 5.92 µs against ≤ 50 µs), so nothing here is shipping broken — but the interpreter allocates on most evaluated nodes, once more per bar for the frame stack and per call for `FunctionInfo`, and does an O(500) memmove per history ring per commit. The doc and PR body claim the opposite; that claim is the finding as much as the cost is. The higher-rate path (`preview`) is not benched at all, and an unbounded kernel length can abort the process rather than error the indicator.
- **Proof**: 14 semantics tests, each targeting a locked decision, and they are the right ones — `each_textual_call_site_owns_its_state_through_call_paths` fails if path interning regresses, `previews_do_not_commit_history` fails if a forming bar reaches a ring, `varip_survives_previews_var_rolls_back` fails if `restore` stops exempting `varip`, `the_loop_budget_errors_instead_of_hanging` fails if a runaway loop stops erroring. What has no failing test: a kernel length that changes between bars, a length large enough to abort, and `reset()` → replay determinism.

PR #104, opened 2026-08-02
feat(app): script library — load .pine indicators (M2, PR 8/8)

milocaetano:
Final PR of milestone M2 of [docs/indicator-system-plan.md](https://github.com/milocaetano/quantick/blob/main/docs/indicator-system-plan.md). **Stacked on #103** (← #102 ← #101 ← #99 ← #98 ← #97 ← #96) — merge in order. **This completes M2: `.pine` scripts load, compile and plot through the same worker/render path as native indicators.**

## What's here

- **`indicators/library.rs`** — one list, two origins: embedded starter scripts (`include_str!`) so the feature works with an empty folder, plus a scan of the indicators dir (`QUANTICK_INDICATORS_DIR`, default `./indicators`, created on first run). A file shadows an embedded script of the same name. Hot reload = M4.
- **`IndicatorSource::Script { name, text }`** — the worker compiles and instantiates; a script that fails to compile becomes an **error-carrying slot**: the menu entry shows the red dot and the full `file:line:col … (CODE)` rendering of every problem. Never a silent drop.
- **M2 acceptance, as a test** — a script with `request.security` loads into a slot whose error reads `sec.pine:3: … (PINE_NO_SECURITY) … activity-sampled`; an embedded starter script plots end-to-end through the worker (title, overlay flag, warmup NaN, full columns).
- **INDICATORS menu** — new *scripts* section listing embedded + folder scripts.
- **`docs/pine-dialect.md`** — the user-facing reference: execution model, grammar, our exact kernel semantics (EMA seed = SMA, warmup NaN, NaN policies), the flow builtins, documented divergences, reject list and the full error-code catalog. A **doc-drift test** asserts every registered builtin and error code appears in the doc.
- Embedded starters: `ema.pine`, `cvd.pine`, `delta_histogram.pine`.

## Verification

- [x] fmt --check / clippy -D warnings / build / test --workspace (38 suites; new: 2 library, 2 worker script-load, 2 doc-drift)
- [x] arch-review: clean (script loading docks as one new `IndicatorSource` variant + one menu section; no existing behavior edited)


milocaetano:
## PR #104 — arch-review

### Blockers

- `crates/app/src/app.rs:790` — the read-failure branch of `add_script_indicator` only emits `tracing::warn!(INDICATOR_SCRIPT_UNREADABLE)`; the UI shows nothing at all. The click produces no slot, no error card, no message — while the doc comment three lines up (`crates/app/src/app.rs:778-780`) promises "a file that no longer reads … becomes the slot's error — shown with lines and codes, never silently dropped". The compile half of that promise is implemented and tested; the read half is not. This is reachable without any exotic setup: the scan does not check `is_file()`, so a *directory* named `foo.pine` is listed in the menu and fails `read_to_string`; likewise a file deleted, renamed or locked between the startup scan and the click. Data honesty (CLAUDE.md) is the rule at stake, and there is no test on this branch of the match. Checked the top of the stack (`crates/app/src/app.rs:915` on `feat/indicators-docs-perf`): still a bare `warn!`, so no later PR fixes it. Fix: reuse the error-slot mechanism this very PR built — allocate a slot and apply the same pair the worker sends on a compile failure, `IndicatorViews::apply(IndicatorEvent::Rebuilt { descriptor titled with the script name, empty plots and columns })` then `apply(IndicatorEvent::Error { EvalError { bar_index: 0, message } })` (both `pub(crate)`, `crates/app/src/indicators/mod.rs:82`), and add a test whose `ScriptOrigin::File` points at a path that does not exist.

### Should fix

- `crates/app/src/indicators/library.rs:60` — `scan()` takes no argument and resolves `QUANTICK_INDICATORS_DIR` internally, so the entire *file* half of the library is untestable without mutating process-global env, and in fact it is untested: the two tests here (`:132`, `:143`) only cover the embedded array. Nothing proves that a `.pine` in the folder appears, that the sort at `:93` gives a deterministic menu, that the shadowing rule at `:101` lets a user file win, or that `read` surfaces an I/O error. The repo has already solved this exact shape: `quantick_replay::library::scan(root: &Path)` (`crates/replay/src/library.rs:221`) takes the folder and lets the caller resolve the env var, and its tests drive it with a `Scratch` temp-dir helper (`crates/replay/src/library.rs:421`). Fix: split `fn scan_dir(dir: &Path) -> Vec<ScriptEntry>` out of `scan()`, keep env resolution in `scan()`, and add the three tests above against a scratch folder.
- `crates/app/src/indicators/library.rs:88` — three ways a file quietly disappears from the menu, in one expression. (a) `ext == "pine"` is a case-sensitive `OsStr` comparison, so on the primary dev platform `EMA.PINE` or `MyScript.Pine` never shows up; the repo's own scan lowercases first (`crates/replay/src/library.rs:315-317`). (b) There is no `file_type().is_file()` check, so a directory named `x.pine` becomes a menu entry that can only fail (feeds the blocker above); the replay scan checks it at `crates/replay/src/library.rs:305`. (c) `.unwrap_or_default()` at `:91` swallows a `read_dir` error into an empty list with no log, while its sibling `create_dir_all` failure at `:72` logs `INDICATOR_DIR_UNAVAILABLE` — inconsistent within eight lines. Fix: lowercase the extension, skip non-files, and log an `INDICATOR_DIR_UNREADABLE` warn on the `read_dir` error.
- `crates/pine/tests/dialect_doc.rs:25` — the doc-drift guard hand-lists all 18 `ErrorCode` variants in an array. The list is complete today (diffed against `crates/pine/src/error.rs:51-89`), but a variant added tomorrow is simply not checked — the test reintroduces the habit its own header comment says it exists to replace ("turns 'update the docs' from a habit into a build rule"). The test directly above it does it right, via `Builtin::all_names()` (`crates/pine/src/builtins.rs:342`). Fix: add `pub const ALL: &[ErrorCode]` next to `as_str` (`crates/pine/src/error.rs:94`) and iterate that, so a new variant fails the doc test until it is documented.

### Consider

- `crates/app/src/app.rs:727` — the script-name list is rebuilt every frame: one `Vec` plus one `String` allocation per entry (3 embedded + N files), at ~60 Hz, whether or not the INDICATORS menu is open. It matches the neighbouring `feeds` / `symbols` / `indicators` fields, so it is the existing pattern's cost rather than a regression, and no frame-time effect was measured — the perf HUD would settle whether it registers. Worth singling out because it is the one list that cannot change after startup in this PR (scan is startup-only), so it is free to fix: cache `names: Vec<String>` in `ScriptLibrary` at scan time and give `ToolbarModel` a `scripts: &'a [String]`.
- `crates/app/src/indicator_worker.rs:60` — `.join(…)` is written with a literal newline inside the string literal, which rustfmt then spreads over four lines. `"\n"` is the one-line, unambiguous form used everywhere else.
- `crates/app/src/indicator_worker.rs:551` — the end-to-end test indexes `EMBEDDED_SCRIPTS[0]` and then asserts `title == "EMA"` and `overlay`. Adding a starter script at the front of the array breaks this test with a message that reads like the EMA script broke. Look it up by name instead.
- `crates/app/src/indicators/library.rs:21` — "Each doubles as a conformance fixture in the pine crate's corpus" is not the case: nothing under `crates/pine` references `crates/app/scripts`. The compile check lives in this module's own test at `:132`. Point the comment there.
- `crates/app/src/indicators/library.rs:72` — `create_dir_all` at startup writes `./indicators` relative to whatever CWD the app was launched from, and `/indicators` is absent from `.gitignore` — so every developer who runs the app from the repo root gets an untracked folder in `git status`. The repo already ignores `/quantick.toml`, the analogous user-owned file (`.gitignore:6`). Adding `/indicators` there is the whole fix; the directory creation itself is well justified and activates nothing on its own.

### Verdict

- **Docking**: clean. A third origin (a bundled pack, a URL) docks by adding one `ScriptOrigin` variant plus one arm in `read` — no other file changes. `IndicatorSource` grew one variant and, in doing so, made the load-error path shared by *every* source rather than script-specific, which is the right direction. `SlotMirror.host_id: Option<InstanceId>` is the honest way to say "this slot never loaded", and `publish_deltas` skipping such slots (`indicator_worker.rs:304`) means an error slot costs zero per bar. The positional index the toolbar uses as script identity looked fragile, but it holds: the library is scanned once per process even at the top of the stack, and M4 hot reload builds on the index without reopening `toolbar.rs`. Nothing activates itself — the scan only fills a menu.
- **Performance**: flat where it matters. Per-bar and per-trade paths are untouched; compilation runs on the worker thread, never the render thread. Startup gains one `create_dir_all` + one shallow `read_dir`. The click-time `read_to_string` is on the UI thread — microseconds for a local file, though a `QUANTICK_INDICATORS_DIR` pointing at a network share would stall the frame that handles the click. The one new per-frame cost is the name-list clone above, ~4 small allocations per frame today.
- **Proof**: `a_rejected_script_surfaces_its_error_with_line_and_code` (`crates/app/src/indicator_worker.rs:516`) fails without the error-slot path — it is the M2 acceptance and it asserts file:line and `PINE_NO_SECURITY`, not just "an error happened". `an_embedded_script_plots_through_the_worker` (`:550`) fails if the scripted pipe stops producing columns. Nothing fails if the file-scan half regresses — that is the second Should-fix, and it is also the gap that would have caught the blocker. CI on #104 is green — run `30748298925`, 2m48s.

PR #105, opened 2026-08-02
feat(indicators,pine,app): draw objects — lines, boxes, labels (M3, PR 9/10)

milocaetano:
First PR of milestone M3 of [docs/indicator-system-plan.md](https://github.com/milocaetano/quantick/blob/main/docs/indicator-system-plan.md). **Stacked on #104** (← #103 ← #102 ← #101 ← #99 ← #98 ← #97 ← #96) — merge in order.

## What's here

- **`indicators::objects`** (§2.6) — `ObjectStore` with the **500-per-kind hard cap** (the 501st collects the oldest: bounded render + clone cost by construction), never-reused ids (stale handles miss, never mutate a stranger), a revision counter so publication needs no diffing.
- **Preview-transient objects** — the forming bar's set rides the `PreviewFrame` and replaces the committed view while the bar forms; on rollback it vanishes. Locked by tests at the script level (a preview label/stretch never reaches the committed store) and the worker level.
- **Language**: `line.new`/`box.new`/`label.new`, `Value::Obj` handles, `handle.set_*()`/`.delete()` method dispatch, `label.style_*` constants. Fixed en route: `na(x)` with the keyword callee now resolves (it slipped through every earlier test).
- **Render** — objects share the overlays' paint slot (after candles, before aggression bubbles), on the price chart and inside panes, riding the candles' own x-mapping so they pan/zoom with the bars.
- **Starter pack + corpus**: `zigzag.pine` (pivots → orange segments + HH/LH/HL/LL labels) and `range_box.pine` (a box stretched over `highest`/`lowest` each bar). *Declared swap vs the plan: these fixtures moved up from PR 10, which now carries only the shape plots.*
- `QUANTICK_INDICATOR_SCRIPTS_AUTOSTART=zigzag.pine,range_box.pine` — scripted validation runs need no clicks.

## Visual validation (M3 acceptance)

Live Binance BTCUSDT, tick(50), 60 fps: the orange zigzag segment connects two confirmed pivots, both carrying **HH** pill labels; the aqua range box hugs highest/lowest of the last 20 bars and **moved with the market between captures** (per-bar `set_lefttop`/`set_rightbottom` mutations working live). A worker test additionally proves rebuild reproduces the identical object set (seek/spec-switch determinism).

## Verification

- [x] fmt --check / clippy -D warnings / build / test --workspace (38 suites; new: 4 store unit, 2 script-semantics object tests, 1 worker zigzag end-to-end)
- [x] arch-review: clean (trait accessor default keeps object-free indicators cost-free; events publish only on revision change)


milocaetano:
## PR #105 — arch-review

### Blockers
- none

### Should fix

- `crates/app/src/indicator_render.rs:399` — `draw_objects` is the only draw function in this file that does **not** take `start`/`end`. Every sibling (`draw_view_plots`, `draw_line`, `draw_bars`, `draw_markers`, `draw_area`) slices to the visible range via `VisiblePlot { start, end }`; this one walks all three retained kinds in full. At the cap that is up to 1500 objects visited per view per frame at ~60 Hz, for a chart showing maybe 200 slots. Worst of it is `indicator_render.rs:439`: `painter.layout_no_wrap(label.text.clone(), …)` runs *before* any visibility test, so a mature `zigzag.pine` pays 500 `String` allocations plus 500 `LayoutJob` builds and galley-cache hashes every frame — ~30k allocations/sec on the render thread — for labels that are mostly off-screen. egui's coarse culling saves the tessellation but not the layout, because the layout happens in our code. Honest magnitude: unmeasured; at ~500 short labels the estimate is well under a millisecond per frame, so this is a slope problem rather than a stall today. The frame-time HUD with `zigzag.pine` running long enough to hit the cap would settle it. Fix: give `draw_objects` the `start`/`end` it is already handed at both call sites and cull first — **overlap**, not endpoint containment (`line.x1.max(line.x2) < start || line.x1.min(line.x2) >= end`, same for `box.left`/`right`), and for labels an anchor test widened by half a pill.

- `crates/indicators/src/objects.rs:160` (and `:170`, `:180`, `:190`, `:196`, `:202`) — `line_mut`/`box_mut`/`label_mut` bump `revision` *before* the lookup, and `delete_*` bumps it whether or not `retain` removed anything. So a handle that the cap already garbage-collected — the documented, expected outcome once a script passes 500 of a kind — keeps moving the revision on every bar. `indicator_worker.rs:369` reads that as "the set changed" and publishes a full `ObjectSnapshot` down the channel every bar: a deep clone of up to 1500 objects (each label re-allocating its `String`) for zero visible change, forever. Bar-rate, bounded, but entirely avoidable. Fix: bump inside the success path — `self.lines.iter_mut().rev().find(…).inspect(|_| self.revision += 1)` won't borrow-check, so take the index first, then `self.revision += 1` only when it is `Some`; for the deletes, compare `len()` before and after.

- `crates/pine/src/eval.rs:199` — `restore` assigns the whole `ObjectStore` back, which rolls `next_id` (`objects.rs:103`) back with it, while `restore` deliberately exempts `varip` slots two lines above. A `varip` handle therefore survives a preview rollback pointing at an id the store will hand out again. Repro:
  ```pine
  varip line a = na
  if na(a)
      a := line.new(0, close, 0, close)   // only the preview run takes this branch
  b = line.new(bar_index, low, bar_index, high)
  a.set_y1(999)                            // mutates b's line
  ```
  Preview allocates id 0 to `a`; rollback resets `next_id` to 0; the commit run skips the `na(a)` branch and gives id 0 to `b`; `a.set_y1` now finds `b`. That is precisely what the doc comment at `objects.rs:89` promises cannot happen ("a stale handle misses instead of mutating a stranger"). Not a Blocker: `varip` is preview-cadence-dependent by design (`script_semantics.rs:126` locks that in), and the committed-only rebuild path stays deterministic and is tested. But the invariant is load-bearing and the fix is free — add `ObjectStore::restore_from(&mut self, previous: Self)` that assigns the vecs and keeps `next_id = self.next_id.max(previous.next_id)`, and call that instead of the bare assignment. Add a `varip` + object test; nothing currently crosses those two features.

- `crates/pine/src/eval.rs:1172` — a non-finite x becomes `0`: `Ok(if v.is_finite() { v.round() as i64 } else { 0 })`. `label.new(na, price)` then draws a real label anchored at bar 0 instead of not drawing. Meanwhile `arg_price` (`eval.rs:1186`) returns `f64::NAN` for the same case and the renderer honestly skips it (`indicator_render.rs:405`, `:427`, `:435`). So x and y disagree about what "missing" means, and the x side silently patches it — the data-honesty rule. Fix: make the two agree. Either `x: Option<i64>` in `LineObj`/`BoxObj`/`LabelObj`, or map non-finite to a named `const OFF_CHART_BAR: i64 = -1` that the existing `x < 0` guards already drop — the second is a one-line change and needs no renderer edit.

- `crates/app/src/indicator_render.rs:449`, `:451`, `:458`, `:419`, `:431` — the label pill's geometry is inline literals while the two constants right above it (`OBJECT_LABEL_FONT`, `OBJECT_LABEL_PAD`) show the convention: the `2.0` anchor gap appears twice, `Rounding::same(3.0)` once, and `.max(0.5)` as a minimum stroke twice. Name them (`OBJECT_LABEL_GAP_PX`, `OBJECT_LABEL_ROUNDING_PX`, `MIN_OBJECT_STROKE_PX`). While you are there: the `match label.style` block is the first non-trivial geometry in this file, and its comment (`indicator_render.rs:444-446`) opens by admitting the mapping reads backwards — that is the signal to extract it as `fn label_rect(anchor: Pos2, size: Vec2, style: LabelStyle) -> Rect`, per this file's own module doc ("geometry decisions … live in small pure helpers; egui only receives finished point lists"). In fairness this file has no `#[cfg(test)]` module today, so the missing test is not a regression this PR introduced — but the helper is what would let `Up`/`Down` placement be asserted once instead of eyeballed in a screenshot.

### Consider

- `crates/pine/src/script.rs:150` and `:155` — a preview run deep-clones the object store twice: once into `StateSnapshot` for rollback, once into the `PreviewFrame`. The first is genuinely required; the second is not, because `restore` overwrites the store on the very next line. Building the frame's `ObjectSnapshot` by draining the post-run store would move the label `String`s instead of re-allocating them, halving the per-preview cost on a path that runs at worker-batch cadence. Object-free scripts are unaffected — three empty `Vec`s clone without allocating.
- `crates/indicators/src/objects.rs:226` — `counts() -> (usize, usize, usize)` is a positional 3-tuple whose shape encodes "there are exactly three kinds"; adding a fourth breaks it and its only caller, `script.rs:86`. Since that caller only asks "is this store non-empty", `pub fn is_empty(&self) -> bool` on `ObjectStore` would serve it and survive a fourth kind untouched. (`ObjectSnapshot::is_empty` at `objects.rs:248` already exists — and currently has no caller, as does `ObjectStore::clear` at `:207`, whose doc says "indicator reset" but `ScriptIndicator::reset` rebuilds the whole `ScriptState` instead. One of those two should go.)
- `crates/app/src/app.rs:2309` — the chart path draws *all* overlays' plots, then *all* overlays' objects, in a second loop over the same iterator; the pane path (`indicator_render.rs:141-142`) draws plots-then-objects per view. Two orderings for one thing, and the chart one puts indicator B's plots under indicator A's objects. Moving the `draw_objects` call inside `draw_overlays`' existing `for view in overlays` loop deletes the duplicate iteration and makes both paths identical.
- `crates/pine/src/compile.rs:956` — once the base of a dotted callee resolves to a variable, the method name is accepted unchecked; `zone.set_leftop(…)` (typo) compiles clean and only errors at `eval.rs:1449` if that line ever executes. Given the dialect is deliberately dynamically typed (`parser.rs:195`: type annotations "recorded nowhere"), per-kind dispatch at compile time is off the table — but the *name* is checkable against the union of all known object methods at load, which is where this repo surfaces every other script problem with a code and a line.
- `crates/pine/src/compile.rs:907` — the `na(x)` resolution fix (a real bug in the pine crate, per the PR body) has no test inside `crates/pine`; `corpus.rs` only lexes and parses, so `corpus/ok/zigzag.pine` does not exercise it. It is covered — `indicator_worker.rs:614` fails without it — but by a downstream crate. A three-line case in `compile_passes.rs` puts the guard next to the fix.

### Verdict

- **Docking**: no. A fourth kind (polyline) forces edits to six existing files and adds none: `objects.rs` (struct, store field, `new_`/`_mut`/`delete_`, `snapshot`, `counts`, `ObjectSnapshot`, `is_empty`), `builtins.rs` (enum + `lookup` + `NAMES`), `compile.rs:969` (the `"line" | "box" | "label"` namespace allowlist), `eval.rs` (`ObjKind`, the builtin arm, a `_new`, the `method_call` match), `script.rs:86` (tuple destructuring), `indicator_render.rs` (a fourth loop). To be fair, most of that is inherent: Pine's cap is *per kind*, so per-kind storage is forced, and the method table is a language surface where a `match` is the honest implementation. The gratuitous parts are the two above — `counts()`'s fixed arity and the renderer's three hand-written loops where one ordered iterator over a kind-tagged view would do.
- **Performance**: flat for every existing indicator — `Indicator::objects()` defaults to `None` (`indicator.rs:141`), `draws_objects()` short-circuits, and cloning an empty store allocates nothing, so the claim in the PR body holds. For object-drawing scripts: **worker** gains two deep store clones per preview run and one bounded snapshot + channel send per revision change (bar rate, or every bar once a stale handle is in play — see finding 2). **Renderer** gains work proportional to *retained* objects rather than *visible* ones at ~60 Hz, with an allocation per label per frame. Growth is genuinely bounded: `MAX_OBJECTS_PER_KIND = 500` is a named constant enforced at creation for all three kinds, and the `Vec::remove(0)` that enforces it is a ~24 KB memmove at bar rate — checked, and not worth changing.
- **Proof**: `objects.rs::the_cap_collects_the_oldest` and `clones_are_bounded_by_the_caps` guard the cap; `script_semantics.rs::the_object_cap_is_enforced_through_scripts` guards it through the language; `script_semantics.rs::draw_objects_commit_mutate_and_respect_previews` fails if preview objects ever leak into the committed store; `indicator_worker.rs::zigzag_objects_flow_to_the_views_and_survive_rebuild` fails if the event plumbing or rebuild determinism regresses. Real gaps: nothing covers `varip` + objects (finding 3), and nothing covers the renderer's geometry or the revision-on-a-miss republish.

PR #106, opened 2026-08-02
feat(indicators,pine,app): shape plots and fills (M3, PR 10/10)

milocaetano:
Final PR of milestone M3 of [docs/indicator-system-plan.md](https://github.com/milocaetano/quantick/blob/main/docs/indicator-system-plan.md). **Stacked on #105** (← #104 … ← #96) — merge in order. **This completes M3.**

## What's here

- **`plotshape`/`plotchar` are real** — `PlotSpec` gains `marker: Option<MarkerSpec>`: marker columns (triangle up/down, circle, label up/down, cross) anchored `abovebar`/`belowbar` (via a bar-extents closure from the chart — markers ride the bar's high/low) or `absolute` (the cell value is the y). `plotchar`'s character rides as marker text. Cells are na (nothing) or a value, so warmup and quiet bars stay honestly empty.
- **`fill(p1, p2, color)` is real** — resolved at **load time** through plot handles; `IndicatorDescriptor` gains `fills`. Bands render as one convex quad per adjacent pair, behind the plots. A fill between non-plots is refused with an honest message.
- **Bug found by the new test**: a `plot()` assigned to a variable (`a = plot(...)`) was never registered as a column — declare-RHS output calls now collect exactly like bare statements.
- `bgcolor`/`barcolor` stay accepted-but-inert (per-bar color columns are their own milestone), message updated.
- Corpus: `shapes_and_fill.pine`; compile tests for marker registration + fill resolution + the refusal; semantics test for staged condition cells; dialect doc updated.

## Verification

- [x] fmt --check / clippy -D warnings / build / test --workspace (38 suites)
- [x] arch-review: clean (marker/fill dock as optional fields defaulting to today's behavior; no existing render path changed)


milocaetano:
## PR #106 — arch-review

### Blockers
- none

### Should fix

- `crates/app/src/indicator_render.rs:211` — `value_range` folds **every** column into the pane's auto-fit range, including the new marker columns. A pane script (`indicator("t")` with no `overlay=true`, exactly the shape used by the new `script_semantics.rs` test) that mixes `plotshape(cond)` with `plot(cvd)` now stages `1.0` into a column `value_range` reads: a CVD pane spanning `1e6..1.1e6` becomes `1.0..1.1e6` and the real plot collapses into a sliver at the top. The staged `1.0` is a flag, not a value — it is never even used as a y, because non-absolute markers anchor to `pane_extents` (`:155`). Priority order puts correctness of what is drawn above everything else here. Fix: skip index `i` when `view.descriptor.plots[i].marker` is `Some(m)` with `m.location != MarkerLocation::Absolute`, and add the first `#[cfg(test)]` case in this module over a two-column `IndicatorView` (marker column + value column) — that test fails today.

- `crates/pine/src/compile.rs:450` — the rewritten bgcolor/barcolor warning contains ~30 literal spaces inside the string literal: `"per-bar color                              columns land with a later milestone"` (verified with `cat -A`). This is a user-surfaced load warning, so the whitespace ships to the UI. The defect was inherited from the previous message, but this PR rewrites that exact line, and the repo's own answer is 170 lines below in this same diff (`compile.rs:646`, `"...to a \` + continuation). Fix: use the `\`-newline continuation.

- `crates/app/src/indicator_render.rs:622` and `:640` — two magic numbers in a renderer: the cross arm stroke `Stroke::new(1.5_f32, color)` and the marker-text gap `r + 2.0` / `-(r + 2.0)`. The module already names `MARKER_SIZE_PX` and `MARKER_GAP_PX` right above (`:35`, `:37`); these two escaped. Fix: `const MARKER_CROSS_STROKE_PX: f32 = 1.5;` and `const MARKER_TEXT_GAP_PX: f32 = 2.0;` at module top.

- `crates/app/src/indicator_render.rs:157` — `pane.top() + MARKER_GAP_PX * 2.0` reuses a constant documented as "gap between a bar's extreme and its above/below marker" as pane padding, with an unexplained `* 2.0`. A pane has no bars, so the closure is a deliberate substitution — but nothing says so. Fix: a named `PANE_MARKER_INSET_PX` plus one line of comment ("a pane has no candles; above/below markers anchor to the pane's own edges").

- `crates/app/src/indicator_render.rs:634` — `let above = matches!(marker.location, MarkerLocation::BelowBar);` is true exactly when the text is placed *below*. Renaming is free at runtime and the current name inverts the reader's model on the two lines that follow. Fix: `let text_below = ...`.

- `crates/app/src/indicator_render.rs:95`, `:249`, `:571` — the `bar_extents: &dyn Fn(usize) -> Option<(f32, f32)>` accessor was threaded through three signatures plus a synthetic closure at the pane call site (`:155`) and one at `app.rs:2299`, and `draw_view_plots` picked up `#[allow(clippy::too_many_arguments)]` (`:240`) to absorb it. The next milestone is announced by this PR's own warning text — per-bar color columns for `bgcolor`/`barcolor` — and it needs per-slot bar data threaded down exactly the same chain, reopening all three signatures again. Fix: bundle the frame-invariant arguments into one `struct PlotFrame<'a> { x: &'a PlotX<'a>, start: usize, end: usize, partial_slot: Option<usize>, bar_extents: &'a dyn Fn(usize) -> Option<(f32, f32)> }` and pass that; the color accessor then docks as a field, not as a fourth signature edit.

### Consider

- `crates/app/src/indicator_render.rs:551` — `draw_fill` is not cached and not invalidated: it recomputes and re-emits one `Shape::convex_polygon` **with its own `vec![Pos2; 4]` heap allocation** per adjacent visible pair, every frame. Concrete rate: `visible_range` is `width / candle_width` with `MIN_CANDLE_WIDTH = 2.0` (`viewport.rs:14`), so a 2560 px chart at full zoom-out is ~1280 quads → ~1280 allocations per frame per fill, ~77 k/s at 60 Hz; at the default 8 px width it is ~320. Nothing in this module caches geometry (every `draw_line` builds its `Vec<Pos2>` per frame too), and `draw_area:402` established this exact per-pair-quad pattern with the same justification comment — so this is the module's design, not a regression this PR introduces, which is why it is not a Should-fix. The real fix is one `Shape::Mesh` per fill: reserve `4 * n` vertices and `6 * n` indices once, push two triangles per pair, `painter.add` once — one allocation instead of `n`, and it makes the module header's "never one shape per bar" claim true for bands. If taken, apply it to `draw_area` in the same pass. Magnitude beyond the allocation count is unmeasured; the perf HUD frame time with the `shapes_and_fill.pine` corpus script loaded at `MIN_CANDLE_WIDTH` would settle it.
- `crates/pine/src/eval.rs:791` — the marker firing rule is `as_cond().unwrap_or(false) || matches!(Num(n) if !n.is_nan() && n != 0.0)`, i.e. a numeric `0.0` draws nothing. That reads deliberate, but TradingView's `plotshape` on a non-bool series draws wherever the value is not `na`, so `plotshape(close - close)` diverges silently. `docs/pine-dialect.md:138` documents the arguments but not this rule. Fix: one sentence in the dialect doc, and a case in `shape_plots_stage_condition_cells` pinning `0.0 → na`.
- `crates/indicators/src/output.rs:101` — `LabelUp`/`LabelDown` are documented as "a small up-pointing tag" but render identically to `TriangleUp`/`TriangleDown` (`indicator_render.rs:596`, `:607`). `draw_objects:497` already draws a rounded pill for `LabelObj`. Either reuse that shape or say in the doc comment that the tag falls back to a triangle for now — the repo's data-honesty rule applies to labels as much as to values.
- `crates/pine/src/compile.rs:565-577` — `shape_call` sets `style: PlotStyle::Circles` and `width: DEFAULT_PLOT_WIDTH` on a spec whose render path (`indicator_render.rs:271`) `continue`s before ever reading either. No other reader of `PlotSpec::style` exists anywhere in `crates/app` at the top of the stack, so these are dead values that will read as intent to the next author. A `#[doc]` note on `PlotSpec::style` ("ignored when `marker` is `Some`") costs nothing.
- `crates/pine/src/compile.rs:655` — `plot_ref_of`'s `!self.reassigned...unwrap_or(true)` guard is the honest part of fill resolution (a plot handle that is later `:=` reassigned cannot be trusted) and it has no test. `a = plot(close)\nb = plot(open)\na := b\nfill(a, b)` should hit the same refusal as `a_fill_between_non_plots_is_refused_honestly`; today nothing proves it.

### Verdict

- **Docking**: the pine side docks cleanly — `shape_call` sits beside `plot_call` as a peer, `fill()` resolution is a separate post-resolution pass with its own `pending_fills` queue, and the honest refusal for unresolvable handles follows the repo's existing warning/error split. The declare-RHS fix in `collect_declarations` (`compile.rs:428`) is the right place and does not disturb input extraction (`Builtin::Input*` still falls to `_ => {}`). The renderer side docked by widening four signatures; the next capability announced by this PR's own warning will widen them again — see the `PlotFrame` finding. Unknown `shape.*`/`location.*` members are refused at resolution rather than silently coerced, which was checked before writing it up as a finding and then dropped.
- **Performance**: eval is flat — the marker arm writes one `f64` into `self.row`, no allocation, no map. Load time gains one pass over a small `pending_fills` vec. Render gains, per frame and only for scripts that use them: ~1 quad + 1 small `Vec` per visible bar per fill (~320 at default zoom, ~1280 at `MIN_CANDLE_WIDTH`), and 1 shape per *fired* marker — markers are sparse by construction since `na` cells `continue`, so they need no batching. Nothing was made slower for scripts that use neither.
- **Proof**: `compile_passes.rs::shape_plots_register_markers_and_fills_resolve` fails without the marker registry and without the declare-RHS fix (it asserts 3 plots where the old compiler produced 1); `a_fill_between_non_plots_is_refused_honestly` fails if the refusal is weakened to a silent skip; `script_semantics.rs::shape_plots_stage_condition_cells` fails if the eval staging drops the `na`/`1.0`/absolute distinction. The render path — `draw_fill`, `draw_shape_markers`, and the `value_range` bug above — has no test, and this module has never had one; `value_range` is a pure function and is the place to start.

PR #107, opened 2026-08-02
feat(app): indicator settings UI from InputSpec (M4, PR 11/12)

milocaetano:
First PR of milestone M4 of [docs/indicator-system-plan.md](https://github.com/milocaetano/quantick/blob/main/docs/indicator-system-plan.md). **Stacked on #106** (← #105 … ← #96) — merge in order.

## What's here

- **Generated settings dialog** (`indicator_panel.rs`) — one widget per `InputSpec` variant: int/float drags with min/max/step, checkbox, color picker, text/options combo, and a source dropdown over every series **including the flow ones** (`delta`, `cvd`, …). The crate declares, the app renders; nothing is per-indicator.
- **Apply = `SetInputs`** — the worker rebuilds the instance from its retained source with the bound values and `replace`s it in the host (construct anew → replace → replay). A running indicator never observes an input changing mid-stream (§2.7 locked rule); the UI gets a full `Rebuilt` carrying the new values. Binding is defensive: missing/mistyped values fall back to declared defaults instead of panicking the worker.
- **Gear button** on each INDICATORS menu entry; `Rebuilt` events now carry bound values so the dialog opens with reality.
- **Worker test**: `SetInputs` (EMA 3→2) yields columns bit-identical to a fresh len-2 host run, descriptor and values following.

Declared deferral: `Progress` events for recompute stay out — measured rebuild cost is 77 ms/100k bars, far below where a progress bar earns its complexity, and the worker never blocks a frame either way.

## Verification

- [x] fmt --check / clippy -D warnings / build / test --workspace (38 suites)
- [x] arch-review: clean


milocaetano:
## PR #107 — arch-review

### Blockers
- none

### Should fix
- `crates/app/src/indicator_worker.rs:277` — **the dialog's opening values are seeded from the *declared* default, not from what the instance is actually running.** `Add` builds `mirror.values` from `descriptor().inputs[..].default_value()`, and `Ema::new(len, source)` always declares `default: EMA_DEFAULT_LEN` (`crates/indicators/src/native/mod.rs:71`) whatever `len` it was constructed with. So an EMA built with any other length mirrors a lie — this PR's own test adds `NativeEma { len: 3 }` (line 764) and the mirror reports `Int(9)`. Opening settings on such a slot and pressing Apply without touching a widget silently rewrites the indicator from 3 to 9. Nothing is visibly wrong today only because both shipped construction sites pass `DEFAULT_EMA_LEN = 9` (`crates/app/src/app.rs:68`) — i.e. the invariant is "two constants in two crates happen to be equal". Correctness/honesty outranks everything else here. Fix: add a defaulted, object-safe `fn input_values(&self) -> Vec<InputValue>` to the `Indicator` trait (same shape as the existing defaulted `objects()`, `crates/indicators/src/indicator.rs:145`) returning the declared defaults; override it in `Ema` (real `len`/`source`) and `ScriptIndicator` (its bound `inputs` field); seed `mirror.values` from it.
- `crates/app/src/indicator_worker.rs:340` — **`mirror.values = values` records what was asked, not what was bound.** Every fallback inside `build_with` is silent: a missing/mistyped native cell keeps the old value (lines 61–66), and a script whose value count doesn't match the compile drops the whole draft and builds from defaults (line 80, guarding `with_inputs`' length assert at `crates/pine/src/script.rs:57`). In both cases the mirror, the `Rebuilt` event and the reopened dialog claim the new values. Neither branch is reachable through today's plumbing — but what makes them unreachable is structure, not a check, and finding #1 shows the mirror already diverges from truth. "Inferred or incomplete data is labelled as such, never silently patched" — a discarded input is exactly that. Fix: same accessor as #1 (mirror what the instance bound), and where a value is rejected, send `IndicatorEvent::Error` instead of nothing.
- `crates/app/src/indicator_worker.rs:57` — **the panel is generated, the apply path is hand-written per source.** Binding values back into an indicator is a `match` arm the author of the next native must remember to extend. Forget it and the dialog still renders every widget, Apply still sends `SetInputs`, the worker still rebuilds — with the values ignored and (per #2) recorded as applied. Scripts have the port (`ScriptIndicator::with_inputs`); natives don't. It matters beyond the dialog: at the top of the stack `SetInputs` is also the state-restore path (`app.rs:1559` on `feat/indicators-docs-perf`), so a native with a missing binding block would lose its saved settings on every restart. Fix: a defaulted `fn rebind(&self, values: &[InputValue]) -> Option<Box<dyn Indicator>>` on `Indicator` (`None` = "I declare no inputs", the honest answer for `Cvd`), so `build_with` collapses to build + rebind and the next native gets the behaviour for free.
- `crates/app/src/indicator_worker.rs:772` — **the test proves one of the two bound cells.** It applies `Source(SourceId::Close)`, which is what the instance already had, so deleting the `values.get(1)` arm at line 64 leaves it green. The script arm is worse off: binding non-default values to a `ScriptIndicator` is exercised nowhere in the workspace (`InputValue::` appears in `crates/pine` only in eval's reader at `eval.rs:1489`), so the second implementer of this port is unproven. Fix: flip the source in the existing test (`Source(SourceId::Hl2)` → expect title `EMA(2, hl2)` and columns equal to `Ema::new(2, SourceId::Hl2)`), and add a worker test over a one-line pine script with an `input.int` proving `SetInputs` moves its output.
- `crates/app/src/indicator_panel.rs:168` — the `_ =>` catch-all means a seventh `InputSpec` variant compiles and silently renders "(unrenderable input)" instead of failing the build where the next author is already working. The tuple match is why the catch-all exists; match on `spec` exhaustively and `if let` the value inside — the mismatch fallback is kept and exhaustiveness comes back. Free at runtime.
- `crates/app/src/indicator_panel.rs:128` — `step.unwrap_or(0.1)`: the drag speed every unconstrained float input gets is a policy number inline at the point of use. Name it (`const DEFAULT_FLOAT_DRAG_SPEED: f64 = 0.1;` at module top). The grid spacing at line 51 is *not* a finding — it matches the house style at `candle_view.rs:204`.
- `crates/indicators/src/input.rs:50` — `SourceId::ALL` is a hand-maintained `[SourceId; 14]` with nothing forcing it to stay complete; a fifteenth series silently vanishes from every settings dropdown (the enum's other matches are exhaustive, so the author gets no signal about this array). Fix: a test that matches a `SourceId` exhaustively and asserts `ALL.contains(&s)` — a new variant then breaks the build in the test.

### Consider
- `crates/app/src/indicator_panel.rs:43` — `format!("Settings — {}", …)` allocates per frame while the dialog is open (same class: `current.clone()` / `option.clone()` per frame in the Str combobox, lines 148–151). By the repo's own rate table this is a cold path and the magnitude is one small `String` per frame — not a performance finding. But the title is stable data; building it once into `SettingsDialog` at open time is free.
- `crates/app/src/indicator_panel.rs:144` — `InputSpec::Int`/`Float` carry an `options` field documented as "rendered as a dropdown instead of a numeric field"; the panel honours it for `Str` only. No producer fills it today (pine writes `options: Vec::new()` at `compile.rs:1254/1266/1291`), so nothing is broken — the first producer that does will silently get a free-form drag. Also note the `Int` `step` is mapped to `DragValue::speed`, i.e. drag sensitivity, not an enforced increment; a declared step of 5 still lets the user type 3.
- `crates/app/src/toolbar.rs:575` — the gear is offered on a slot that never loaded (`host_id: None`), where `SetInputs` no-ops at worker line 334: the dialog opens claiming "this indicator declares no settings" (which is false — the script declares them, it just didn't compile) and Apply does nothing. Either hide the gear for those entries or say why the dialog is empty.

### Verdict
- **Docking**: half. The *render* half docks properly — a new input kind is one arm in one cold-path function and nothing per indicator, which is the right port. The *apply* half does not: `build_with` is a per-source switch with no trait behind it, and its failure mode is silence rather than a compile error. One defaulted trait method in `quantick-indicators` closes findings 1–3 together.
- **Performance**: flat where it counts. Nothing added to tick ingest, bar close or the render loop. The rebuild runs on the worker thread (`host.replace` catches up over retained bars, `crates/indicators/src/host.rs:88`); the render thread only moves a `Rebuilt`. Cost is one full column copy per Apply — a click, correctly labelled in the gear's hover text. The only per-frame allocation is the dialog title, negligible.
- **Proof**: `set_inputs_recomputes_like_a_fresh_instance` (`crates/app/src/indicator_worker.rs:754`) fails if the rebind stops replaying from bar zero, or if the descriptor or values stop following. Nothing fails if the source-cell binding or the entire script binding path regresses.

_Every finding verified by opening the file on `feat/indicators-settings`; CI on this head is green (`gh pr checks 107` → ci pass)._

PR #108, opened 2026-08-02
feat(pine,app): switch + script hot reload + vwap starter (M4, PR 12/12)

milocaetano:
Final PR of milestone M4 of [docs/indicator-system-plan.md](https://github.com/milocaetano/quantick/blob/main/docs/indicator-system-plan.md). **Stacked on #107** (← #106 … ← #96) — merge in order. **This completes M4.**

## What's here

- **`switch` lands in the dialect** — subject form (`switch x` + `value => body`, matched by `==` rules) and condition form (`switch` + `cond => body`), bare `=>` as the default arm, inline or block bodies. Covered by parser, corpus and semantics tests.
- **Hot reload** (§4.5) — file-backed scripts are polled (~1 s mtime check, throttled); a changed file recompiles and replays. On errors, **the last good version keeps running** and the menu entry turns amber (*stale — the file on disk has errors*). The worker test proves the stale instance even keeps consuming new bars, and that a slot whose first load failed is healed by its first good save.
- **`vwap_cumulative.pine`** joins the starter pack (honest name: cumulative from loaded history — no sessions to anchor to on activity-sampled charts). The starter pack is now the plan's full six: ema, cvd, delta_histogram, vwap_cumulative, zigzag, range_box.
- **Corpus harness upgraded** to the full compile pipeline — every ok/ fixture now proves whole-script loadability, not just syntax.

## Verification

- [x] fmt --check / clippy -D warnings / build / test --workspace (38 suites; 471 app tests)
- [x] arch-review: clean (reload docks as one worker command + one event; the poll is throttled and walks only file-backed slots)


milocaetano:
## PR #108 — arch-review

### Blockers

- `crates/app/src/indicators/mod.rs:117` — **the stale flag is silently wiped by any unrelated rebuild.** `IndicatorEvent::Rebuilt` sets `view.stale = None`, and `publish_deltas` (`indicator_worker.rs:443`) emits `Rebuilt` for *every* slot whenever the batch contained a `Backfilled` or `Rebuild` — i.e. on prepended history (`app.rs:1340`, the user scrolling left), a spec switch, or a source reset. Sequence: broken save → amber "stale"; user scrolls back → `Rebuild` → amber dot gone, while the host is still running the pre-edit instance and the file on disk still has errors. It never comes back, because `poll_script_files` already advanced `seen_mtime` and will not re-fire until the next save. This is the data-honesty rule in `CLAUDE.md` (priority 0 in this skill's order): the UI asserts "healthy" for code that is not what is on disk. `hidden` and `errored` both survive a rebuild — `stale` is the only status that a routine chart interaction can erase. Fix: make the worker own the flag. Add `stale: Option<String>` to `SlotMirror`, set it in the `Reload` error arm, clear it in the success arm, and carry it on `IndicatorEvent::Rebuilt` so the UI mirrors the worker instead of guessing. `IndicatorViews::apply` then assigns `view.stale = stale` rather than `None`. Regression test: `Reload` (broken) → `Rebuild(bars, partial)` → assert `view.stale.is_some()`.

### Should fix

- `crates/app/src/indicator_worker.rs:371` — **a reload silently resets the user's input values to the declared defaults.** The `Reload` arm calls `source.build()` (defaults) and then overwrites `mirror.values` with `InputSpec::default_value`, unlike `SetInputs` at `:347` which builds with the bound values. Effect: a user sets `len = 50` in the settings dialog, then edits a comment in the same `.pine` file, and the indicator jumps back to `len = 20` with no message — the chart changes for a reason the user did not ask for. `build_with` already degrades safely when the input set changed (it falls back to defaults unless `values.len() == compiled.inputs.len()`), so the fix is `source.build_with(Some(&mirror.values))`, keeping `mirror.values` only when the length still matches and otherwise taking the new defaults *and* saying so. Test: reload a script after `SetInputs` and assert the bound value survives; reload a script whose input list changed and assert the defaults are taken.
- `crates/app/src/indicator_worker.rs:386` — **the heal path has no test.** `None => mirror.host_id = Some(host.add(indicator))` is the branch the commit message advertises ("a slot whose first load failed is healed by its first good reload"), and it is reachable in the app: `add_script_indicator` registers the watch on a successful *read*, so a script that reads but does not compile leaves `host_id: None` with a live watch entry. Neither `reload_tests` test exercises it — `drive()` always starts from `GOOD_V1`. It is also the one reload branch that calls `host.add` instead of `host.replace`, so nothing proves the healed instance catches up over the existing history. Add a third test: `Add` a broken script, `Backfilled(bars)`, then `Reload` with a good one, and assert `view.error.is_none()` and `view.rows() == bars.len()`.
- `crates/pine/src/parser.rs:550` — **a default arm in the middle silently kills every arm after it.** The arm loop accepts a bare `=>` at any position, and `eval.rs:449` matches it unconditionally, so `switch x` with `=> 0` as the second of four arms compiles clean and the last two arms are dead code with no diagnostic. Real Pine requires the default arm last; the parser already validates "a switch needs at least one arm" two lines below, so the honest version costs one `if`: after the loop, if any arm other than the last has `condition.is_none()`, push a `PineSyntax` error ("the default `=>` arm must be the last arm"). Fixture: `corpus/err/switch_default_not_last.pine`. Everything else about the missing-arm behaviour is honest — no match yields `Value::Na`, which is what Pine does, not a fabricated default.

### Consider

- `docs/pine-dialect.md:43` — the new `switch` paragraph documents both forms and the bare `=>` default, but never says what a switch evaluates to when nothing matches and there is no default (`na`). One clause, and it is the question a user hits first.
- `crates/app/src/toolbar.rs:558` — the amber dot carries no hover text while all three buttons next to it do, so the compile errors the pine crate rendered with file:line:col and stable codes are stored in `IndicatorView.stale` and never shown. Carrying `Option<String>` in `IndicatorMenuEntry` and adding `.on_hover_text(message)` is free. (The red `errored` dot has the same gap today — pre-existing, but this PR is the natural place to fix both.)
- `crates/app/src/app.rs:1409` — `script_files` is never pruned when an indicator is removed, so the poll keeps stat'ing the file and firing `Reload` + an `INDICATOR_SCRIPT_RELOAD` info log for a slot the worker no longer knows. Harmless (the worker's `slots.get_mut` drops it) and **already fixed later in the stack** — the top branch has `self.script_files.retain(|(s, ..)| *s != SlotId(slot))` in the remove path. Noted only so it is not lost if the stack is re-ordered.
- `crates/app/src/app.rs:866` — `add_script_indicator` now inlines `allocate_slot` + `send(Add)` because it needs the slot id, giving two ways to add an indicator (`add_indicator` at `:1453` is the other). Making `add_indicator` return `SlotId` removes the fork.
- `crates/app/src/app.rs:3920` — `poll_script_files()` sits between `draw_indicator_settings` and the status bar, inside the block whose comment explains how chrome panels claim their zones outside-in. It is periodic housekeeping, not drawing; `ensure_book_capture()` / `maybe_emit_summary(now)` at `:3905` is where the frame's other debounced work lives.

### Verdict

- **Docking**: clean. `Reload` docks onto the existing command/event port with no new type switch, and the worker stays source-agnostic — it only ever sees `IndicatorSource`, so a future non-file script origin needs no worker change. `NodeKind::Switch` follows the `If`/`For`/`While` shape exactly: `statement_value` routes it like the other block keywords, `compile.rs` walks it like the others, `eval.rs` evaluates only the taken arm like `if` and the ternary. The `values_equal` extraction removes the duplicate `==` table rather than adding a second one. The one thing that does not dock is the watch registration: it lives in `add_script_indicator`, so any future path that instantiates a script (a saved layout, a drag-and-drop) must remember to push to `script_files` — a `watch(slot, index)` helper next to `add_indicator` would close that.
- **Performance**: flat. Per frame the poll costs one `Vec::is_empty` plus one `Instant::elapsed` before returning; once per second it does N `fs::metadata` calls and N `PathBuf` clones on the UI thread, N being the number of file-backed script indicators (single digits in practice) — fine on a local disk, and the syscall is the thing to move to the worker if the scripts directory ever lives on a network mount. The interval is a named `Duration` constant matching the repo convention (`SUMMARY_INTERVAL`, `STYLE_LOG_DEBOUNCE`); recompilation happens on the indicator thread, so a broken save never stalls a frame. The app repaints every 16 ms (`app.rs:3995`), so the poll fires on an idle chart too. In the evaluator, a `switch` costs one `NodeKind` clone per evaluation (its `arms` Vec) — the same class as `Block`, `Tuple` and `Call` in this evaluator, and an if/else chain of the same arity allocates one Vec per block, so the new construct is no more expensive than what it replaces.
- **Proof**: `indicator_worker.rs:897 a_broken_reload_keeps_the_last_good_version_and_flags_stale` would fail if a bad edit took the chart away, and it goes further than the name suggests by asserting the stale instance still consumed a new bar; `script_semantics.rs switch_matches_subjects_conditions_and_defaults` covers condition arm, mirrored subject arm and default in one replay; the corpus harness now runs `quantick_pine::compile` instead of lex+parse, so every `ok/` fixture proves whole-script loadability, and `embedded_scripts_compile_against_the_dialect` already covers the new vwap starter. Gaps: the poll's own contract (mtime advances even when the compile fails, so a broken save does not re-fire every second) is only asserted in a doc comment, the heal path is untested, and nothing guards the stale flag against a rebuild — which is how the Blocker got in.

PR #109, opened 2026-08-02
feat(app): persist the indicator set (M5, PR 13/14)

milocaetano:
First PR of milestone M5 of [docs/indicator-system-plan.md](https://github.com/milocaetano/quantick/blob/main/docs/indicator-system-plan.md). **Stacked on #108** (← #107 … ← #96) — merge in order.

## What's here

- **`indicators-state.toml`** (§4.6) — versioned TOML next to the config (`QUANTICK_INDICATORS_STATE` override), the drawing-presets discipline: unknown version or garbage **starts empty and says so** — never a half-read workspace. Scripts persist **by library name** (files stay files; no stale source embedded), natives by kind, inputs as typed values with sources by dialect name — the file is hand-readable and hand-editable, with honest fallback to declared defaults for nonsense.
- **Restore** goes through the same commands the menu sends (add → `SetInputs` → hidden once the view lands); a missing script is skipped with a log line.
- **Save** is debounced (1 s) and event-driven — add/remove/eye/settings mark dirty; restoring does not; nothing rides the frame path.
- Declared, not snuck: pane layout stays at the plan's fixed 20% fraction; draggable dividers remain future polish.

## Verification

- [x] fmt --check / clippy -D warnings / build / test --workspace (38 suites; 3 new persistence tests)
- [x] arch-review: clean


milocaetano:
## PR #109 — arch-review

### Blockers
- `crates/app/src/app.rs:1509` and `crates/app/src/app.rs:1572` — the PR's headline behaviour, restore and debounced save, has no test. The three new tests all sit on the pure serde layer in `state_file.rs` (round-trip, bad version, input conversion); nothing exercises `restore_indicator_state` / `maintain_indicator_state`, so no test fails if the wiring regresses — and the wiring is where the defect below actually is. This is not a "hard to test UI" case: the app.rs test module already builds a headless `QuantickApp::new(...)` (`crates/app/src/app.rs:4240`), and `indicator_state_path` is a plain field, so the arrangement is two lines. Add: (1) write a fixture file, set `app.indicator_state_path`, call `restore_indicator_state()`, assert `slot_kinds` and `pending_hidden` match the file and that a `SavedKind::Script` naming a script the library lacks adds nothing; (2) set `last_indicator_change` to an instant older than `INDICATOR_STATE_SAVE_DEBOUNCE`, call `maintain_indicator_state()`, and assert the file on disk equals the live set — including that a slot with no `slot_kinds` entry is the only thing omitted.

### Should fix
- `crates/app/src/app.rs:1535` — `self.slot_kinds.last().map(|(slot, _)| *slot)` assumes `add_script_indicator` always pushes exactly one entry as its last act. It does not: the push at `crates/app/src/app.rs:912` only happens in the `Some(Ok(text))` arm, so when `script_library.read(index)` returns `Some(Err(_))` (file deleted or unreadable between `ScriptLibrary::scan()` in the constructor and this call) `.last()` silently returns the **previous** entry's slot — the missing script's saved inputs are then bound to an unrelated indicator via `SetInputs`, and its `hidden` flag toggles that unrelated indicator invisible. Correctness before everything else; the trigger window is narrow today but the coupling breaks on any future early return in `add_script_indicator`. Fix: change its signature to `fn add_script_indicator(&mut self, index: usize) -> Option<SlotId>` and use the returned value here (the other two call sites ignore it).
- `crates/app/src/app.rs:623` with `crates/app/src/app.rs:913` — a scripted validation run now writes itself into the user's persistent state. `restore_indicator_state()` is inserted *before* the `QUANTICK_INDICATOR_SCRIPTS_AUTOSTART` hook, and that hook goes through `add_script_indicator`, which marks the state dirty — so one second later the autostart scripts are appended to `indicators-state.toml` and open by themselves on the next plain launch. The sibling hook behaves the opposite way: `QUANTICK_INDICATORS_AUTOSTART` adds its two natives through `add_indicator` directly, never registers them in `slot_kinds`, so they are silently dropped from every save — and, because they are added before the restore, a validation run over a saved EMA+CVD set now opens four indicators instead of two. Two contradictory rules from one feature family, and the first one violates "config/env presence alone never activates anything". Fix: give `add_script_indicator` a `persist: bool` (or a small `Provenance` enum) so autostart does not dirty the file, and apply the same answer to the natives hook.
- `crates/app/src/app.rs:1557` — one unresolvable input silently discards *every* input of that entry: `values.len() == entry.inputs.len()` fails, no `SetInputs` is sent, and there is no log line. `crates/app/src/indicators/state_file.rs:73` documents the opposite ("the worker then falls back to *that* input's declared default"). Data honesty: a hand-edited or stale file loses the user's whole parameter set without saying so. Fix: emit an `INDICATOR_STATE_INPUTS_DROPPED` warn (same shape as `INDICATOR_STATE_SCRIPT_MISSING`) carrying the kind and both counts, and correct the doc comment on `to_value` to say the entry — not the single input — falls back to declared defaults.
- `crates/app/src/app.rs:1603` — the save rewrites `inputs` from `view.input_values`, which is **empty** for a slot whose `Add` failed to build: the worker's error path sends `Rebuilt { inputs: Vec::new(), .. }` (`crates/app/src/indicator_worker.rs:310-322`). So a script that does not compile at startup is restored, gets an empty view, and the next unrelated change (an eye toggle anywhere) rewrites its entry as `inputs = []` — the user's saved parameters are gone before they even fix the script. Fix: skip the `inputs` rewrite for a view with `error.is_some()` and carry the values loaded from the file for that slot instead.

### Consider
- `crates/app/src/indicators/state_file.rs:161` — `std::fs::write` truncates before writing, so a crash or power loss mid-write leaves a truncated file; `load` then reports it unreadable and starts empty, i.e. the whole workspace is lost rather than silently corrupted. This matches `crates/app/src/drawings/presets.rs:99` exactly, so it is repo precedent, not a new invention — but a shared `write_atomic(path, text)` helper (temp sibling + `rename`) applied to both stores would close it once.
- `crates/app/src/indicators/state_file.rs:207` and `:218` — both tests write into a fixed `temp_dir()/quantick-indicator-state-test`. The repo's own answer is `scratch_path` in `crates/app/src/drawings/presets.rs:173`, which suffixes `std::process::id()` and removes the file up front. Reuse that shape so two concurrent runs on one machine cannot collide.
- `.gitignore` — `indicators-state.toml` is written to the process CWD, so running the app from the repo root leaves an untracked file in `git status` (the same is already true of `quantick-drawing-presets.toml`). One line each closes it.
- `crates/app/src/app.rs:1585` — `settled` is computed before the `indicator_state_dirty` check, so `Instant::elapsed()` runs every frame once the user has ever changed anything (`last_indicator_change` is never cleared, only the dirty flag is, at `:1589`). The cost is one clock read at ~60 Hz — negligible, called out only because clearing `last_indicator_change` alongside the flag makes the state machine read correctly.

### Verdict
- **Docking**: good. Restore replays the exact commands the menu sends (add → `SetInputs` → hide), so there is no second construction path for indicators, and `state_file.rs` depends only on `quantick_indicators` types — no reverse edge. `SavedKind` is a closed enum that a third native would reopen, but it mirrors the already-closed set of native `ToolbarAction`s, so a new native touches three files instead of two; that is the existing repo shape, not a regression. The two things a next author *is* forced to open are named above: `add_script_indicator`'s missing return value and the autostart/persist rule.
- **Performance**: flat. The per-frame guard in `maintain_indicator_state` is one `is_empty()` and one `Instant::elapsed()` at ~60 Hz; serialization and `fs::write` happen at most once per second of editing and only after a change settles, behind the named `INDICATOR_STATE_SAVE_DEBOUNCE` (`crates/app/src/app.rs:73`, unit carried by `Duration::from_millis`). Restore is one file read at startup. `pending_hidden` drains within a few frames and its scan is O(pending × views) with both in single digits. The write path is honest on values: `InputValue::Float` is `f64` end to end (`crates/app/src/indicator_panel.rs:117` drags the `f64` in place), colours persist as `u8` channels and sources by dialect name, and `toml::to_string_pretty` emits Rust's shortest round-tripping decimal — no `0.78 → 0.7799999713897705`. The file is machine-owned, so nothing loses comments in practice, though the module doc's invitation to hand-edit (`state_file.rs:8`) does not survive the next save.
- **Proof**: today only `state_file::tests::the_state_round_trips_through_disk` (`crates/app/src/indicators/state_file.rs:205`) would fail if the on-disk format regressed, and `unknown_versions_and_garbage_start_empty` (`:216`) guards the never-half-read rule. Nothing at all would fail if restore bound inputs to the wrong slot or if the save dropped an entry — which is exactly the gap the Blocker names.

_CI on this PR is green (`gh pr checks 109` → pass); cargo was not run locally per the review instructions._

PR #113, opened 2026-08-03
feat: three months of venue-native candle history in the time pane

milocaetano:
## What

Stacked on #112. The time pane now opens with **90 days of venue-native 1-minute candles** (`TIME_HISTORY_SPAN_MS`), resampled locally to the pane's interval — a timeframe switch never reaches the network.

- **One contract for every feed**: `FeedCommand::FetchOhlcv { span_ms }` → exactly one `FeedEvent::OhlcvHistory { interval_ms, bars }` (empty on failure/unsupported), gated by the new `FeedCapabilities::ohlcv_history`. Replay honestly answers empty (a recording has ticks only).
- **Binance**: `/api/v3/klines` pager with the workspace's first rate-limit handling (bounded retries, Retry-After honored, ban-length waits refused, page budget) and a labeled-partial result on persistent refusal. Real taker-buy split → the prefix's `delta()` means something.
- **Hyperliquid**: `candleSnapshot` over a dedicated short-lived websocket (`post` method, no new dependency; request/response state never touches the per-print trade loop).
- **MetaTrader**: the bridge protocol gains an optional, schema-1-compatible rates block (`rates_start` / chunked `rate` / `rates_end`, sized against the 64 KiB line cap with the math in PROTOCOL.md). The Python bridge pushes it after tick backfill (`--rates-months`, default 3). The published `ohlcv_history` capability means exactly "a block is in hand" — it rises when the block arrives, which is what makes the pull-shaped app contract work over a push-only protocol. EA rates support deliberately deferred: an EA session declares nothing and the capability stays false (honest absence).
- **App**: the venue prefix lives OUTSIDE `ChartState` (the engine's deterministic rebuild path is untouched — zero engine edits) and is composed at read time; the seam is re-trimmed on history prepends; a dashed amber divider distinct from the solid backfill divider marks the venue↔engine seam, and the status bar's count goes three-way (`26000v+240+61 bars`). Indicators see the full composed series — an EMA spans three months. Volume-honesty: venues without an aggressor split store half the volume per side so `volume()` is exact and `delta()` is identically zero — a stored representation, documented in §11 and at the mapping sites, never presented as measurement.

## Architecture review

Two-lane arch-review over the branch diff found **2 Blockers** (MT5 capability rising before any block exists → time pane empty all session on MT5; `HistoryPrepended` invalidating the seam trim → duplicated windows and broken slot ordering three clicks away) and 12 Should-fix findings — all resolved in the last five commits, each with a test that fails without the fix.

**Deliberately deferred** (with reviewer agreement):
- Progressive delivery for the Binance fetch (~15-20 s for 90 days, answered as one event; the pane stays fully usable and the spinner is honest — chunked replies are a contract change if the wait ever bothers in practice).
- EA (`QuantickBridge.mq5`) rates support — the Python bridge is the primary live path; an EA session reports the capability false.
- `chart::price_window` / `PriceScale::auto` widened from `&[Bar]` to `impl IntoIterator` so the seam view chains two slices allocation-free; all existing callers compile unchanged.

## Verification

`cargo fmt --all -- --check` · `cargo clippy --workspace --all-targets -- -D warnings` · `cargo build --workspace` · `cargo test --workspace` — all green; 1070+ tests workspace-wide (595 in the app crate), ~50 new across klines paging/rate limits, candleSnapshot decoding, the MT5 rates block (including chunk-size math and a cross-language drift guard on `MAX_BARS_PER_RATE_LINE`), the resample fold, seam trimming, and the request lifecycle.


milocaetano:
Final pre-merge arch-review round (cross-PR coherence pass) found and fixed two Blockers at the PR joints: (1) the add-symbol picker could persist a sidecar state the port cross-check rejects at next launch — additions are now validated against a full config clone before persisting, and a sidecar that fails validation at load is dropped with a warning instead of blocking startup; (2) the MT5 candle capability was a monotone latch — an empty first block buried every later one; FeedCapabilities now carries ohlcv_generation, bumped per stored block and consumed tab-side. Also landed with this round: rates_end gains an optional partial flag and OhlcvHistory carries complete on every provider (Binance stops flattening it); the paging loop halves-and-retries against the terminal's Max-bars refusal; busy-port refusals reach the chart as a notice (and the two refusal tests now assert that event as contract); python paging logic gained a stubbed test suite run from cargo. Correction to this PR body: the per-commit failing-test claim did not hold for two earlier fix commits (c0e00b2, 811cf47) — both paths are covered by tests as of cff1a07/e5c1a11.

PR #115, opened 2026-08-04
feat: TradingView-style drawing toolbar rail and inspector UX

milocaetano:
## What

Redesigns the drawing-tools chrome to TradingView quality, driven by three complaints: the toolbar read as a placeholder, the properties popup opened on top of the drawing and ate clicks, and the toolbar could not dock on a lateral edge.

Design spec: `docs/drawing-toolbar-ux.md` (written by a dedicated UX pass; supersedes the corner-docked geometry in `ui-design-model.md` §7).

### Toolbar (`toolrail.rs`, rewritten)
- Four **edge** docks replace the four corners: a vertical 44 px rail on Left (new default) / Right, horizontal on Top / Bottom. Grip drag shows a drop-preview band on the candidate edge (normalised nearest-edge rule, so a wide window doesn't bias top/bottom); **Esc cancels** the drag; `View → Drawing toolbar` is the keyboard path.
- Leading / trailing clusters with the trailing cluster (repeat, hide-all, lock-all, Objects) **pinned to the rail's far end**; hairline separators; accent **edge marker** on the armed tool; new pressed state and focus ring; draft-progress and object-count badges.
- The two Fib tools fold into one **family slot** (declared per tool in the registry — no central match): left-click arms the last-used member, right-click or the caret corner opens a flyout listing members with shortcuts.
- Overflow has three axis-aware stages (Full 417 px / Compact 345 px / Minimal 191 px for the shipped registry); `main.rs` gains `with_min_inner_size` so the floor is unreachable.

### Inspector (`app.rs`)
- Placement scores **eight clamped candidates** (four beside the object, four chart corners) by least overlap — the blind `candidates[0]` fallback that parked the panel on top of the drawing is gone.
- A chart narrower than 1180 px opens the inspector **pinned**, until the user touches the pin (their preference then holds).
- Custom 28 px title bar: the whole bar drags (the body never does), icons for hide / pin / close; **manual position sticks for the session** — selection changes never snap it back, the only automatic move is a re-clamp when the pane shrinks; double-click the bar to re-arm auto placement.
- **The inspector is opaque to the pointer**: the raw-press exception that let a press land twice (once on the panel, once as a drawing drag — every press near a horizontal line hit it) is removed. The gate applies at press time only, so a drag that started on the canvas keeps running across the panel; the `interactable(!drag)` workaround is gone.
- The object manager opens beside the rail's inner edge in all four docks.

### Tokens (`theme.rs` / `widgets.rs`)
`press_tint()` (accent @ 33%), `TEXT_SUPPORT` for explanatory lines (4.5:1 on chrome), `TOOLRAIL_ICON` 18/32, active-marker + focus-ring constants, 350 ms tooltip delay. Amber stays provenance-only (grep-guarded test).

## Verification

- `cargo fmt --all -- --check` ✓ · `cargo clippy --workspace --all-targets -- -D warnings` ✓ · `cargo build --workspace` ✓ · `cargo test --workspace` ✓ (527 app tests; 43 new/updated across rail geometry, docking drag, overflow stages, family folding, placement, manual-position stickiness, auto-pin, pointer routing, tokens)
- arch-review run over the diff: 4 findings (per-frame registry refold, magic paint numbers ×2, missing manager-placement test) **fixed in `9af8ab3`**.

## Deferred (from arch-review)

- The tool-family port is exercised by a single real family (Fibonacci). A fake second family can't be registered from `toolrail` tests because `DrawingTool` wraps a private static trait object; when a second family lands, add a registry-level folding test for two adjacent distinct families.

## Behavioural changes to note

- Toolbar default moves from a top strip to a **left vertical rail**.
- `inspector_never_blocks_drawing_drag` encoded the old click-through exception and was replaced by two tests of the new contract (opaque press + drag continuity).

https://claude.ai/code/session_016rZDvD7peQrgE5e9UvbTVG

### DIFF
--- a/crates/app/src/app.rs
+++ b/crates/app/src/app.rs
@@ -555,6 +555,10 @@ pub struct QuantickApp {
     // Whether the user ever toggled the pin — the auto-pin width rule stops
     // firing once they have expressed a preference.
     inspector_pin_touched: bool,
+    // Set on the unpin frame: the side panel still occupies that frame's
+    // layout, so the floating host waits one frame and places against the
+    // settled chart instead of the pinned-era geometry.
+    inspector_settle_frame: bool,
     // The chart pane from the last frame (excludes axes and the live lane),
     // for inspector placement and manager centring.
     last_chart_area: Option<egui::Rect>,
@@ -694,6 +698,7 @@ impl QuantickApp {
             inspector_last_selection: None,
             inspector_pos: None,
             inspector_pin_touched: false,
+            inspector_settle_frame: false,
             last_chart_area: None,
             drawing_manager_open: false,
             drawing_manager_was_open: false,
@@ -4012,6 +4017,14 @@ impl QuantickApp {
             // The user has expressed a preference: the auto-pin width rule
             // stops firing for the rest of the session.
             self.inspector_pin_touched = true;
+            if !self.inspector_pinned {
+                // Unpinning re-opens the floating window. The pinned host
+                // has been claiming the selection each frame, so treat it
+                // as fresh again — otherwise automatic placement never runs
+                // and the window falls back to the fixed default corner.
+                self.inspector_last_selection = None;
+                self.inspector_settle_frame = true;
+            }
         }
         if actions.delete {
             self.request_delete_selected(now);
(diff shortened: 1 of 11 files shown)


milocaetano:
### Code review

Found 1 issue:

1. Unpinning the inspector after the narrow-chart auto-pin reopened the floating window at the fixed default corner (90,120) instead of beside the selection (bug due to the auto-pin early return in `draw_drawing_inspector` never claiming the selection while the pinned host claims it every frame, so `selection_changed` was false on unpin and automatic placement never ran)

https://github.com/milocaetano/quantick/blob/86588d39f54b63adebc66cb9b27d3abd6dcf50f9/crates/app/src/app.rs#L4132-L4141

Fixed in 27170132f6b10d6ada10eb8eb6b0552facdfd411: unpinning re-marks the selection as fresh and waits one settle frame so placement runs against the settled chart; the auto-pin test now asserts the reopened window's position.

https://github.com/milocaetano/quantick/blob/27170132f6b10d6ada10eb8eb6b0552facdfd411/crates/app/src/app.rs#L4019-L4028

Also checked: CLAUDE.md compliance, git-history context (the removal of the raw-press inspector exception is a deliberate, documented reversal with replacement tests), prior-PR review themes, and code-comment contracts — no further issues.

PR #118, opened 2026-08-04
feat: paper-trading simulator with chart trading UI

milocaetano:
## What

A trade simulator (paper trading) for practice on any feed and on market replay, per the goal "simulate trades didactically": buy/sell the charted symbol, place orders on the chart (including stop-loss and take-profit as draggable lines), persist closed trades to a folder as durable history, and read a performance report computed from that history. UX spec: `docs/ux/paper-trading.md`; it claims exactly the zones the design model's growth map reserved.

## Design

- **New pure domain crate `sim` (`quantick-sim`)** — deterministic tape-based core: market/limit/stop entries, netting position, protective brackets, conservative fill model (market fills at the *next* print; a limit at its own price on touch-through; a stop at the triggering print — gaps fill honestly worse). P&L is reported in **points**, never currency: the workspace has no per-instrument tick-value table, and an uncomputable number is not shown. Depends only on `engine`; guarded by the workspace dependency test.
- **App integration** — BUY/SELL toolbar group (folded into the collapse plan), Trading dock tab (position card, order entry with click-to-place arming, pending orders, session summary), draggable order/position price lines with gutter chips (drawn above drawings, below the last price), SIM P&L status-bar cell, and a performance report window that reads the history folder fresh on open.
- **Honesty contract** — every surface says SIM; rejections surface the sim core's didactic messages verbatim ("a buy stop must sit above the market (it chases strength) — to buy below the market use a limit"); a timeline rebuild (replay seek, feed/symbol switch) flattens at the last mark, labeled `reset`, toasted, and journaled — the drawings-cleared contract's sibling. Backfill only seeds the mark; filling against the past would be look-ahead.
- **History on disk** — `paper-trades/<SYMBOL>/<session>.csv` (`QUANTICK_TRADES_DIR` override), append-friendly `quantick-trades 1` CSV; torn tails are reported, never silently dropped; session names derive from venue time, so the same replay run names the same file.

## Tests

- `quantick-sim`: 34 inline tests — fill model (touch, gap-through, worse-fill stops), netting/reversal/partials, bracket arming (the entry print never triggers its own bracket), validation rejections, reset semantics, history format round-trip + torn tail, report metrics hand-computed, and a double-run determinism harness (same tape + same commands ⇒ identical events).
- `quantick-app`: 16 paper tests — journal lifecycle on disk, timeline reset, offset brackets, snap precision, armed placement (accept + reject-stays-armed), stop-loss drag repricing, entry-line gesture blocking, plus two end-to-end tests through the app's real event path (backfill seeds but never fills; a reset flattens and journals).
- Full loop green: `cargo fmt --check`, `clippy -D warnings`, `build`, `test --workspace` (45 suites).

## Arch-review

Run over the branch diff; two Should-fix findings fixed in `d419906` (named renderer constants; direct tests for the chart input routing). Deliberately deferred, per the priority order:

- **performance (Consider)** — the resting-order sweep allocates a small replacement `Vec` per print *while limit/stop orders rest*; bounded by the manual order count (single digits) and dwarfed by existing per-trade work. A fixture bench would settle the magnitude before optimizing.
- **standardisation (Consider)** — a second toast slot beside `DrawingToast`: deliberate, because `DrawingToast` couples a `&'static str` to the drawings undo stack (Undo affordance) and generalizing it would modify shipped drawing behavior.
- **future work** — an instrument table (tick value/currency) would upgrade points to currency P&L; explicitly out of scope here (data honesty over invented numbers).

Heads-up: open PRs #112 (tabs) and #115 (toolbar rail) touch `toolbar.rs`/`dock.rs`; whichever merges second will need a small conflict pass on the collapse-plan states and the dock tab list.

https://claude.ai/code/session_01KLUaEBthRtzjhHSB7P42PN


milocaetano:
### Code review

Found 4 issues:

1. Switching feed/symbol from the toolbar never flattened the simulated position — `maybe_switch_feed` inlines its reset logic instead of calling `reset_market_state()`, so the paper position, pending orders and SIM P&L cell survived onto another market's tape with no flatten, toast or journal entry (the `set_symbol` doc even claimed every switch flattens)

https://github.com/milocaetano/quantick/blob/52cdf7ce88f776a64be990b29a9bf6ccdf61bac5/crates/app/src/app.rs#L1245-L1253

2. The report window's "file(s) unreadable / row(s) skipped" warning used `theme::AMBER`, which theme.rs reserves as "not-live provenance only: replay, backfill, inferred data" and "never decoration"; `theme::WARN` ("threshold breaches and errors") is the documented fit

https://github.com/milocaetano/quantick/blob/52cdf7ce88f776a64be990b29a9bf6ccdf61bac5/crates/app/src/paper_trading.rs#L1170-L1180

3. Two tests used fixed temp-dir names without the `std::process::id()` suffix (also at L1491), so concurrent test runs on one machine collide and delete each other's files mid-test — the pattern PR #109's review established, and which this PR's own app.rs test already follows

https://github.com/milocaetano/quantick/blob/52cdf7ce88f776a64be990b29a9bf6ccdf61bac5/crates/app/src/paper_trading.rs#L1447-L1449

4. Paper trading read Escape independently inside `handle_navigation` instead of joining the app's single escape stack in `handle_drawing_keys` (toolrail.rs: "Escape lives in the app's escape stack ... not here"), so one press could fire two cancels in the same frame — and it cleared an armed placement but not an in-progress line drag

https://github.com/milocaetano/quantick/blob/52cdf7ce88f776a64be990b29a9bf6ccdf61bac5/crates/app/src/app.rs#L2362-L2364

All four (plus lower-confidence findings, including re-checking attached stop/take-profit levels against the actual fill price so an outrun level is dropped and reported instead of exiting with a mislabeled reason) are fixed on the branch in 5ed09ed, with regression tests.

PR #130, opened 2026-08-06
feat(app): indicator panes draw across the live lane, and answer the chart's gestures

milocaetano:
An indicator pane stopped at the live lane's divider while its value labels sat
in the gutter on the far side of the tape's band. The curve and the numbers
describing it were a lane-width apart with bare canvas in between, and the only
way to move that curve was to travel to the gutter and drag it there.

## What changes

**The pane is one strip, from the chart's left edge to its gutter.** Inside the
lane it draws in *tape time* rather than bar slots — the tape's own linear
mapping, so a pane's curve lands under the prints it was computed from.

**The lane segment is evaluated, not extrapolated.** `IndicatorHost::walk_partial_prefixes`
previews a ladder of forming-bar prefixes and restores the retained bar
afterwards: every rung is a real evaluation of a real prefix of the run, and
the walk commits nothing. The prefixes are folded with `Bar::opened_by` /
`Bar::extend` — the fold every builder already used privately, now engine API,
with a test asserting the public fold and the builders' own partial agree.

**What cannot be sampled is not invented.** A commit is one-way, so a bar that
already closed cannot be re-entered. Bars that closed inside the lane window
hold flat and step at their close — the committed resolution — and only the
forming bar gets rung-by-rung detail. The step shape *is* the label.

**Cost is bounded by the lane, not by the feed.** The rung budget is the lane's
own width (one rung per 6 px, capped at 64). A chart with no lane asks for no
rungs, sends no run of trades, and walks no ladder. The walk itself runs on the
indicator worker's existing coalesced cadence: per drained batch, never per
print and never per frame.

**Pane bodies answer the chart's gestures.** Press and drag pans the pane's own
scale vertically and the shared time axis horizontally, scroll zooms time, and
double click hands the scale back to auto-fit — the same verbs the candles use,
so the gesture language is one language. Time is applied once per drag however
many panes are stacked.

## Performance

Declared per touched path:

| Path | Rate | Cost |
| --- | --- | --- |
| Ladder walk | per drained worker batch | `rungs` previews per indicator |
| Lane polyline | per frame | one polyline per line-shaped plot, ≤ rungs + steps points |
| Lane window / steps | per frame | one published-book read, one bounded Vec |
| Pane body gesture | per frame | O(1) per pane |
| Per-trade / per-depth | — | untouched |

Measured, not asserted:

- `cargo bench -p quantick-indicators` (new `lane walk` case): a 64-rung walk
  over EMA + CVD costs **60.2 µs**, an 8-rung walk **7.6 µs** — linear in rungs,
  against a single preview at 938 ns.
- `APP_HEALTH_SUMMARY` A/B on a live BTCUSDT tape: same hooks (book + bubbles
  + live strip + the `cvd` script in a pane), 45 health windows each,
  alternating `main`, branch, `main`, branch so drift in the tape does not all
  land on one side. `frame_cpu_ms` median per run:

  | Run | frame_cpu median | heatmap_cells | trades/s |
  | --- | --- | --- | --- |
  | `main` r1 | 2.38 ms | 688 | 5.0 |
  | branch r1 | 2.39 ms | 499 | 3.6 |
  | `main` r2 | 2.26 ms | 475 | 3.4 |
  | branch r2 | 2.74 ms | 729 | 6.7 |

  The runs are not load-matched — the market decided how busy each one was, and
  frame cost tracks `heatmap_cells` closely. Fitting each build's two points
  against cell count and reading both at 600 cells gives **2.33 ms (`main`) vs
  2.54 ms (branch): about +0.2 ms of frame CPU**, on a ~16.7 ms budget, for a
  curve, a wider background fill and gridlines that were not drawn before.

  **This is not flat, and it is the honest number.** An earlier measurement of
  this branch showed +0.66 ms; most of that was a second published-book read
  per frame (a worker mutex on the render thread), fixed in b62eb57. What is
  left is the cost of the new ink.

  Both builds render the chart area **white** when launched this way in this
  environment, at fps 19 with `frame_cpu` around 2.4 ms — the surface does not
  present for a process started detached here. `main` behaves identically, so
  it is an environment state rather than a regression; `frame_cpu_ms` is real
  CPU work per frame either way, which is why the comparison above uses it and
  not fps.

## Deferrals

- **Histogram and marker plots are not drawn in the lane.** A column is a
  statement about one bar at that bar's width, and a time axis has no bar
  widths to honour. A delta-histogram pane therefore still shows an empty
  band. Line, step-line and area plots are drawn.
- **The committed and sampled halves of the lane curve share one stroke.** The
  stair vs. curve shape distinguishes them; a dimmer stroke for the coarser
  half was considered and not done.
- **Visual QA did not run.** The app's chart area captures white in this
  environment for both this branch and a `main` control build (see
  Performance), so no screenshot of the feature exists. The geometry, the
  stepping, the rung placement, the budget and all three pane gestures are
  covered by tests instead — but nobody has looked at it yet. Worth a human
  glance before merge.

## Verification

- `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`,
  `cargo build --workspace`, `cargo test --workspace` — all green.
- `arch-review` run over the diff. Three findings it raised were fixed in the
  branch rather than deferred: the forming bar's trades were cloned on every
  drain even with no lane to draw them in; the lane samples were cloned a
  second time on their way out of the worker; and reading the lane's live edge
  took a second worker-mutex lock per frame on the render thread.
- `trader-ux-review`: one Should-fix it raised was fixed rather than deferred —
  a pane body that had become an interactive surface answered a drag on y only
  and ignored scroll entirely, while the candles' body pans both axes and zooms
  time. The gesture set is now the same everywhere.

### Tests

| Behaviour | Test |
| --- | --- |
| The public fold reaches the builders' own partial | `engine::bar::folding_trades_by_hand_reaches_the_builders_own_partial` |
| A rung equals a standalone preview of the same prefix | `indicators/host::every_rung_equals_a_standalone_preview_of_the_same_prefix` |
| A walk commits nothing and restores the live preview | `indicators/host::a_walk_leaves_committed_state_and_the_live_preview_untouched` |
| The ladder is a port: a failing instance drops out, its neighbour does not | `indicators/host::a_failing_instance_drops_out_of_the_ladder_without_taking_its_neighbour` |
| The newest print always gets a rung, whatever the budget | `app::indicator_worker::a_ladder_respects_its_budget_and_always_reaches_the_newest_print` |
| The lane's live end is the preview | `app::indicator_worker::the_lane_samples_end_where_the_preview_does` |
| Dropping the lane clears the curve | `app::indicator_worker::dropping_the_lane_clears_the_samples_already_published` |
| Tape time maps linearly and clamps | `app::indicator_render::a_lane_maps_tape_time_linearly_and_clamps_outside_its_window` |
| Committed values step instead of interpolating | `app::indicator_render::committed_values_step_across_the_lane_instead_of_interpolating` |
| Bar-shaped plots stay out of the lane | `app::indicator_render::bar_shaped_plots_stay_out_of_the_lane` |
| No lane, no budget | `app::pane::the_rung_budget_follows_the_lane_and_is_zero_without_one` |
| A body drag pans the pane and the shared time axis | `app::dragging_a_pane_body_pans_that_pane_and_the_shared_time_axis` |
| Time pans once however many panes are stacked | `app::a_pane_drag_pans_time_once_however_many_panes_are_stacked` |
| Double click returns a pane to auto-fit | `app::double_clicking_a_pane_body_returns_it_to_auto_fit` |

### DIFF
--- a/crates/app/src/app.rs
+++ b/crates/app/src/app.rs
@@ -3903,6 +3903,146 @@ mod tests {
         areas.pane_gutters[index]
     }
 
+    /// The plot band of pane `index` — where its curve is drawn, as opposed to
+    /// the gutter where its numbers are.
+    fn pane_body(app: &QuantickApp, index: usize) -> egui::Rect {
+        let pane = &app.active_tab().flow_pane;
+        let areas = plot_split(
+            pane.last_plot_area.expect("a frame has been drawn"),
+            pane.live_strip_width(),
+            pane.indicators.visible_panes().count(),
+        );
+        areas.indicator_panes[index]
+    }
+
+    /// The bar the viewport has at its right edge — how far the chart is
+    /// panned along time.
+    fn right_edge(app: &QuantickApp) -> f32 {
+        let pane = &app.active_tab().flow_pane;
+        pane.viewport.right_edge_bar(pane.slots())
+    }
+
+    /// A pane body is a piece of the chart, so the chart's own gestures have
+    /// to work on it: drag to move, and the pane's own axis for the vertical
+    /// half. Before this the body answered nothing at all — the only way to
+    /// move a pane's curve out of its own way was to travel to the gutter on
+    /// the far side of the tape and drag it there.
+    #[test]
+    fn dragging_a_pane_body_pans_that_pane_and_the_shared_time_axis() {
+        let (mut app, _cmd_rx) = app_with_history(200);
+        let ctx = egui::Context::default();
+        let flow = add_pane_indicator(&mut app, "cvd", (0..200).map(f64::from).collect());
+        let other = add_pane_indicator(&mut app, "delta", (0..200).map(f64::from).collect());
+        run_frame(&mut app, &ctx); // the frame that fits each pane and records it
+
+        let body = pane_body(&app, 0);
+        let (lo, hi) = pane_range(&app, flow);
+        let edge_before = right_edge(&app);
+
+        drag_chart(
+            &mut app,
+            &ctx,
+            body.center(),
+            body.center() + egui::vec2(40.0, 30.0),
+        );
+
+        let (panned_lo, panned_hi) = pane_range(&app, flow);
+        assert!(
+            ((panned_hi - panned_lo) - (hi - lo)).abs() < 1e-6,
+            "a pan moves the window without resizing it: {lo}..{hi} -> {panned_lo}..{panned_hi}"
+        );
+        assert!(
+            panned_lo > lo,
+            "the candles' direction: pull the content down and the window climbs ({lo} -> {panned_lo})"
+        );
+        // Not the neighbour's resolved range: panning time changes what is
+        // visible, so every pane still on auto legitimately refits. What must
+        // not happen is the neighbour being taken off auto by a drag that was
+        // never over it.
+        assert!(
+            pane_is_auto(&app, other),
+            "one pane, one scale: the neighbour still fits its own values"
+        );
+        assert!(
+            !pane_is_auto(&app, flow),
+            "and the dragged one took control"
+        );
+        assert!(
+            app.active_tab().flow_pane.price_view.is_auto(),
+            "and the candles' own price scale is not a pane's to move"
+        );
+        assert!(
+            (right_edge(&app) - edge_before).abs() > f32::EPSILON,
+            "time is shared: the sideways half of the drag moved the chart"
+        );
+    }
+
+    /// Time is moved once per drag, not once per pane. Three stacked panes
+    /// answering the same sideways drag would pan the chart three times, and
+    /// the bars would run away from the pointer.
+    #[test]
+    fn a_pane_drag_pans_time_once_however_many_panes_are_stacked() {
+        let ctx = egui::Context::default();
+
+        let mut travelled = Vec::new();
+        for panes in [1_usize, 3] {
+            let (mut app, _cmd_rx) = app_with_history(200);
+            for index in 0..panes {
+                add_pane_indicator(
+                    &mut app,
+                    &format!("pane{index}"),
+                    (0..200).map(f64::from).collect(),
+                );
+            }
+            run_frame(&mut app, &ctx);
+
+            let body = pane_body(&app, 0);
+            let before = right_edge(&app);
+            drag_chart(
+                &mut app,
+                &ctx,
+                body.center(),
+                body.center() + egui::vec2(40.0, 0.0),
+            );
+            travelled.push(right_edge(&app) - before);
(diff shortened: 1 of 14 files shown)


milocaetano:
## arch-review — `main...feat/pane-live-lane`

Clean. No blockers.

**Docking** — the ladder is a port, not a CVD feature: `IndicatorHost::walk_partial_prefixes` hands each rung out through `PreviewsAt`, and `a_failing_instance_drops_out_of_the_ladder_without_taking_its_neighbour` exercises it with a second (failing) implementation. A new indicator gets its lane curve for free; a new *plot shape* on the lane is one arm of the `matches!` in `draw_lane_plots`, deliberately narrow because a histogram has no bar width to honour on a time axis.

**Determinism** — the engine change is a pure move: `open_bar`/`extend_bar` become `Bar::opened_by`/`Bar::extend`, and `folding_trades_by_hand_reaches_the_builders_own_partial` is the guard that no builder grew a private variant of the fold. Saturating side totals preserved and tested.

**Performance** — the ladder is walked once per drained worker batch, never per print or per frame; `MAX_LANE_RUNGS = 64` bounds it and `lane_rungs()` scales it to the lane's pixel width (0 = no lane, no ladder). `bench_lane_ladder` reports the walk cost. The render thread lost a per-frame mutex wait: `live_lane_width_px` + `live_end_ms` became one `live_lane()` read.

**Proof** — `every_rung_equals_a_standalone_preview_of_the_same_prefix` (no interpolation, no state leaking between rungs), `a_walk_leaves_committed_state_and_the_live_preview_untouched` (the walk is a read), `the_lane_samples_end_where_the_preview_does` (the curve's live end and the pane's headline are the same fact).

### Consider — the run clone is per drain and proportional to the forming bar

`crates/app/src/pane.rs` — `partial_command()` clones the forming bar's whole run of trades (`trades[len - trade_count..].to_vec()`) on every drain that took live trades, i.e. up to ~60 Hz, on the render thread. On a `time:1m` bar over BTCUSDT aggTrades that run is thousands of `Trade` (~56 B each), so the allocation is hundreds of KB per frame where it used to be one `Bar` clone.

It is correctly gated (nothing is cloned when `lane_rungs == 0`) and bounded by one bar, so this is not a regression the branch has to hold for. The cheaper shape is incremental: send only the trades that arrived this drain and let the worker keep the run, resetting it on `BarClosed` / `Rebuild` / `partial = None`. The measurement that settles the magnitude is the perf HUD frame time on the `dense tape btc` preset with a time or dollar spec, lane on.

Deferred deliberately — filed as follow-up rather than held against this branch.

PR #132, opened 2026-08-06
feat(app): the indicator band holds up in a small window

milocaetano:
> Stacked on #130 (`feat/pane-live-lane`). Both change `plot_split`,
> `split_panes` and `draw_pane`, so cutting this from `main` would conflict in
> exactly those functions. Review #130 first; this diff is `feat/pane-live-lane...HEAD`.

The band below the chart was written for a big window and had no answer for a
small one. Reported as "quando eu coloco janela menor, os dados do eixo X de
tempo comprime e fica feião" — and the axis was only the visible half of it.

## The three defects

**The time axis counted its labels instead of measuring them.**
`draw_time_strip` asked for a fixed ~6 (`step = (visible / 6).max(1)`) however
narrow the strip had become. Six `HH:MM:SS` want some 300 px of text, and the
history strip stops having that as soon as the live lane takes its share of a
smaller window. Counting labels cannot answer a question about pixels — and the
price axis one column over had solved this years of code ago with
`AXIS_LABEL_MIN_GAP_PX`.

**Panes were a bare fraction with no floor.** `PANE_HEIGHT_FRAC` per pane, up
to three, so a short window handed out three slivers and took 60% of the chart
to do it.

**Panes were the only band in the app that could not be resized.** The canvas
split has a draggable divider and so does the live lane. Same product, three
answers to "can I adjust this?".

## What changes

- **Labels are spaced by measured pixels**, from the candle width and the
  label's own width. Seconds are dropped before labels are: a strip too narrow
  for two full timestamps writes `HH:MM` rather than showing one lonely
  instant. A label must now fit *whole* inside the strip — containing only its
  centre let the one at the live end draw its right half over the gutter.
- **Nothing is drawn below `MIN_PANE_HEIGHT_PX`**, and a pane with no room
  becomes a labelled strip carrying its name and its live value. Collapsing
  hides the curve, never the number. The candles keep their own floor: panes
  are read *against* the bars.
- **A disclosure at each pane's corner** opens and closes it by hand, and an
  explicit choice is served before the automatic ones — otherwise clicking a
  strip open would be undone by the layout on the very next frame.
- **A pane's top edge is a resize handle**, with the resize cursor as the only
  affordance, which is the bargain the other two dividers already strike.
  Double click hands the pane back to the automatic layout. The floor is
  enforced by the layout rather than by the drag, so a divider cannot recreate
  the squeezed band this PR exists to prevent.
- **`QUANTICK_WINDOW_SIZE=WxH`**, registered in the `ui-harness` table.

## The number that mattered

`MIN_PANE_HEIGHT_PX` is added up, not guessed: a title row (28 px) + two axis
labels with their gaps and edge margins (32 px) + curve amplitude worth reading
a shape from (40 px) = **100 px**. The first value tried was 72 px, and a test
at the app's own minimum window showed three panes landing at ~81 px each —
above it. The floor would never have triggered at any window the app allows and
the whole rule would have been dead code. Deriving the number is what caught
that.

## Verification

- `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`,
  `cargo build --workspace`, `cargo test --workspace` — all green.
- Tests drive the app at `MIN_WINDOW` (900x560, the app's own minimum) rather
  than the roomy default. At the default nothing collapses, and a test that
  cannot see the defect cannot guard the fix — the first version of the
  collapse test passed vacuously for exactly that reason.

### Performance

Declared per touched path: the layout split and the label measurement are
**per frame** and stay O(panes ≤ 3) and O(labels); the divider drag is **per
frame**, O(1); no per-trade or per-depth path is touched. The one genuinely new
per-frame cost is text measurement, and it is at most three `layout_no_wrap`
calls of monospace samples, which egui serves from its galley cache. A per-frame
`Vec` allocation was *removed* on the way through (0dda7e9).

Measured: `APP_HEALTH_SUMMARY` `frame_cpu_ms`, same hooks, 40 health windows
each, alternating parent / branch four times on a live BTCUSDT tape.

| Run | parent | branch |
| --- | --- | --- |
| 1 | 2.24 ms (650 cells) | 2.59 ms (1036) |
| 2 | 2.55 ms (1337) | 3.52 ms (1142) |
| 3 | 2.94 ms (952) | 2.92 ms (1359) |
| 4 | 2.68 ms (1023) | 2.52 ms (700) |
| **mean** | **2.60 ms** (sd 0.25) | **2.89 ms** (sd 0.40) |

**The result is inconclusive, and that is the honest word for it.** The branch's
mean is 0.29 ms higher, which is 0.79 pooled standard deviations at n=4 — well
inside the run-to-run spread, and the branch happened to average 7% more
heatmap cells. Two branch runs at nearly identical load (1036 and 1142 cells)
differ by 0.93 ms between themselves, so within-arm variance swamps the
between-arm gap on this machine.

What can be said without a measurement: nothing added here scales with data
volume. If a reviewer wants a verdict rather than a range, the measurement to
run is a bench over a fixed frame rather than a live tape.

### Reviews, and what they changed

`arch-review` over this diff found a per-frame allocation: `pane_sizing`
returned a `Vec` and `plot_split` runs more than once per frame. Fixed in
0dda7e9 rather than deferred — it writes into a caller-owned array now.

**Visual QA found what no test could.** The disclosure chevron used the Unicode
geometric triangles, which are not in the bundled Phosphor font: every pane
header rendered as a tofu box. Fixed in f714878. This is the argument for
taking the screenshot even when the suite is green.

Also worth recording: the environment problem that blocked visual QA on #130
was **not** a limitation of this machine. It was the compositor going quiet on
an idle desktop. With the desktop active the capture runs at 59-60 fps and is
real — which also means #130's lane curve has now been seen working for the
first time, in the pane band of these captures.

### Known defects seen but not fixed here

At 900x560 the floating indicator legend and the aggression legend overlap the
candles, and the status bar's segments collide (`17/50 ticks10072 trades`).
Both are pre-existing and outside this branch's scope; they deserve their own
pass.

### Tests

| Behaviour | Test |
| --- | --- |
| A narrow chart gets fewer labels, never overlapping ones | `chart::a_narrow_chart_gets_fewer_labels_rather_than_overlapping_ones` |
| The stride survives degenerate geometry | `chart::the_stride_is_never_zero_however_degenerate_the_geometry` |
| Comfort is a floor on spacing, not a cap | `chart::very_wide_candles_still_leave_room_between_labels` |
| Seconds are dropped before labels are | `chart::a_strip_too_narrow_for_the_full_format_drops_the_seconds` |
| A label must fit whole, not just its centre | `chart::a_label_must_fit_whole_not_just_its_centre` |
| A short band collapses what it cannot draw legibly | `indicators::a_short_band_collapses_the_panes_it_cannot_draw_legibly` |
| The floors hold at every band height | `indicators::the_floors_hold_at_every_band_height` |
| A hand-collapsed pane stays collapsed in a tall band | `indicators::a_hand_collapsed_pane_stays_collapsed_in_a_tall_band` |
| A dragged height is honoured but never below the floor | `indicators::a_dragged_height_is_honoured_but_never_below_the_floor` |
| The smallest window collapses rather than squeezing | `app::the_smallest_window_collapses_rather_than_squeezing_every_pane` |
| A roomy window still draws every pane | `app::a_roomy_window_draws_every_pane` |
| Clicking a strip opens it, and it stays open | `app::clicking_a_collapsed_strip_opens_it_and_it_stays_open` |
| The disclosure closes as well as opens | `app::the_disclosure_closes_a_pane_as_well_as_opening_it` |
| Dragging a top edge resizes that pane | `app::dragging_a_panes_top_edge_resizes_that_pane` |
| A divider cannot be dragged past the floor | `app::a_divider_cannot_be_dragged_past_the_readable_floor` |
| Double click returns a pane to automatic | `app::double_clicking_a_divider_returns_the_pane_to_automatic` |
| The window hook is honoured and floored | `main::a_requested_size_is_honoured_and_floored_at_the_windows_own_minimum` |
| A malformed size is rejected, not guessed at | `main::a_malformed_size_is_rejected_rather_than_guessed_at` |

### DIFF
--- a/.claude/skills/ui-harness/SKILL.md
+++ b/.claude/skills/ui-harness/SKILL.md
@@ -25,6 +25,7 @@ is the source of truth, this table is the index):
 | --- | --- |
 | `QUANTICK_CONFIG=<toml>` | full feed/symbol config override (use a scratchpad toml; **never** edit the user's root `quantick.toml`) |
 | `QUANTICK_DEFAULT_FEED` / `QUANTICK_DEFAULT_SYMBOL` | startup feed/symbol |
+| `QUANTICK_WINDOW_SIZE=WxH` | the size the window opens at, floored at the app's own 900x560 minimum. Window size decides whether the indicator band has room for its panes and the time axis for its labels, so without this that whole class of defect is invisible to anything but a human dragging a corner. With it plus `QUANTICK_INDICATORS_AUTOSTART`, the collapsed-pane strip is reachable from a fresh launch. |
 | `QUANTICK_BOOK_AUTOSTART=1` | L2 heatmap layer |
 | `QUANTICK_LIVE_STRIP_AUTOSTART=1` | live strip |
 | `QUANTICK_BUBBLES_AUTOSTART=1` | aggression layer (bubbles + live-column footprint) |
--- a/crates/app/src/app.rs
+++ b/crates/app/src/app.rs
@@ -181,17 +181,23 @@ const STYLE_LOG_DEBOUNCE: Duration = Duration::from_millis(350);
 /// `live_strip_width` of zero means the strip is off and the chart runs
 /// straight into the gutter, exactly as it did before the strip existed.
 ///
-/// `pane_count` is the number of *visible* pane indicators: the band they
-/// claim is carved here, once, rather than by each caller — a chart rect that
-/// two call sites disagree about is two price scales for the same pixels.
-pub fn plot_split(area: egui::Rect, live_strip_width: f32, pane_count: usize) -> PlotAreas {
+/// `pane_sizing` carries one entry per *visible* pane indicator, top to
+/// bottom: the band they claim is carved here, once, rather than by each
+/// caller — a chart rect that two call sites disagree about is two price
+/// scales for the same pixels. Sizing rather than a count, because how tall a
+/// pane is and whether it has room to be drawn at all is the same decision.
+pub fn plot_split(
+    area: egui::Rect,
+    live_strip_width: f32,
+    pane_sizing: &[crate::indicators::PaneSizing],
+) -> PlotAreas {
     let plot = area.shrink(16.0);
     let strip_width = live_strip_width.max(0.0);
     let gutter_x = (plot.right() - AXIS_GUTTER).max(plot.left() + 20.0);
     let split_x = (gutter_x - strip_width).max(plot.left() + 20.0);
     let split_y = (plot.bottom() - TIME_STRIP).max(plot.top() + 20.0);
     let body = egui::Rect::from_min_max(plot.min, egui::pos2(split_x, split_y));
-    let (chart, indicator_panes) = crate::indicators::split_panes(body, pane_count);
+    let (chart, indicator_panes) = crate::indicators::split_panes(body, pane_sizing);
     // The gutter is banded exactly like the body it labels: the candles' price
     // scale owns the height of the candles and not a pixel more, so a drag
     // over a pane's numbers can only ever move that pane.
@@ -200,7 +206,7 @@ pub fn plot_split(area: egui::Rect, live_strip_width: f32, pane_count: usize) ->
     };
     let pane_gutters = indicator_panes
         .iter()
-        .map(|pane| band(pane.top(), pane.bottom()))
+        .map(|pane| band(pane.rect.top(), pane.rect.bottom()))
         .collect();
     PlotAreas {
         chart,
@@ -319,7 +325,7 @@ pub struct PlotAreas {
     pub chart: egui::Rect,
     /// Stacked indicator panes below the candles, top to bottom. Empty when
     /// no pane indicator is visible.
-    pub indicator_panes: Vec<egui::Rect>,
+    pub indicator_panes: Vec<crate::indicators::PaneSlot>,
     /// The gutter band beside each pane, in the same order: where that pane's
     /// value labels are drawn and where its own zoom gesture lives.
     pub pane_gutters: Vec<egui::Rect>,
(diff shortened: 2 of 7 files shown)


milocaetano:
## arch-review — `main...feat/pane-layout-ux`

No blockers. One Should-fix, applied on the branch (see below).

**Docking** — `PaneSizing` is the port the band was missing: `Auto` / `Manual(px)` / `Collapsed`, resolved in one pure `split_panes`, so a fourth way to size a pane (persisted layout, a preset, a keyboard command) is a new variant plus its `desired()` arm and nothing else. `TimeLabelFormat` is the same shape for the time axis — `ALL` is walked longest-first, so a `HH` form is one variant and one `sample()`. Neither the renderer nor the input pass grew a branch on "which window size is this".

**Performance** — `plot_split` runs more than once per frame, so `pane_sizing` writes into a caller-owned `[PaneSizing; MAX_PANES]` instead of returning a `Vec`: no allocator per frame for at most three copies of an eight-byte enum. The time strip trades a fixed label count for two `layout_no_wrap` measurements per frame (one per format, monospace so one sample answers for every label) and gains a stride that cannot produce overlap — a per-frame cost paid to remove a per-frame defect. `split_panes` allocates one `Vec<f32>` of at most `MAX_PANES`; flat in practice.

**Proof** — `the_smallest_window_collapses_rather_than_squeezing_every_pane` and `the_floors_hold_at_every_band_height` (the reported defect, plus tiling and no-seam invariants at every height), `a_divider_cannot_be_dragged_past_the_readable_floor` (the drag cannot recreate the state the branch exists to prevent), `clicking_a_collapsed_strip_opens_it_and_it_stays_open` (the automatic rule must not undo the click on the next frame — the subtle one), `a_narrow_chart_gets_fewer_labels_rather_than_overlapping_ones` and `the_stride_is_never_zero_however_degenerate_the_geometry`.

Nothing hardcoded: `MIN_PANE_HEIGHT_PX` is derived in a table in its own doc comment and justified against the app's own minimum window (three panes at ~81 px each — a floor below that would never bite), `COLLAPSED_PANE_HEIGHT_PX`, `MIN_CHART_HEIGHT_PX`, `TIME_LABEL_MIN_GAP_PX`, `TIME_LABEL_SPACING_PX`, `PANE_DIVIDER_HANDLE_PX` and `DEFAULT_WINDOW_PX` / `MIN_WINDOW_PX` are all named with units. The `QUANTICK_WINDOW_SIZE` hook is registered in `ui-harness/SKILL.md`, as the skill requires of every new surface.

Gesture registration order is stated where it matters (gutter → body → disclosure → dividers, later claim wins the overlap in egui), which is the trap this area has hit before.

### Fixed on the branch — misplaced doc block

`crates/app/src/main.rs` — the doc comment written for `window_size()` had landed on `DEFAULT_WINDOW_PX`, which then carried two contradictory summary lines and left the function undocumented. Moved to the function it describes; the constants keep their own one-liners. Zero runtime cost, so under the priority order there is no reason to leave it.

PR #133, opened 2026-08-06
feat(app): the indicator band holds up in a small window

milocaetano:
The small-window work of #132, replanted straight onto `main` so it can land
without the live-lane branch (#130) in front of it. Same intent, smaller
surface.

Fixes the reported defect — "quando eu coloco janela menor, os dados do eixo X
de tempo comprime e fica feião" — plus the usability of the band underneath it.

## What changes

- **Time labels are spaced by measured pixels** instead of a fixed count of
  ~6 (`step = (visible / 6).max(1)`). Six `HH:MM:SS` want some 300 px of text,
  and the history strip stops having that on a smaller window, so they
  collided. Seconds are dropped before labels are, and a label must now fit
  *whole* — containing only its centre let the one at the live end draw its
  right half over the gutter.
- **Panes get a readable floor.** Nothing is drawn below `MIN_PANE_HEIGHT_PX`,
  and a pane with no room becomes a labelled strip carrying its name and its
  live value. Collapsing hides the curve, never the number. The candles keep a
  floor of their own: panes are read *against* the bars.
- **A disclosure at each pane's corner** opens and closes it by hand, with an
  explicit choice served before the automatic ones — otherwise clicking a strip
  open would be undone by the layout on the very next frame.
- **A pane's top edge is a resize handle**, the grammar the canvas split and
  the live lane already use, with the floor enforced by the layout rather than
  the drag.
- **`QUANTICK_WINDOW_SIZE=WxH`**, registered in the `ui-harness` table. Window
  size is the trigger for this whole class of defect and nothing could ask for
  a small one before.

## The number that mattered

`MIN_PANE_HEIGHT_PX` is added up, not guessed: title row (28 px) + two axis
labels with their gaps and margins (32 px) + curve amplitude worth reading a
shape from (40 px) = **100 px**. The first value tried was 72 px, and a test at
the app's own minimum window showed three panes landing at ~81 px each — above
it. The floor would never have triggered and the rule would have been dead code.

## Difference from #132

Replanting onto `main` drops what belonged to #130: the pane body's vertical
pan and shared time gesture, and the lane segment inside the pane. Those stay
on `feat/pane-live-lane` and will need rebasing on top of this.

## Verification

- Four checks green **on this replanted branch**, not inherited from #132 —
  the resolution produced code neither had compiled before.
- Tests drive the app at `MIN_WINDOW` (900x560, the app's own minimum). At the
  roomy default nothing collapses, and a test that cannot see the defect cannot
  guard the fix.
- Visual QA on the #132 form of this work caught a tofu chevron (the Unicode
  triangles are not in the bundled Phosphor font); the fix is included here.
- Performance on #132 measured inconclusive (+0.29 ms, 0.79 sd at n=4 per arm)
  against its own parent. Nothing added scales with data volume, and a per-frame
  `Vec` allocation was removed on the way through.

### Known defects seen but not fixed here

At 900x560 the floating legends overlap the candles and the status bar's
segments collide (`17/50 ticks10072 trades`). Both pre-existing.

https://claude.ai/code/session_015E5WCdrsTG5XitARmMDnsr

### DIFF
--- a/.claude/GOAL-archive-pane-layout-ux.md
+++ b/.claude/GOAL-archive-pane-layout-ux.md
@@ -1,0 +1,122 @@
+# Goal
+
+Make the indicator pane band below the chart hold up at any window size:
+nothing rendered below the height at which it can be read, manual control over
+the split, and a time axis that measures its own labels instead of counting
+them.
+
+Branch `feat/pane-layout-ux`, worktree `../quantick-worktrees/feat-pane-layout-ux`,
+stacked on `feat/pane-live-lane` (PR #130) — both change `plot_split`,
+`split_panes` and `draw_pane`, so cutting from `main` would conflict in exactly
+those functions.
+
+## The defects this fixes
+
+- **D1 — the time axis counts labels instead of measuring them.**
+  `pane.rs:draw_time_strip` uses `step = (visible / 6).max(1)`: always ~6
+  labels whatever the width. An `HH:MM:SS` in monospace 10px is ~50 px, and the
+  history strip is narrower than the chart because the live lane takes its
+  share — so on a smaller window the labels collide. The price axis already
+  solved this with `AXIS_LABEL_MIN_GAP_PX`; two axes, two rules.
+- **D2 — pane height is a bare fraction with no floor.**
+  `PANE_HEIGHT_FRAC = 0.20` of the body *per pane*, up to `MAX_PANES = 3`, so
+  three panes take 60% of the chart at every size and each one still has to
+  hold a rule, a title, a headline value, gridlines and a curve.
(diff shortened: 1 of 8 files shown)


milocaetano:
Closing as a duplicate of #132, which carries the same five commits.

This branch was the same work rebased straight onto `main` *without* the live-lane change, as an alternative to the stack. #130 is now merged, so #132 sits on `main` with its dependency in place, has run green on the four checks with `main` merged in, and is the version that ships. The extra `chore(app): drop a test helper the lane work took with it` commit here only existed because this branch lacked that lane work — the suite is green on #132 without it.

Nothing in this branch is lost.

PR #142, opened 2026-08-08
feat(app): profile draws as a silhouette over the liquidity heatmap

milocaetano:
## What

Stacked on #140. With the L2 heatmap on, the fixed-range profile's translucent fill composes into the map's cells and disappears — the worst case measures **1.002:1 contrast** against the ramp's cyan plateau (the fill's alpha signal is `α·(fg−bg)`, which is zero when the ink meets its own hue). This PR makes the profile keep reading over the map without touching the map.

## The design-panel decision

Chosen by a three-seat panel (order-flow trader, color/contrast specialist, UX designer) run over the real scene and the render code; full proposals compared before coding:

- **Over candles**: unchanged — fill, alphas, everything as today. Nothing to relearn where there is no map.
- **Over the map**: the fill gives way to the histogram's **staircase envelope, double-stroked** — a `#05070C` casing under the object's own color. The pair survives every stop of the ramp: 12:1 on the cyan plateau, ~24:1 on the yellow band.
- **Value area by weight, not alpha**: in-VA ink 1.9px full brightness, out-VA 1.25px at 60% — an axis the heatmap does not occupy.
- **POC/VAH/VAL keep their fill** with an always-on casing (POC `#FFD54F` vs the ramp's yellow band is **1.05:1**, the worst number of the scene; the casing carries it at ~21:1). Dashed casings share phase with the VA dashes so the map shows through the gaps.
- **Labels and status get a glyph knockout** (same galley painted 4× offset in casing + 1× in color) — no plates, and it fixes label legibility everywhere, not only over the map.
- **Zero uncovered map pixels altered.** No scrim, no dimmer — the panel's unanimous red line: attenuated cells would fake liquidity out of the book ("is that wall thin, or did the app dim it?").
- **Spoken**: the status chain gains `· outline over heatmap` while the switch is active — same grammar as `grouped` / `to live` / `side inferred`.
- **Inspector**: `outline over liquidity map` (default on; off = always fill). Rides presets; older presets stay loadable (serde default).

Rejected alternatives, for the record: scrim behind the object (erases the map exactly where the trader anchored), per-region dimmer (visual dishonesty — breaks the eye's absolute-intensity calibration), fill+knockout only (keeps two translucent fields fighting).

## Mechanics

- The cut is a **single x per frame**: the map's oldest covered slot, read from the projection (`VisibleOrderflow::first_heat_slot`, O(cells), computed only while a profile object exists and depth is visible) and carried on the FRVP cache **beside** the key — the map growing moves the cut, never re-merges the fold.
- Paint converts slot→x through the anchors' own affine map (the same trick the developing mode's live edge uses) — no new plumbing through `DrawContext`.

## Declared performance (per-frame path)

- Per trade: zero (unchanged).
- Per frame without profile objects: zero added — the O(cells) scan is gated on `wants_range_profile()`.
- Per frame with a profile over the map: the fill pass is clipped at the cut (cheaper), plus ~2N+2 silhouette segments × 2 strokes (N = visible rows); the color panelist's accounting has the silhouette *cheaper* than the fill it replaces. `frame_cpu_ms≈4.4` under live tape + book in the harness run; before/after capture pending user-present desktop (idle desktop renders white — known environment state), will be attached as a comment.

## Verification

- fmt / clippy `-D warnings` / build / test green — 952 app tests (2 new: heat boundary rides the cache without re-keying; preset round-trip covers the new flag), 72 engine.
- Scene reachable by hooks: `QUANTICK_BOOK_AUTOSTART=1` + `QUANTICK_FRVP_DEMO=1`.
- visual-qa matrix (pending capture): fill left of cut / silhouette right, POC casing over yellow band, knockout labels, `Always fill` off-switch, slow-pan flicker at the map edge (hysteresis is the plan-B if it flickers).


milocaetano:
## Visual + perf evidence (harness run, live BTCUSDT + L2 book)

Scene: `QUANTICK_FRVP_DEMO=compare` + `QUANTICK_BOOK_AUTOSTART=1` — two adjacent 25-bar profiles over the same stretch of dense map (2057 cells, bubbles layer on), one per over-heatmap mode, single frame.

**Outline (Auto, left object):** POC line + `POC 65179` label fully legible over dense cells (glyph knockout carrying it), VAH/VAL dashes readable, status line (`vol 31.44 · Δ 7.15 · rows 1`) readable directly over the map. No uncovered map pixel altered inside the range.

**Always fill (right object, the before):** the fill visibly washes into the cells around 65164 — the 1.002:1 composition failure this PR removes as the default.

**Performance (APP_HEALTH_SUMMARY, user-present desktop):** 59 fps, frame_avg 16.7 ms (vsync-locked), frame_cpu 5.7 ms with 2 profiles + 2057 cells + aggression bubbles; earlier single-profile run: frame_cpu 4.4 ms @ 286 cells. Consistent with the declared accounting (silhouette strokes ≤ the fill quads they replace; O(cells) boundary scan gated on a profile existing).

Deferred watch item for visual-qa: cut-boundary flicker under slow pan at the map edge (plan B: pixel hysteresis on the cut x).

Captures archived in the session scratchpad (`frvp_compare2.png`, `frvp_over_heatmap_after.png`).

Issue #148, opened 2026-08-08
Volume profile over venue history: approximated ladder from OHLCV, honestly labeled


milocaetano:
## Problem

On time charts the history is venue klines (25k+ candles on 5m) with only the newest bars tape-built. A fixed-range profile over that history honestly says **"no tape in range"** — correct, and useless for the time-chart trader. Screenshot evidence in PR #142's session: a full-chart range with zero data.

## Proposal (user's own suggestion, refined by the panel discussion)

Approximate the ladder for bars that have no tape, from what the candle does carry:

- **Engine** (test-first): `approximate_ladder(bar, group) -> BarFootprint`-shaped fold — the bar's volume distributed uniformly across the buckets its high–low range crosses (exact Decimal division, remainder to the close's bucket so totals conserve). Binance klines carry taker-buy volume, so the buy/sell split may be real even when the price distribution is approximated — keep the two axes of honesty separate (price approximated ≠ side approximated).
- **Refresh**: prefix slots fold approximated ladders instead of contributing nothing; `bars_covered` counts them separately, e.g. cache gains `bars_approximated`.
- **Honesty (the point)**: status line gains `· approximated from OHLC (N of M bars)` in the same chain as `side inferred`/`grouped`. Delta/buy-sell split is disabled whenever any approximated bar lacks a real side split — a partial delta shown as whole is a lie. Value area computed over the mixed fold is fine (volume totals are real), but the label stays on.
- **Config**: on by default? Panel question — a profile silently mixing approximated data vs a trader opting in. Suggest: on by default *because* the label speaks, off-switch in the Profile tab ("approximate from candles").
- Session/composite/VPVR profiles inherit this for free once it lives in the refresh.

## Acceptance sketch

- Engine tests: uniform spread conserves volume exactly; one-bucket candle; remainder lands in the close bucket; taker-buy split preserved when the bar carries it.
- App: mixed range (prefix + tape) shows both counts in the status; delta hidden when any approximated bar has no side split; presets untouched.
- Gates: 4 checks, arch-review, trader-ux (label legibility), visual-qa (mixed-range scene via a QUANTICK_* hook), PR.

Depends on #140/#142 (the profile itself).

milocaetano:
## Precision tiers (how the incumbents actually do it)

[...]

PR #149, opened 2026-08-08
feat(app): venue history joins the range profile as approximated ladders (#148 v1)

milocaetano:
## What

Closes the v1 tier of #148. A range profile over the time chart's venue history (klines, no tape) was honest and empty — now those candles join the fold as **approximated ladders**, spoken as such.

## Mechanics

- **Engine** (`BarFootprint::approximated`, test-first): the candle's volume spread uniformly over the buckets its high–low crosses; shares rounded **down** and remainders landing in the close's bucket, so every total conserves *exactly* (proved by test). The venue's real taker-buy split (Binance klines carry it) spreads the same way — sides stay venue-reported even where placement is approximated. Spans wider than the cap coarsen first and report aggregated.
- **App**: prefix slots fold prebuilt ladders (`frvp::ApproxLadders`, cached per (group, prefix length) — arch-review caught the per-re-key rebuild storm on 25k-candle charts and it builds once now); exact and approximated bars counted apart.
- **Honesty**: status chain gains `· approximated from OHLC (K of M bars)`; `approximate from candles` toggle on the Profile tab, preset-safe, off restores exactly the previous behaviour.

## Declared deviation (arch-review)

This defaults **on**, which changes what a prefix-range profile shows (histogram instead of "no tape in range"). Deliberate and user-requested: the label speaks whenever approximation happens, and the off-switch is one checkbox.

## Declared performance

- Per trade / per frame: zero added.
- Ladder build: O(prefix) once per backfill page or group refold (~25k small ladders, one-time), then borrowed.
- Fold on key change: unchanged shape; deep-drag over a 25k-bar range is a large merge by nature — incremental prefix sums are the known follow-up, noted in #148.

## Verification

- fmt / clippy -D warnings / build / test green: engine 76 (4 new: conservation, one-row, zero-volume, oversized-span), app 955 (2 new: mixed-range counts, off-switch restores).
- Scene: existing `QUANTICK_FRVP_DEMO=1` straddles the prefix seam and photographs the mixed label.
- Stacked on #142 (→ #140). v2 (intrabar subdivision) and v3 (real aggTrades backfill) tracked in #148.

### DIFF
--- a/crates/app/src/drawings/fixed_range_profile.rs
+++ b/crates/app/src/drawings/fixed_range_profile.rs
@@ -90,6 +90,10 @@ pub struct FrvpCache {
     pub empty: Option<FrvpEmpty>,
     /// Bars whose ladders went into the fold (the partial counts once).
     pub bars_covered: usize,
+    /// Bars folded from an **approximated** ladder — venue candles with no
+    /// tape, their volume spread over their own high–low. Spoken by the
+    /// status line, never blended away.
+    pub bars_approximated: usize,
     /// Bars the anchors span on the chart, prefix candles included.
     pub bars_total: usize,
     /// The oldest global slot the L2 heatmap covers this frame — where the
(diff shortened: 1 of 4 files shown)


milocaetano:
Superseded by #151 (same content rebased onto main after the stack cascade auto-closed this one) — merged.

PR #160, opened 2026-08-11
fix: bubble size means the same quantity at every zoom

milocaetano:
## Problem

Zooming rescaled every aggression bubble: the automatic size references (`VisibleP99` / `VisibleMax`) were computed over the clusters visible in the current window (`size_reference` in `orderflow/projection.rs`), so zooming out let bigger clusters into view, raised the reference, and shrank every bubble on screen. The same aggression read smaller for no reason a trader could see — reported live with screenshots showing one print changing size across three zoom levels.

## Fix

The size scale is now a statement about the **session**, never the screen. `SessionScale` (new module, `orderflow/scale.rs`) accumulates every recorded print into a cluster-quantity distribution as it arrives — clustered at native price grouping + `bubble_cluster_ms`, the capture resolution — and the projection reads P99/max from it. Zoom still merges clusters via display grouping, and a merged cluster now always reads **bigger** than the prints it swallowed, never smaller; anything the viewport merges past the scale saturates at max radius. The scale never shrinks: retention trims what the chart draws, not what a quantity means. It resets only with the session itself (symbol switch, price-grouping reset) or when the clustering window changes.

UI labels updated to match ("Auto · session P99" / "Auto · largest in session"); wire format (`visible_p99`) and shipped presets unchanged, so existing configs load as-is.

## Performance (paths classified by rate)

- `SessionScale::record` — **per trade**: 2–3 `BTreeMap` ops, O(log n).
- `p99()`/`max()` — per settled rebuild (≥220 ms cadence): walks ~1% of distinct quantities.
- Settled rebuild — no longer clusters the live half except when the candle summary needs its pie.
- Rebuild on `bubble_cluster_ms` change — rare: O(retained prints), documented.

Measured, 90 s dense replay (100k-print synthetic tape, 4×, ~615 trades/s), debug build, same machine:

| | main | fix (first cut, rejected) | fix (shipped) |
|---|---|---|---|
| fps avg | 59.56 | 59.51 | **59.64** |
| frame_cpu avg | 2.51 ms | 2.54 ms | **2.35 ms** |
| projection_ms | 2.5–4.4 flat | 4.8 → **29.8 growing** | 2.3–3.5 flat |

The first cut reclustered the whole retained tape per rebuild — caught by this measurement and replaced with the incremental scale before shipping.

## Tests (written first, failing on main)

- `a_prints_bubble_keeps_its_size_when_the_window_zooms_out` — same quantity, same normalized size through two price windows, all three reference modes (main: 1.0 → 0.14).
- `a_cluster_merged_by_zooming_out_reads_bigger_not_smaller` — adaptive grouping merges prints; merged bubble must outsize its parts (main: 0.32 < 1.0).
- `SessionScale` unit tests: clustering rules, P99 rank parity with the projection's percentile, open-cluster recount, dust ignored.
- History: eviction never moves the scale (ratchet); `bubble_cluster_ms` change rebuilds from retained prints; grouping reset clears it.

## visual-qa

Same deterministic recording (playhead reproducible to ±1 print across runs), fix + main control, three window sizes (1000×700 / 1600×900 / 2200×1200), two frames 1.2 s apart per cell, health-gated (fps 59–60 in every run):

- **Scale stability: PASS** — pixel-measured diameters of the same bubbles in the shared tape region: 30/32/34 px class identical across window sizes on the fix.
- **Settings tab (`QUANTICK_DOCK_TAB=bubbles`): PASS** — "Auto · session P99" renders untruncated.
- **Motion: PASS** — 8% of pixels changed between paired frames, bbox confined to the chart, no chrome movement.
- **Integrity/small window: PASS** — 1000×700 collapses the toolbar into overflow correctly, nothing clipped.
- Honest note: the main control did not *visibly* reproduce the shrink on the synthetic tape (its sweeps are uniform, so every window holds some and visible P99 ≈ session P99). The defect's proof is the unit test above and the original live screenshots.

## trader-ux-review

No Blockers, no Should-fixes. One Consider, deferred: under "largest in session" a monster sweep anchors the scale for the rest of the day (chosen semantics; the P99 default is robust to one outlier). A manual "reset scale" affordance is a possible follow-up.

## arch-review

Run over `git diff main...HEAD`. Two Should-fixes found and resolved in `34a34cc` (ratchet/rebuild tests; scale counted in the history memory estimate). Deferred, noted deliberately:

- `VisibleP99`/`VisibleMax` variant names are wire-compat legacy; documented at the enum.
- ~~Summary pies keep their visible-summary reference~~ — resolved in b35ff33 after live testing: pies now read on a session scale of per-minute level totals (`SummaryScale`), viewport-free like the print scale.
- The liquidity heatmap's `IntensityMode::VisibleP99` (cell colour, not bubble size) is still viewport-based — related but out of scope.

Blast radius: 1 file added, 5 edited (docs/rename, history wiring, projection consumption, UI labels).

https://claude.ai/code/session_01ACWmTJChLJiE84d27DVQB1

### DIFF
--- a/crates/app/src/orderflow/config.rs
+++ b/crates/app/src/orderflow/config.rs
@@ -317,17 +317,6 @@ impl ConsumptionMark {
     }
 }
 
-impl BubbleSizeReference {
-    /// Whether the reference is measured from the recorded prints, as opposed
-    /// to [`Fixed`](Self::Fixed) — pinned by the user so a bubble means the
-    /// same quantity across sessions and symbols, which is exactly why
-    /// nothing the renderer decides is allowed to rescale it.
-    #[must_use]
-    pub fn is_automatic(self) -> bool {
-        !matches!(self, Self::Fixed)
-    }
-}
-
 /// How a bubble's fill is painted.
 ///
 /// `Flat` is the classic solid disc. `Sphere` shades each bubble like a ball
--- a/crates/app/src/orderflow/config.rs
+++ b/crates/app/src/orderflow/config.rs
@@ -251,18 +251,24 @@ impl IntensityMode {
 
 /// How the quantity that maps to a full-size bubble is chosen.
 ///
-/// `VisibleP99` keeps one outlier sweep from shrinking every other print to a
-/// dot, at the cost of making the top 1% all render at the maximum radius.
-/// `VisibleMax` restores a strict ordering (the biggest print is the only one
-/// at full size); `Fixed` pins the scale so bubble size means the same thing
-/// across sessions and symbols.
+/// The automatic modes measure the whole recorded tape, never the visible
+/// window: zoom decides what is on screen, not what a quantity means, so the
+/// scale only moves when the recording itself does — a new print beyond the
+/// old reference, or retention trimming the floor. `VisibleP99` keeps one
+/// outlier sweep from shrinking every other print to a dot, at the cost of
+/// making the top 1% all render at the maximum radius. `VisibleMax` restores
+/// a strict ordering among recorded prints (clusters a zoomed-out view merges
+/// past it saturate at full size); `Fixed` pins the scale so bubble size
+/// means the same thing across sessions and symbols. The `Visible*` names are
+/// the wire format shipped presets already use; what they measure stopped
+/// being the viewport the day zoom stopped being allowed to rescale a print.
 #[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "snake_case")]
 pub enum BubbleSizeReference {
-    /// 99th percentile of the visible clustered prints.
+    /// 99th percentile of the recorded clustered prints.
     #[default]
     VisibleP99,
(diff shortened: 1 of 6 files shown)


milocaetano:
Live testing on the first build caught the deferral being wrong in practice: with `candle_summary` on (the shipped default), summary pies are most of what the chart draws — and they were still sized against the *visible* summaries, so a millimetre of zoom rescaled the chart anyway (pie size 1.0 → 0.2 across two price windows, now pinned by `a_pie_keeps_its_size_when_the_window_zooms_out`, which failed before the fix).

Pies now read on their own session scale, as viewport-free as the print scale: per-minute totals per native price level (`SummaryScale`), accumulated print-by-print with the same ratchet. The window is fixed tape time rather than the bar duration on purpose — a pie's area keeps its meaning across a bar-spec switch exactly as across a zoom. The visible-summary pass in `project_settled` is gone with the bug, and with it the live-half reclustering it required; `is_automatic()` and the cluster-slice `size_reference()` became dead and were removed.

PR #161, opened 2026-08-13
feat: copilot.pine — range-reversal context indicator

milocaetano:
## What

A new embedded Pine script, `copilot.pine`: an operate/don't-operate copilot for range-reversal order flow. Four layers, context to trigger, every cutoff an `input`:

1. **Regime** — last N bars must fit inside k ATRs; otherwise a red band paints the no-trade window.
2. **Structure** — double top/bottom via pivots with ATR tolerance, drawn with a line and a DT/DB label.
3. **Effort** — CVD divergence on the second touch + absorption (volume per point ≥ its average × threshold).
4. **Trigger** — delta z-score beyond the threshold between the touch and its confirmation.

A marker prints only when all four agree, on the confirming bar — never repainted.

## Also in this change

- **Dialect trap found and documented**: `and`/`or` short-circuit + `ta.*` call-site identity means a stateful kernel on the right of `and` warms up late and reads a gapped series — it silently muted the marker. Fixed by hoisting kernels to top level; one **Divergence** paragraph added to `docs/pine-dialect.md`.
- `crates/pine/tests/copilot_semantics.rs`: proves the SELL marker fires on the confirming bar of a divergent double top and stays silent without divergence (this test caught the bug above).
- `library.rs` test pinning the embedded copy to the corpus fixture so the two can't drift.

## Blast radius

Added: `crates/app/scripts/copilot.pine`, `crates/pine/tests/corpus/ok/copilot.pine`, `crates/pine/tests/copilot_semantics.rs`. Edited: `library.rs` (1 registration line + 1 test), `docs/pine-dialect.md` (1 paragraph). No engine, orderbook, sim or feed changes.

## Performance

Per-closed-bar/preview evaluation in the indicator host, only when the user adds the script. Zero new per-trade or per-frame cost. Under a 2 200-print replay at 100×: fps 59–60, frame_avg 16.7 ms (`APP_HEALTH_SUMMARY`).

## Verification

- `cargo fmt --check` / `clippy -D warnings` / `build` / `test --workspace`: all green (48 suites).
- arch-review: run over the final diff; one finding (fixture drift) resolved with the pinning test.
- visual-qa via ui-harness on a synthetic WINQ26 replay: script loads via `QUANTICK_INDICATOR_SCRIPTS_AUTOSTART`, legend well-formed; DB line+label and BUY marker render; regime band paints exactly the trending stretch; a malformed file degrades to a visible error slot; defaults on a short tape honestly draw nothing (warmup). SELL/DT is the mirrored code path of BUY/DB, proven by the semantics test.
- trader-ux-review: no Blocker / no Should-fix. Deferred Considers: on-chart explanation of the band (needs active `bgcolor` or per-plot legend text in the dialect); ATR-scaled tolerance can mark shallow structure on violent tapes (calibration via inputs).


milocaetano:
**Trader-first visual redesign** (after live feedback: the regime band read as phantom price action — a trader saw a "gap"):

- Regime area fill → **ribbon** of marks one ATR above the recent highs: teal dot = context OK, red cross = stand down, absence = warming up. Never covers a candle.
- **Flip hysteresis**: the verdict holds until N consecutive bars disagree (input, default 3) — no flicker on borderline bars.
- **"stop" line** drawn at the touch extreme when a setup fires — the structural invalidation, on the chart.
- Pivot confirmation default 5→3 bars: the marker lands while the aggression is still tradeable.
- Inputs renumbered into reading order: 1 context · 2 structure · 3 proof · 4 trigger · advanced.

Semantics tests extended (ribbon columns asserted); fixture re-synced; four checks green. This resolves the deferred UX Consider about the band being misread as price.

PR #163, opened 2026-08-13
fix: calibrate the mini index bubble preset against a real WIN session

milocaetano:
## What

Recalibrates the `mini index sweeps` bubble preset using a real recorded session (WINQ26 2026-08-12, 389 622 prints — downloaded via #162) instead of guessed values.

The floor filters what the pipeline actually measures — the (side, price bucket, 300 ms) **cluster** — so the numbers that matter are cluster quantiles: p50=6, p90=35, p95=56, p99=120, max=3415 contracts.

| knob | before | after | why |
|---|---|---|---|
| min_quantity | 10 | **15** | top ~quarter of clusters; ~37 marks/min at the quietest lunch hour — informative, never soup, never dark |
| size_reference | fixed 200 | **fixed 250** | ≈ cluster p99.9 — a full ball means the same thing all session |
| max_radius | 22 | **14** (φ ladder) | the old ceiling drowned candles at any zoom |
| labels | qty+count | **qty only, ≥ ~200 lots** | ink for whales, not for everything |
| summary | bare defaults | **cluster 300 ms · dust 1500 ms · candle pies on** | one sweep = one ball; history reads as summary |

Both copies of the presets file (`crates/app/config` embedded + repo root) move together — the sync test enforces it.

## Evidence

- Soup (floor 0, same tape): 27 fps / frame 36.7 ms / cpu 21 ms, candles unreadable.
- This preset: **60 fps / 16.7 ms / cpu 4.6 ms**, candles + structures readable, whale labels legible (screenshots in session).

## Verification

fmt / clippy / build / test all green (49 suites). One unrelated flake observed once in `quantick-app` during a full-workspace run (did not reproduce in three subsequent runs; not touched by this diff — config-only change).

### DIFF
--- a/config/bubbles.toml
+++ b/config/bubbles.toml
@@ -109,21 +109,21 @@ show_quantity_labels = false
 show_trade_count = false
 label_min_radius = 9.9
 
-# Calibrated against a real WIN session (WINQ26 2026-08-12, 389 622 prints).
-# What the floor filters is the *cluster* (one side, one price bucket, one
-# 300 ms window — a sweep is one ball), so the numbers that matter are the
-# cluster-quantity quantiles of that tape: p50=6, p90=35, p95=56, p99=120,
-# max=3415 contracts. The old thresholds (min_quantity 10 per cluster,
-# reference 200, max radius 22) still drew ~36% of all clusters and drowned
-# the candles; these draw roughly the top quarter, and never go dark at
-# lunch (~37 marks/min at the day's quietest hour):
-#   * min_quantity 15 per cluster — the mark means "someone swept", not
-#     "someone printed".
-#   * fixed reference 250 ≈ cluster p99.9 — a full-size ball keeps one
-#     meaning all session long instead of rescaling with what is visible.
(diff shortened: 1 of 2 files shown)


milocaetano:
Final numbers after live tuning + a 10-persona flow-trader panel (avg 4/5, 9/10 'tradeable today'): flat discs, 1 px rim, fixed 600 reference, floor 50, 2 px side separation, labels on whales. Panel's structural findings filed separately: absorption-mark emphasis, regime lane, candle z-order, zoom LOD (#164).

milocaetano:
For the record before merge — the trader identified that 2026-08-12 was the WINQ26 **rollover day** (389k prints, a dying contract): the calibration tape was atypical. The representative front-month day (2026-08-11, 1.4M prints) measures: cluster p50=30, p90=179, p95=282, p99=634, p99.9≈1200, max 59k — an equivalent-selectivity floor there is ~250, not 50. No further absolute retune: #169 (session-relative floor) supersedes any fixed number and carries both days' targets in its notes. Shipping as calibrated, documented.

PR #190, opened 2026-08-16
docs(arch-review): run the native code review as step 0

milocaetano:
## What

`arch-review` now runs the bundled `code-review` skill as **step 0**, before it
looks at a single file for shape.

The skill already declared correctness as priority 0 and then, two lines later,
left the bug pass to whoever remembered to type `/code-review`. One command now
covers both passes, in the order the priority list already claimed.

## The rule

- **Step 0** invokes `code-review` over the same target as the review scope
  (`main...HEAD`, a PR number, or the working diff), `high` effort for a branch
  or PR and `medium` for a small uncommitted diff, and waits for it.
- Confirmed correctness findings enter the arch-review as **Blockers**, listed
  before any shape finding. A branch does not pass with one open.
- Findings are cited by `file:line`, never restated — this review adds the six
  dimensions, it does not re-report bugs in other words.
- The review header names the effort level used and how many findings came
  back, including zero; if the skill could not run, it says so there.
- No `--fix` / `--comment` from inside the review: arch-review is the thing that
  reports.
- `Verify before reporting` gained item 0 (step 0 must have run — a shape review
  without it is incomplete, not clean) and the verdict opens with a
  **Correctness** line.

## Second commit — findings from reviewing the first one with its own rule

- **Scoping now fetches first.** `main...HEAD` with a stale local `main` shows
  other branches' merged work as this branch's — it did exactly that here, 26
  files from #188 — so the scope block starts with `git fetch origin` and says
  how to spot the case.
- **`CLAUDE.md` and `.claude/hooks/README.md` caught up.** The workflow rule
  names step 0, and the README says the one `arch-review-ok` marker now stands
  for both passes while the gate still only matches the sha.
- **The invocation is unambiguous** against a plugin skill of the same name.

## Verification

Documentation-only change under `.claude/` plus one line of `CLAUDE.md`; no
crate source touched.

- `cargo fmt --all -- --check` — clean.
- `clippy` / `build` / `test` left to CI: the diff contains no compiled code and
  a fresh `target/` in a worktree costs several GB on this machine.
- Step 0 on this very branch was launched and returned no output (background
  task lost), so the correctness pass here was manual over a markdown-only
  diff — noted per the rule this PR introduces.


milocaetano:
Step 0 ran on this branch: `code-review` at `medium`, 15 findings, 12 acted on in d0e704a. Two deferred, both deliberate:

- **`.claude/hooks/guardrails.sh:156,161`** — the deny messages still tell the agent to review `git diff main...HEAD` without a fetch. Left alone: the messages are string-matched by `guardrails_test.sh`, and the skill now fetches as its first scoping step, so the stale-main case is caught one line later anyway.
- **`git fetch` is not in the `.claude/settings.json` allowlist**, so the new mandatory fetch adds one permission prompt per review. Not changed here — widening the project allowlist is @milocaetano'''s call, not a side effect of a docs PR.

milocaetano:
Second review, this time over the PR itself: `code-review` at `high`, 14 findings, 13 acted on in 61222ec.

The load-bearing ones were about the rule leaking at its edges:

- The scope block fetched and then diffed `main...HEAD` — a fetch does not move the local `main`, so the inflated-diff bug this PR claims to fix survived its own fix. Ranges now name `origin/main`.
- `.claude/skills/mission/SKILL.md` (which grants the docs/skills waiver) and `.claude/skills/ship/SKILL.md` (which runs the pre-PR sequence) had never heard of step 0. A skills-only branch could take the waiver and skip the bug pass; `/ship` would push and open the PR while step 0 was still running in the background. Both updated.
- An empty code-review pass is now a scope failure to re-invoke, not a clean bill of health; a lost notification has a defined recovery; an efficiency finding lands in the performance dimension instead of being capped at Consider.

**One finding refuted:** `ReportFindings` was said to reject a `correctness` category. Its schema takes a free-form kebab-case slug and names `correctness` as an example — no change made.

PR #193, opened 2026-08-16
feat(paper): the aim rides the pointer, and a resting order tag rests small

milocaetano:
Two things on the chart-trading surface were fighting the price action.

## The aim label

It parked against the right edge whatever the pointer was doing, so aiming
at a price on the left of the plot meant crossing the whole chart to click
the thing you were already pointing at. It now rides beside the cursor — a
fixed gap to its left, flipping right near the left edge, never under the
crosshair it belongs to — with the dashed line running from under the
cursor out to the axis, so the label beside the hand and the price on the
gutter read as one statement.

**A label that follows the pointer can never be landed on** — move toward
it and it moves with you — so the click no longer needs it. The held
modifier is the deliberate act, the label states what the click will do,
and the hand cursor is up wherever the aim is. This was raised with the
user and chosen deliberately: the first design shipped a target that could
not be reached.

The aim is therefore the **last claimant on the canvas**, and stands *down*
rather than merely refusing its press — no line, no label, no hand cursor,
no place — wherever something already holds the pixel: an annotation a
press would grab or the canvas's own chrome (the pane answers that, exactly
as it answers "a tool is armed"), an armed limit/stop from the ticket, this
module's own lines and ✕s, and the layer switch. That last one matters:
hidden means unpainted, and with the label no longer the target, a hidden
layer would otherwise turn the whole plot into an invisible order button.

## The resting order tag

`#6 BUY STP 1 @ 180615 ✕` sat over the newest candles for as long as the
order worked. It now rests as `BUY LMT 1` — an order line is `ACCENT`
whatever its side, so side, kind and size are the only things the chart
cannot say otherwise; the price is already on the gutter chip and the id
only matters once you mean to act. The full statement, with its ✕, arrives
under the pointer, on the dock row's hover, and throughout a drag.

Roughly 72 px at rest against 215 px open (`final-pill.png` vs
`final-open.png`). **The cancel costs no extra gesture:** the ✕ sits at a
fixed x independent of the text, and the tag opens across the whole price
row, so it is already painted before the pointer arrives at it.

## Evidence

- **Four checks** green on a branch cut from `origin/main`: `cargo fmt --all
  -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`,
  `cargo build --workspace`, `cargo test --workspace` (59 suites, 1363 app
  tests).
- **Performance — declared and measured.** Every touched path is
  **per-frame**; nothing per-trade or per-depth is touched.
  `refresh_open_tags` is O(orders) and refills its buffer, so hovering an
  order allocates nothing after the first frame; the new yields in
  `compute_cmd_preview` run only while a modifier is down; the pane's gate
  costs three bool reads otherwise. The aim line now reaches the axis and
  pays up to 96 dash segments against `main`'s ≤40, bounded by
  `CMD_LINE_MAX_DASHES`.

  Paired, interleaved release runs under a live Binance tape, identical
  hooks (`QUANTICK_PAPER_DEMO=1`), `APP_HEALTH_SUMMARY`:

  | | fps | frame_avg_ms | frame_cpu_ms |
  |---|---|---|---|
  | `main` ×3 | 59 / 59 / 59 | 16.81 / 16.69 / 16.84 | 2.245 / 2.439 / 2.210 |
  | branch ×3 | 59 / 59 / 59 | 16.70 / 16.68 / 16.71 | 2.318 / 1.834 / 2.739 |

  Mean `frame_cpu`: **2.298 ms vs 2.297 ms** — flat, and the spread within
  each side is larger than the difference.
- **visual-qa** — six captures, all PASS, taken off the virtual desktop so
  the run never competed with the trader: aim flipped right at the left
  edge, mid-band, and against the axis at its minimum line length; the pill
  and the open form; and the aim absent from the pane that does not own
  paper input.
- **trader-ux-review** — no Blocker. Rafa loses a gesture and a mouse
  journey on entry and loses nothing on cancel; Marina keeps the
  Shift-to-level drawing drag; Duda cannot reach anything irreversible.
- **ui-harness** — every changed surface has a hook, added here:
  `QUANTICK_CMD_PREVIEW=<side>[@<0..1>]` parks the aim at a stated x,
  `QUANTICK_PAPER_ORDERS=<rungs>` rests orders on the first print (the
  scripted demo's own order is 220 prints away, 0.4 % out, and a lively
  tape filled it before the shutter three runs running), and
  `QUANTICK_PAPER_ORDER_HOVER=1` opens every tag with nobody at the mouse.

## arch-review

`code-review` at `high`, run twice — the second time over the final HEAD.
**15 findings, 14 confirmed and fixed here**, 3 deferred below. Four were
money-losing:

- moving onto the tape chip nulled the paint's pointer but not the press's,
  so the pill collapsed with no ✕ while a plain click on the chip cancelled
  the working order underneath;
- with the layer hidden the whole plot placed and journaled orders nothing
  painted;
- a modifier resting under the hand turned an armed ticket's click into a
  stop instead of the limit it armed, and left the ticket armed;
- sweeping toward the axis popped a ✕ open under the cursor, so the click
  cancelled an order while everything on screen said "place".

The root was one thing: the two deletions this branch made — `preview.hover`
from the press, the unconditional ✕ from the paint — each removed half of a
paint/press pair, and what replaced them were *shared formulas* rather than
shared values, handed different pointers and different rects.
`handle_chart_input` now decides once and both sides read the answer, the
pattern `cmd_preview` already had one field away.

### Deferred, deliberately

- **The open tag repeats the feed's full price precision** (`snap` rounds
  to `mark.scale()`, and a Binance mark has scale 8), so the *opened* form
  is wider than the tag this change shrinks. Predates this branch, does not
  bite an integer-quoted instrument like WIN, and fixing it changes what
  price an order is placed at — outside this mission's scope.
- **`QUANTICK_PAPER_ORDERS` does not survive a timeline reset**: a replay
  seek mid-capture sweeps the rungs and the hook is already disarmed.
  Development hook only; relaunching reaches the state.
- **The aim label covers the last stretch of its own minimum-length line**
  within 120 px of the axis. Cosmetic: the label sits on the aim's row and
  is the marker there, so the row is never unmarked.

### Out of scope

The position, stop and target tags keep the full form — there is at most
one of each, and none of them is the tag that filled the screen. Horizontal
parking and a density setting in the Trading tab were both considered with
the user, who chose hover-expand over them.

https://claude.ai/code/session_01AMpLr1FDLQnWkukq8Dq7EK


milocaetano:
**CI note.** The first run failed on `app::tests::a_hidden_layer_paints_nothing` (353 shapes against a baseline of 329 on the backfill divider) and passed on a re-run with nothing changed. It is a pre-existing flake, not a regression here: that test compares exact shape counts, and its own comment says the frame after a change does not count — the live lane's divider and the price range are computed by one draw and read by the next — but only the baseline was given that settling frame. The paths this branch touches are inert in that test (no orders, no position, no pointer in the input, no modifiers held). Hardened in `9d7ecef` by settling every measurement the way the baseline always was, rather than leaving a known flake for the next branch to trip on.

milocaetano:
**Angle findings — all eight triaged.** Six were already fixed in `9673d87`/`1653d56` (verified in the current tree, not from memory): the rung step's tick floor, the tape chip and pane chrome claiming their pixel, the `hover_pos`/`latest_pos` split behind the ✕, the order line's emphasis back on `line_at`'s own 10 px band, the forced aim never placing, and the dragged order's phantom cancel.

Two were not, and one of them is a **panic**, now fixed in `db1cd26`:

- **`clamp_tag_center` panics on a chart band under 20 px.** It ends in `y.clamp(top + 10, bottom - 10)`, and `f32::clamp` does not saturate when its bounds cross — it panics. The band is reachable: `plot_split` floors the plot at 20 px (`app.rs:305`), but `indicators::split_panes` then carves the indicator strips out of it with `(chart.bottom() - total).max(chart.top())` and no floor of its own (`indicators/mod.rs:651`). Every in-plot tag, every ✕ hit-test and the aim's layout go through it, so a trader with an order on screen and a tight layout loses the session. Pre-existing, but this branch gave it a second per-frame caller in `tag_row_hit`. Proven by reverting the guard — the new test panics inside `core::num::f32::clamp`.
- **`buy@nan` / `buy@inf`** parse as f32 and `clamp` passes NaN through. The finite filter was already in from `9673d87`; the test that covers it was not.

PR #194, opened 2026-08-17
feat: semi-automatic strategy anchors — arm a force-bar bot on a chart rectangle

milocaetano:
## Strategy anchors — the semi-automatic operation

The human draws and names the congestion region (a rectangle); the machine watches every closed bar and pulls the trigger in paper trading. The operational document demands exactly this split: intensity signals never fire alone on WIN — context is human, reaction time is the machine's.

**Scope note (gate of Onda 0):** the roadmap gates trading tools on the edge measurement (WP-02/03, not yet run). Decision taken with Camilo on this mission: this tool *is* the measuring instrument for the hybrid setup — the human region is not honestly mechanisable, so replay + paper with the bot pulling the trigger is how that setup gets measured. `--strategy force-region` measures the mechanical half over recorded sessions. Nothing here claims edge; everything here is paper.

### What landed

- **`quantick-strategy` (new domain crate)** — deterministic kernel: force-bar ruler (the `force_bar.pine` band: body between 1.5×–2.5× the average of 20 bodies; exhaustion does not fire), `Region`, projected brackets (TP/SL = close ± mult × trigger-bar range), and the armed-instance state machine (armed → fired → in position → done/disarmed, every disarm named). Trigger is a port (`Trigger`), proven by a fake in tests; the BEI docks later without surgery.
- **Drawing foundation** — `DrawingId` (stable identity where the Vec index is only position; survives delete/undo/reorder), optional user names ("congestão 108k"), and the per-drawing right-click menu (audit S7 + S4): name, rename, the strategy seat, lock/hide/delete under the lock's guard.
- **App runtime** — Add strategy… on a price-band rectangle opens the arming dialog (bank preset or custom; save/delete presets); armed instances evaluate on closed bars and fire `PlaceMarket` + bracket through the same funnel manual orders use (journal, toasts, ledger, report). On-chart badge with the state; safety sweeps disarm **by name** on timeline reset, bar-spec change, market switch, drawing delete, entry rejection, manual sweep. Flat gate: the bot never trades against any open position, manual included. One shot per arming by default; auto re-arm opt-in.
- **Strategy bank** — `quantick-strategies.toml`, versioned, env-overridable (`QUANTICK_STRATEGY_PRESETS`), exact-decimal strings, refuses whole what it cannot execute. Deliberately the surface a future natural-language layer writes into (preset row + `arm(preset, drawing, side)`).
- **Backtest** — the strategy port grows a default `on_events` hook (the same event stream the chart feeds its instances); `ForceRegion` docks the kernel behind `--region low:high:side` (one STRATEGIES row + one match arm). Harness test walks the *same tape* as the strategy crate's golden: one brain, two consumers.
- **Docs** — `docs/ux/strategy-anchors.md`; paper-trading §10's "no automated strategies" decision revoked by name; ui-harness hooks table extended.

### Performance impact (by path)

- **Per trade (hot):** `paper.on_trade` gains one bool branch; with a bot armed it clones simulator events only on the prints where something happened (fills/cancels — rare). `pane.ingest_live_trade` gains one emptiness check on the bar-close branch; the closed `Bar` is cloned only while instances exist. `tab.run_strategies` runs per ingested print: two early-out branches, one `mem::take` of an (almost always empty, non-allocating) Vec and one bool store when idle; real evaluation happens only on closed bars — O(instances) state matches plus one O(drawings) `index_of` per instance per closed bar.
- **Per frame:** the badge — one text galley + one rect per armed drawing (typically 1–3, zero when none). Same cost class as the existing off-band caret / chips.
- **Per interaction:** per-drawing hit-test on right-click press; menu/popup widgets.
- **Rare:** bank TOML load at boot, write on save. No new per-depth work, no locks, no HashMap where order reaches output (bank is BTreeMap).

Not measured against a `main` control run: no existing per-trade/per-frame code path was touched beyond the branches listed above, and every added cost is gated behind "instances exist". APP_HEALTH comparison deliberately deferred to the visual-qa session (blocked on app-launch authorization, below).

### Gates

- [x] Four checks green on top of `origin/main` (fmt / clippy -D warnings / build / test: 2064 tests, 0 failed)
- [x] arch-review over `git diff main...HEAD` — **step 0: bundled code-review at high, 15 confirmed findings (plus below-cap cleanups), all 15 fixed in `703063c`** (dropped-bracket protection, manual-flatten event routing, queued-order flat gate, composed-slot regions, per-session backtest state, cross-tab dialog, hidden-drawing pause with always-visible badge, unconditional trigger feed, re-arm ruler reset, bank validation, demo staging, menu rename commit, immediate instance removal on delete) + shape pass (magic values named); 2 findings deferred with reasons below
- [x] new-extension: kernel docks as a crate (3 contract files edited: root Cargo.toml, CLAUDE.md, workspace_deps whitelist); trigger port proven by fake; backtest docking is one row + one arm; defaults change nothing until a strategy is armed
- [x] Golden/determinism: kernel golden (same tape twice → identical run), engine untouched
- [x] ui-harness: `QUANTICK_STRATEGY_DEMO=1|popup`, `QUANTICK_STRATEGY_PRESETS` added in this change and documented in the skill
- [ ] visual-qa / trader-ux-review: **BLOCKED** — launching the app requires Camilo's explicit authorization (house rule); requested, will run on grant
- [x] PR opened; merge is not part of the mission

### Deferred, with reasons

- Direct tests for the `MarketChanged` / `BarSpecChanged` disarm sweeps: the mechanism (`disarm_all`) is unit-tested and the twin wiring (`TimelineReset`) has an integration test; exercising the other two paths requires spawning real feed threads. Three-line call sites, same shape.
- `quantick-strategies.toml` in the workspace bundle: the bank is cockpit-global like the drawing presets, but widening the bundle format is its own decision — follow-up.
- CLI flags for the force band / multipliers on `--strategy force-region` (fixed at the shipped defaults, auto re-arm): the measuring run that needs sweeps will add them; the docking point is in place.
- Non-rectangle shapes as strategy carriers; BEI trigger; real order routing (recorded as out of scope in the feature doc).


milocaetano:
**Live-fire report (WINV26 2026-08-17, paper):** armed on a real congestion rectangle, the bare relative band fired 5 operations in 12 minutes (11:02–11:14), all stopped — trigger bars like a 55-point body at 1.62×, invisible on the chart because bearish force painted red-on-red. Root cause: on volume bars every bar carries the same volume by construction, so the body average shrinks in congestion and the band alone is promiscuous — measured **247 of 1,355 bars** called force. Fixed in `d7d86a3`: `min_body` floor on the trigger (form default 100 pts → **7 elephants** in the same session, bodies 100–195, all in the volatile open) and visible force colours (white/yellow; biggest cedes to silver/gray). Verdict table reproduced from the exported tape with the kernel itself.

PR #216, opened 2026-08-20
fix(app): the book's marks leave with the book, and the tape stops eating the drag

milocaetano:
Three defects the trader found in one session driving the live WIN tape,
straight after #211 merged. None of them was reachable from a test that
existed, and two had been shipping for a while.

## The drag died over the tape band

#211 anchored the tape's live edge on prints, which gave a tape to every feed
that streams trades and no book — MetaTrader without market depth, every
recorded replay. It also gave those feeds a band across a third of the canvas
where pressing and pulling did nothing at all, in silence.

The wheel had already been handed back to the candles for exactly this reason;
the comment in `pane.rs` still says why — *"one gesture, two meanings, and
nothing on screen saying which one you were about to get"*. The drag never
caught up. `gesture_hits_lane` claimed the whole lane to protect a ten-pixel
resize handle, which is a dead zone rather than a guard.

Only the divider's own handle is off limits now (`gesture_hits_lane_divider`,
tested). The tape is pinned to the live edge and does not pan, so a drag across
it had no second meaning to lose.

## A reduction kind switched off kept painting

`draw_liquidity_events` walked the projection and drew every event **without
reading a switch**. The projection filters these by *threshold* and never by
the trader's choice — deliberately, because both kinds are factual and the
retained history stays complete — so the renderer was the only place left to
ask, and it was not asking.

Unticking "L2 reduction (unattributed)" dropped the legend entry and left every
violet mark on the canvas: the legend stating the layer was off while the
trader looked straight at it. The coverage gaps had the identical hole —
`show_gaps` reached the legend and nothing else.

## Switching the book off left the book's marks behind

Every one of those marks is a statement about the order book, so it goes when
the book does. Per pane, because the two maps switch apart: a first attempt
asked "is a book drawn anywhere", which left the candles violet whenever the
tape still had one — and that is the exact state the report came from.

## Test

`a_reduction_leaves_the_canvas_with_the_book_that_explains_it` builds four
reductions — one of each kind on each pane — and pins the ink that survives
each switch: the candles' book off clears only the candles, both maps off
leaves not one pixel, one kind off takes that kind and leaves the other.

Verified red without the change:

```
com a correção:  test result: ok. 1 passed
sem a correção:  test result: FAILED. 0 passed; 1 failed
```

`only_the_divider_handle_is_off_limits_to_the_pan` pins the drag boundary.

## Not in this PR

Two things the same session surfaced that are **not** regressions, checked
rather than assumed:

- **The book "disappearing" was the symbol.** `WIN$N` is the continuous
  contract and publishes no DOM; the bridge reported `images_sent=0`. On
  `WINV26` the same build reports `MT5_BOOK_AVAILABLE`, `book_status="live"`,
  10 levels a side.
- **The tape "lagging" is feed arrival.** `feed_arrival_ms = -622` — the MT5
  bridge delivers ticks ~0.6 s late, and the status bar says so. The renderer
  held `fps=59`, worst frame 21 ms, zero `APP_SLOW_FRAMES`. A long `auto` tape
  window (22 s under tick(2000) bars) adds to the impression.

And one real defect left open, with numbers: the **`dense tape` preset drops to
37 fps** at 2544 heat cells, with `heatmap_projection_ms` at 20 ms. Measured
against a control tree at `origin/main` under the same live book, both builds
hold 59 fps at ~400 cells, so it is the preset's price grouping and not this
branch. A shipped preset should hold 60 fps; that wants its own mission.

https://claude.ai/code/session_01R39RVJqyYDuPEkLbj8p8L3

### DIFF
--- a/crates/app/src/app.rs
+++ b/crates/app/src/app.rs
@@ -379,13 +379,17 @@ pub fn plot_split(
     }
 }
 
-/// Whether a pointer at `x` is over the live lane rather than the candles.
+/// Whether a pointer at `x` is on the lane divider's own resize handle.
 ///
-/// The divider itself counts as the lane, so the gesture that resizes it and
-/// the gesture that pans the candles can never both fire on the same pixel.
-/// Without a lane every pixel belongs to the candles, exactly as before.
-pub fn gesture_hits_lane(divider_x: Option<f32>, x: f32) -> bool {
-    divider_x.is_some_and(|divider| x >= divider)
+/// The one strip of the tape a chart gesture may not have: the resize drag and
+/// the pan must never both fire on the same pixel. Everything else in the band
+/// is the candles' — the tape is pinned to the live edge and does not pan, so a
+/// drag across it has no second meaning to protect, and reserving a third of
+/// the canvas to guard a ten-pixel handle is a dead zone rather than a guard.
+/// Without a lane there is no handle, and every pixel is the candles'.
+#[must_use]
+pub fn gesture_hits_lane_divider(divider_x: Option<f32>, x: f32, half_width: f32) -> bool {
+    divider_x.is_some_and(|divider| (x - divider).abs() <= half_width)
 }
 
 /// `X,Y` in screen points, for the harness hook that parks the properties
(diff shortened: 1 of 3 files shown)


milocaetano (commented):
Approved

PR #232, opened 2026-08-25
chore(control): land the observer stack on main

milocaetano:
## Why

PRs 3, 4, 5a and 5b each merged into the branch *below* them and never into `main`:

| PR | head | base it actually merged into |
| --- | --- | --- |
| [#221](https://github.com/milocaetano/quantick/pull/221) | `feat/control-gateway` | `feat/control-observer` |
| [#222](https://github.com/milocaetano/quantick/pull/222) | `feat/mcp-observer` | `feat/control-gateway` |
| [#223](https://github.com/milocaetano/quantick/pull/223) | `feat/control-events` | `feat/mcp-observer` |
| [#231](https://github.com/milocaetano/quantick/pull/231) | `feat/control-annotate` | `feat/control-events` |

So the trunk carried only PR 2 and the contract hardening. `crates/mcp`,
`crates/control-local`, `gateway.rs`, `journal.rs`, `events.rs`, `contract.rs`
and `emit_semantic_changes` did not exist on `main` at all — `emit_semantic_changes`
appeared there only as a string inside `roadmap.md`.

This was found while starting roadmap 5.1, which docks onto
`emit_semantic_changes` and cannot be built on a trunk that lacks it.

## What this does

Merges `origin/main` into the stack tip and resolves the three files where the
trunk moved underneath it. **Both sides were additive — no resolution drops work.**

- **`codec.rs`** — the stack added `HandshakeReply` (an accepted response *or* a
  redacted refusal) over the trunk's `HandshakeResponse`. `read_handshake_reply`
  supersedes `read_handshake_response`: a client that can only read an acceptance
  cannot tell a refusal from a truncated frame. The one trunk test that used the
  narrower door now reads the reply and unwraps it.
- **`schema.rs`** — the stack added `CompiledSchema` to pay the compile once; the
  trunk (#220) added `compile()` pinning patterns to the linear-time regex engine.
  `CompiledSchema::new` now builds *through* `compile()`, so the cache inherits the
  ReDoS guard, and `validate_instance` keeps the trunk's shape that does not
  re-meta-validate per instance.
- **the plan's status line** now says what is true.

Three stack-born schema documents were stale against the merged code — the trunk
bounded `request_timeout_ms` at 5000 and they had not seen it. Regenerated with the
documented commands; one added line each.

## Blast radius

| | |
| --- | --- |
| Files resolved by hand | 3 (`codec.rs`, `schema.rs`, the plan doc) |
| Lines of resolution | 5 conflict hunks |
| Test adapted | 1 (`codec_contract.rs`, to the superseding door) |
| Schemas regenerated | 3, one line each |
| Schemas arriving unmodified from main | 2 (byte-identical to `origin/main`) |

## Verification

- `cargo fmt --all -- --check` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo build --workspace` — clean
- `cargo test --workspace` — 2016 pass, 0 fail

One red in the local run, `the_bridge_paging_tests_pass`, is the known Windows
`python3` Store-alias environment failure, not a regression: both Python suites pass
when run directly — `bridge/mt5/tests/test_paging.py` (18 tests) and
`tools/mt5/test_export_session.py` (14 tests). No `.py` or `.mq5` file is touched by
this branch, so `ruff` has nothing to say about it.

## arch-review

Step 0 ran **inline rather than via the sub-agent** (this session was instructed not
to spawn agents), at high effort, scoped to the merge resolution rather than to
`main...HEAD` — the latter is the whole already-reviewed stack. 2 findings, both
confirmed:

1. **Fixed in this branch** (`46d3e54`) — the resolution dropped `HandshakeResponse`
   from the codec imports while writing a rustdoc link to it, giving a broken
   intra-doc link. clippy does not run rustdoc lints, so the four checks stayed green
   and it would have shipped. Now linked through the full path.
2. **Deferred, filed as Consider** — adding `"maximum": 5000` to `request_timeout_ms`
   *narrows* three v1 documents. The crate's own `compare_upper_bound` classifies
   `(None, Some(_))` as breaking, and `require_compatible_version` would reject it at
   the same version — but `schema_compatibility.rs` only tests hand-written fixtures,
   so committed v1 documents can narrow silently. Not introduced here: #220 already
   made this call on `main` for the two documents that live there, and these three are
   catching up to the same protocol. Naming it because roadmap 5.1 adds more v1
   schemas and the same gap will apply to them.

Performance: flat, cold path. Operability: no new surface. Docking: this merge adds
no port.

## Next

Roadmap 5.1 (`feat/control-snapshots`) builds on this. It is stacked on this branch
until this one lands on `main`.

https://claude.ai/code/session_014UrTTLKSAvkfAh535dwXHx


milocaetano:
Closing: superseded by #233, which is merged (`95cb0b7`).

Both PRs did the same job — land the control-plane stack on `main`. They were
opened 45 minutes apart by two sessions that could not see each other. #233 is
the one that landed, and it is the more complete of the two, so there is
nothing here to recover.

**This branch is now strictly behind `main`.** `git diff main...HEAD` over the
control area is 105 insertions against 191 deletions — merging it would *remove*
work that is already on `main`:

- the `script.rs` panic fix (a Pine span inside a multi-byte character took down
  the application thread, reachable by any client holding `annotate.script`)
- two authority fixes in `gateway.rs`: a refused trace-replay latching the
  recorded author onto the next action's object, and a revoked client's queued
  request still executing
- the elevated-Windows ACL fix in `control-local/discovery.rs`
- the MCP profile-ceiling fix in `mcp/server.rs`

It also predates the `profile_fold` / `frvp` engine work that is on `main`.

The one commit unique to this branch, `46d3e54` (repair the intra-doc link), is
not needed: it was only required because this branch resolved `codec.rs` by
dropping `HandshakeResponse` from the imports. #233 kept both typed handshake
readers, so the link resolves — `cargo doc --workspace` exits 0 with no warning
in `codec.rs` or `handshake`.

No action needed. The branch is left intact rather than deleted, so this is
reversible if anyone disagrees.

PR #239, opened 2026-08-26
fix(app): let the shipped layers reach the trader

milocaetano:
A trader reported that the four flow layers in the toolbar's top-right group —
bubbles, L2 heatmap, footprint, live strip — open **off** every launch, and that
the PR which shipped them on (#229) had changed nothing.

Both halves were true, and the first thing the reproduction settled is that the
defaults were never the problem. A launch with every cockpit store pointed at a
non-existent scratch path — a virgin machine — already opens Binance BTCUSDT in
the time+flow split with all four lamps lit:

```
APP_STARTING   feed=binance symbol=BTCUSDT
CANVAS_LAYOUT  layout=TimeAndFlow
CHART_LAYERS_RESTORED  hidden=1
```

So this is not a change of default. It is why the shipped answer never reached
anyone who already had a cockpit.

## The cause

`chart_layers::load` returned **only** the keys in the trader's file. A layer
absent from it fell through to the code baseline, which for all four flow layers
is *off*. Any `chart-layers.toml` written before #229 therefore answered "off"
to a question it had never been asked, on every launch. And it froze there:
`maintain_chart_layers` rewrites the whole map on the first switch of a session,
so one unrelated click turned that silence into an explicit `false` nobody
chose. A default that cannot reach anyone who already has a cockpit is not a
default.

Silence now resolves to the shipped answer. An explicit `false` still wins — the
promise that makes a shipped default acceptable — and a layer absent from *both*
files is still the app's to decide, which keeps the order-flow preset the only
home of `lane_marks`.

Three smaller ones travel with it, each a different way the same answer failed to
arrive:

- **The time pane never received the layer switches.** `apply_pending_layout`
  copied `hidden_layers` and stopped, leaving out every per-pane layer that does
  not live in that set — the footprint being the one that has one. This is the
  finding #229 deferred by name.
- **The heatmap lamp reported capture, not the switch.** `depth_visible()` is
  `enabled && show_depth`, so the button read unlit while book capture was
  starting; a trader pressing a dark button switched *off* the layer they were
  reaching for.
- **A new tab applied the startup map** rather than the live state its own
  comment promised — harmless only while that map was partial, and not once
  silence resolves.

And because this file is the trader's own answer and outranks the shipped
default from the next launch on, **every switch that moves is now named in the
log before it is written down** (`CHART_LAYER_SWITCHED`). The bug above cost a
day to chase because the only record was the file itself: a trio of `false`s
with no timestamp and nothing to say whether a hand had been near them.

## What is *not* fixed, and why that is the deliverable

A second, independent defect was found and could not be closed honestly.

On this machine a **maximised** window lays out larger than its own surface:
client `2560x1369 px`, egui told `2560x1369 pt` at `scale=1.5`, painting
`3840x2054 px` into it. The right and bottom third fall outside the window —
exactly where the four lamps, the price axis, the live strip and the dock rail
live. So the switches this PR turns on were not merely off for that trader; they
were off-screen.

Three corrections were built and measured, and each failed:

1. **Against `ViewportInfo::monitor_size`.** That field is in *points*
   (`egui-winit-0.29.1/src/lib.rs:970` builds it as
   `monitor.size().to_logical(pixels_per_point)`), so the check condemned every
   honest window on a 150% display: 1400x900 points evaluates `2100 > 1707` and
   `1350 > 961` and would have been shrunk to 933x600, then banked into the
   workspace file. **`arch-review` step 0 caught this as a Blocker** against an
   earlier commit on this branch; it is withdrawn.
2. **Against `inner_rect`.** Wrong together with `screen_rect`.
3. **Against the platform's own `GetClientRect`.** Returns **3840x2052** from
   inside the process where the same call from outside returns 2560x1369 —
   Windows virtualises coordinates into the caller's DPI context, so the
   "independent" reading agrees with egui to the pixel and never fires.

That is the finding: every observable available *inside* the process is
self-consistently wrong, because what is wrong is the process's own coordinate
space. A real fix has to key off something outside it — the GL framebuffer's
size, or a DPI comparison made in another awareness context — which is a change
to how the app declares DPI awareness, not a fix-up on the way into a frame.

Upstream has the same family open and unfixed: `emilk/egui#7648` (reproduces at
0.33.1, no PR) and `emilk/egui#7095`. The upgrade from the 0.29 we build against
was priced: **262 compile errors**, and `egui-phosphor` lags behind besides.

So what ships for it is the measurement, not a guess: `client_px` beside
`screen_pt` and `scale` in `APP_HEALTH_SUMMARY`, `QUANTICK_WINDOW_MAXIMIZED=1`
to reach the state from a cold launch, and `window_scale`'s module doc recording
all three dead ends. Those three numbers are what turned "the buttons are gone"
into a measurement. **Follow-up issue to come**; nothing here pretends it is
handled.

## Review

`step 0: code-review at xhigh, 11 findings, 11 confirmed` — all resolved, none
deferred. The first is the withdrawal above. The rest, resolved in `ae0be9d`:
a tautological perf-summary field; a doc comment stolen from `draw_toolbar` (the
same misplacement the branch had already fixed once for `maintain_workspace`); a
tab *switch* banking the other tab's layers as a choice and filling the new log
with switches no hand moved; seeding the time pane pushing an undo entry into an
empty drawing store; a `hidden_layers` clone overwritten one line later;
`CHART_LAYERS_RESTORED hidden=` no longer meaning what it said (renamed to `off`
with a `layers` total rather than quietly redefined); an unreachable branch that
returned before its own log on the one launch where an operator most needs it;
and a doc naming a caller this branch removed.

The shape pass found three behaviour changes with no test between them. All
three now have one, and the two that assert new behaviour were run against
`main`'s to confirm they fail without their fix:

```
the_time_pane_opens_on_the_same_layers_as_the_flow_pane        fails on main
a_new_tab_inherits_the_live_layers_not_the_startup_file        fails on main
  "the new tab reached past the session and brought back the startup file's answer"
a_file_from_before_the_defaults_still_opens_the_flow_layers    fails on main
  left: None  right: Some(true)   "heatmap is unanswered in this file"
```

Reaching the heatmap lamp needed it lifted out of the toolbar's frame into a
named reading — worth having anyway, and dimension 7's *read* half: the lamp is
now assertable without painting a toolbar.

## Verification

Four checks green. `cargo test --workspace`: **1660 passed, 0 failed** in the app
crate; `language_guard` passes and the prose, branch name and commit messages
were read. `the_bridge_paging_tests_pass` fails locally on Windows only — it
tries `python3`, which resolves to the Microsoft Store alias stub; the Python
tests themselves pass (18 checks) run directly, and CI's ubuntu image has a real
`python3`. Untouched by this branch.

**Performance.** Measured on Binance BTCUSDT at 1600x1000, eight health
summaries each:

| | fps | frame_avg | frame_cpu |
| --- | --- | --- | --- |
| `main` | 59-60 | 16.66-16.67 ms | 2.27-3.44 ms |
| this branch | 59-60 | 16.66-16.68 ms | 2.42-3.02 ms |

Every path touched is rare — startup, a layout change, a new tab, a layer switch
— and the one new per-frame cost from the withdrawn correction went with it.
`client_size_px` runs on the 2 s summary, not per frame.

**Visual QA.** Virgin launch and cold maximised launch captured and read: the
four lamps lit, both price axes, the live strip, the dock rail and the status bar
all present in the windowed case. The maximised case remains cut by the defect
above, which is stated rather than accepted as fixed.

https://claude.ai/code/session_017nQssV6ZTFWqKXCr91Lgm4

### DIFF
--- a/.claude/GOAL-archive-first-run-cockpit.md
+++ b/.claude/GOAL-archive-first-run-cockpit.md
@@ -1,0 +1,96 @@
+# Mission: the first launch opens the cockpit the chart is for
+
+**Objective:** on a machine with no saved cockpit, quantick's first launch opens
+on Binance **BTCUSDT** in the **timeframe + flow** split with all four flow-layer
+switches — bubbles, L2 heatmap, footprint, live strip — **on**; every later
+launch keeps them on unless the trader switched one off; and the window the
+trader actually works in shows them at all.
+
+Branch: `fix/first-run-cockpit` · Worktree: `../quantick-worktrees/fix-first-run-cockpit`
+
+## What the reproduction changed about this mission
+
+The three settings asked for were **already the shipped answer** — `feeds.toml`
+has `default_symbol = "BTCUSDT"` and `default_layout = "time+flow"`, and PR #229
+put all four flow layers in `config/chart-layers.toml`. A launch on `main` with
+every store pointed at a non-existent scratch path proved it: `feed=binance
+symbol=BTCUSDT`, `canvas_layout=TimeAndFlow`, `CHART_LAYERS_RESTORED hidden=1`,
+and all four lamps lit in the capture.
+
+So the mission was never to change a default. It was to find why the shipped
+answer does not reach the trader, and two independent causes were found.
+
+**Bug A — the trader's file replaced the shipped default instead of covering
+it.** `chart_layers::load` returned only the keys in the trader's file. A layer
+absent from it fell through to the code baseline, which for all four flow layers
+is *off*. Any cockpit written before PR #229 therefore answered "off" to a
+question it had never been asked, on every launch — and `maintain_chart_layers`
+rewrites the whole map on the first switch of a session, freezing that silence
+into an explicit `false`. That is precisely why PR #229 changed nothing for the
+trader who reported it.
+
+**Bug B — a maximised window lays out larger than its own surface.** Measured:
+client `2560x1369 px`, egui told `2560x1369 pt` at `scale=1.5`, painting
+`3840x2054 px` into it. The right and bottom third fall outside the window —
+exactly where the four lamps, the price axis, the live strip and the dock rail
+live.
+
+**Not fixed, and the reason is the deliverable.** Three corrections were built
+and measured. Against `monitor_size`: that field is in *points*
+(`egui-winit-0.29.1/src/lib.rs:970`), so the check would have shrunk every
+honest window on a 150% display to 933x600 — caught by `arch-review` as a
+Blocker. Against `inner_rect`: wrong together with `screen_rect`. Against the
+platform's own `GetClientRect`: returns 3840x2052 *from inside the process*
+where the same call from outside returns 2560x1369, because Windows virtualises
+coordinates into the caller's DPI context. Every observable inside the process
+is self-consistently wrong, because what is wrong is the process's coordinate
+space. Upstream has it open and unfixed (`emilk/egui#7648` at 0.33.1); the
+upgrade from 0.29 was priced at 262 compile errors for a release that still has
+it. What ships is the measurement — `client_px` beside `screen_pt` and `scale`
+in the health summary — plus `QUANTICK_WINDOW_MAXIMIZED=1` to reach the state,
+and `window_scale`'s module doc recording all three dead ends for whoever picks
+it up.
+
+## Acceptance criteria
+
+1. Reproduced on `main` before any code was written — capture plus log. **met**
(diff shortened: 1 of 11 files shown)


milocaetano:
Follow-up issue for the maximised-window defect is open: #240 — it carries the measurements, the three dead-end corrections and the two places a real fix could come from.

Also worth a separate look: `the_trade_paint_layer_switch_stops_the_marks` failed once on this PR's first CI run and passed on a re-run of the identical commit, so it is flaky on Linux rather than affected by this branch.

PR #243, opened 2026-08-27
feat(app): evidence bundles — one coherent, hashed, redacted capture of a session (5.4 / PR 5c)

milocaetano:
## Summary

The last item of the MVP (roadmap §5.4, plan PR 5c). Every other read answers
one question; an investigation needs the answers to have been true *at the same
instant*. `evidence.capture` freezes the named scopes, the events around them
and the effective configuration into one bundle, hashes it with Quantick
Canonical JSON, and answers with a manifest. `evidence.read` pages it back as
byte runs of that same canonical text through the `retained_resource` cursor
the contract already declared, so concatenating the chunks and hashing the
result reproduces the digest the manifest attests.

An optional screenshot is stamped with the *same* capture revision as the scene
taken beside it, which is the whole mechanism: `control_regions` then gives each
named control its rectangle in that image.

Three things it refuses to do. It never writes to disk — export is a cockpit
action. It never launders a scope: a capture requires `observe.evidence` plus
every permission the named scopes require plus the scopes every bundle always
carries, and every chunk read rechecks the bundle's own source scopes against
the current grant. And it never claims to be the whole session — `coverage`
names every scope omitted, every field a projection could not fill (derived by
walking for the availability shape, with the JSON Pointer that finds each one),
and every subject this tier does not carry at all, each with a coded reason
rather than a sentence.

Full record: `docs/control-plane/pr5c-evidence.md`.

## Rate class and tier

**On-demand captures.** Nothing runs unless a client asks.

| Path | Rate | What this change does there |
| --- | --- | --- |
| Aggregator, tick ingest | per trade | nothing |
| Book state, depth projection | per depth update | nothing |
| Renderer, per-frame view | ~60 Hz | one boolean, one `is_empty`, one `Option` clear — inside the gateway's own frame service, which runs only while local access is enabled. With nothing armed the input scan does not run and no pixels are touched |
| Evidence capture and chunk read | one per client call | the work of this change |

The application thread does only what needs application state: the projection
pass (already budgeted and guarded), a bounded journal read, a copy of the
effective configuration, and — when one was asked for — a clone of the frame
image's *handle*. Pixel conversion, PNG compression, canonical JSON, hashing,
chunking and retention all happen on the response worker that already
serializes a snapshot. `evidence.read` is a worker read and never enters the UI
queue at all.

`evidence_costs_the_frame_nothing_until_a_client_asks_for_it` runs thirty
frames with the tier granted and asserts nothing was ever armed.

**Tier:** two sensitive, off-by-default scopes — `observe.evidence` and
`observe.screenshot`, both already declared before this change.

## Docking / blast radius

| Added | Edited |
| --- | --- |
| `crates/app/src/control/evidence.rs` | `contract.rs` — two capability registrations, two prepare handlers, the `UiReadContext` port, `readable_scopes` |
| `docs/control-plane/pr5c-evidence.md` | `gateway.rs` — the store handle, the screenshot slot and its lifecycle, `invoke_local_read`, `forget_evidence` |
| `schemas/control/evidence-*.schema.json` (5) | `mod.rs`, `schema_catalog.rs`, `system.rs`, `config.rs` — registration and one shared helper |
| | `crates/control/` — three named bounds, `Base64Bytes`, one contract test |
| | `crates/mcp/` — `quantick_capture_evidence` and its guards |
| | `crates/app/Cargo.toml` — `png`, already in this crate's tree via `eframe -> image -> png` |
| | `.claude/skills/{ui-harness,visual-qa}`, `docs/control-plane/{roadmap,control-contract,README}.md`, `CLAUDE.md` |

No new crate. Nothing depends on `evidence.rs` but the contract that registers
it — a leaf, like every other module here.

## Acceptance (roadmap §5.4)

| Criterion | Test |
| --- | --- |
| 1. An agent explains the running session without a screenshot | `an_evidence_bundle_explains_the_session_without_an_image_and_its_events_keep_reading` — connects over the loopback socket, captures, pages back, verifies the digest, reads identity, build, host, capture revision and the workspace/feed/chart/health/scene projections out of it |
| 2. Feed, replay, indicator and connection changes appear through the cursor | same test feeds `events.next_cursor` straight back into `events.read`; `a_bundle_carries_the_events_around_the_capture_not_the_oldest_it_holds` pins the window itself |
| 3. The bundle reports omitted information and coverage gaps | `an_evidence_bundle_names_what_it_omitted_and_why_as_codes_not_prose`; `a_noisy_scope_cannot_spend_another_scopes_share_of_the_coverage_report` |
| 4. A screenshot bundle maps every named control to a region | `a_bundle_with_a_screenshot_maps_every_named_control_to_a_region_of_the_image`; `a_capture_that_wants_an_image_waits_for_the_frame_instead_of_answering_blind` |
| 5. A validation skill asserts through the live control plane | `ui-harness` gained *Reading the running app through the control plane*; `visual-qa` gained §3 *Ask the app what it believes, then look*, which takes a structured reading before the pixels and treats a scene/image disagreement as a FAIL. Fixture: `QUANTICK_CONTROL_EVIDENCE`, proved by `the_evidence_launch_hook_captures_through_the_same_read_a_client_calls` |
| 6. No token, user path, user text or config key in the bundle | `no_token_user_path_user_text_or_redacted_config_key_reaches_an_evidence_bundle` — plants the trader's note, a configured journal path, a bridge command naming their home, a routable bind address and the connection's real bearer token from the descriptor, then hunts all six through the reassembled bundle *and* the manifest |
| 7. Retention and size bounded by named constants | `retention_evicts_by_the_earlier_of_count_bytes_and_age`, `a_bundle_past_its_retention_is_gone_even_when_it_is_not_at_the_front`, `a_bundle_larger_than_its_own_share_is_refused_instead_of_emptying_the_store`, `a_bundle_too_large_for_one_chunk_still_pages_back_over_the_socket`, `withdrawing_access_forgets_every_retained_bundle` |

Also: `a_bundle_requires_the_scopes_it_always_carries_however_few_were_named`,
`evidence_capture_is_refused_without_its_own_scope_and_cannot_launder_another`,
`a_chunk_read_rechecks_the_grant_the_bundle_aggregated`,
`an_evidence_chunk_encodes_within_the_codecs_string_ceiling`,
`a_bundle_is_read_back_through_invoke_since_the_contract_names_no_tool_for_it`,
`the_reported_graphics_backend_is_the_one_the_manifest_selects`.

### Step 0: `code-review` at `max` — 15 findings, 15 confirmed, all fixed

None deferred. Six would have shipped broken, and each was invisible to the tests
that existed:

1. **No bundle over ~192 KiB could ever be read back.** The chunk size was
   picked against `CONTROL_MAX_RESPONSE_BYTES` when the bound that binds a
   chunk is `CONTROL_MAX_STRING_BYTES` — a chunk is one base64 *string*, and
   the codec prescans every string in every frame. Hidden because the
   screenshot fixture was a smooth ramp that deflated to a few kilobytes.
2. **The bundle laundered `observe.events` and `observe.market`** — it always
   carries an event page and the feed configuration but required neither, and
   because the manifest did not record them the read-time recheck was not
   looking for them either.
3. **The event page was the oldest the journal held**, not the newest.
4. **A picture could be stamped with a later capture's revision** — an
   orphaned frame from an abandoned capture was kept and handed on.
5. **One swallowed viewport command disabled screenshots for the session.**
6. **The give-up path answered `control.timeout`** where its own comment
   promised a bundle with a gap.

Two were inert-by-omission: `serialize_ui_result` discarded the `ControlError`
the widened signature exists to carry, and an oversized image failed the whole
capture instead of being dropped for the text. The rest were quality: the
coverage budget, four defects in the launch hook, two second copies (the
`host:port` splitter; `evidence.read` declared in the adapter with no test),
the double canonicalisation, the published scale not reproducing its own
regions, and the framebuffer copy on the application thread.

Shape dimensions: docks through the existing ports; the second operator has a
named call, a readable result and a registry entry for both halves; no
hardcoded configuration value; `language_guard` green and the prose, branch
name and commit messages read.

### Second pass over the opened pull request — 5 findings, 5 fixed (`5d0f872`)

All confirmed, none deferred. Two of them the first pass could not have found,
because the tests that should have caught them were shaped so they could not
fail:

1. **The event page carried the operator's own words.** The projections strip
   user text before it reaches a snapshot — but a bundle also embeds a page of
   the *journal*, and the journal records prose verbatim (`actions.rs:371`
   writes a mark's note; `notify.rs:398` a notification's message). So
   `coverage` asserted user text was redacted while the trader's typed note sat
   in the same document. The event page is now stripped on its way into the
   bundle; `events.read` still serves prose to a client holding
   `observe.events`, but a bundle must not become the one durable object where
   every scope's text is gathered together. The redaction test planted only a
   *drawing* note, which the projections already strip, so it passed for a
   reason unrelated to the path that leaked; it now plants a mark note and
   fails without the fix (`the trader's mark note reached the bundle`).
2. **A capture still encoding when access was withdrawn was retained anyway.**
   The ingredients are collected on the application thread and the bundle is
   built on a response worker milliseconds later, so `clear()` could run in
   between and the insert land after it — into a store emptied *because* the
   grant behind it was withdrawn. The store now carries an epoch that `clear()`
   bumps; a capture from an older epoch is refused with
   `control.resource_gone`.
3. **Coverage could claim a truncation that never happened** — the budget was
   checked at the top of the walk rather than where a field is added.
4. **The screenshot waiters ran outside the frame's admission guard**, so a
   frame could execute eight projection passes against a documented ceiling of
   four. They now count against the same budget the drain uses.
5. **The launch hook raised the screenshot notice before the scope check**, so
   asking for an image without `observe.screenshot` told the trader their
   window had been captured on the way to refusing it — weakening the one
   indicator `visual-qa` asserts on.

### Rebase onto `main` after #244, and a flake I introduced

Rebasing turned the suite intermittent: four of my own tests failed under the
parallel load of the whole workspace, and so did an unrelated pre-existing one
about a stalled reader. Twelve runs on `main` were clean and twelve on this
branch were not, so it was mine.

The cause was the waiting, not the capture. These tests spent a fixed eighty
frames waiting for a capture to park on an image, then published the pixels
whether it had parked or not — so under load the frame dropped an image nobody
had claimed (the staleness rule working correctly) and the capture sat parked
for its whole five-second timeout. With only eight global buffered-response
slots, a few tests doing that at once starved an unrelated client's
`control.describe`. The wait is now generous and asserted. Twelve workspace
runs clean on both sides afterwards.

**Worth its own change, not this one:** `CONTROL_MAX_IN_FLIGHT_PER_CONNECTION`
is 8 and `CONTROL_MAX_BUFFERED_RESPONSE_SLOTS` is also 8, so one connection can
hold every global slot. That is what let slow tests starve a fast one, and it
predates this branch — `gateway_a_client_that_never_reads_does_not_stall_another`
asserts a headroom the constants do not guarantee. Left alone here; it belongs
in a change made with the threat model's starvation rule (O-11) in hand.

## Deferred

Nothing from either review pass. Two deliberate deviations, both recorded in
the evidence document:

- **No named MCP tool for the chunk read.** Contract §8 names
  `quantick_capture_evidence` and no companion, and says the long tail uses
  `quantick_invoke`. `evidence.read` goes through `invoke`, under test.
- **`read_only: true` on a capability that retains something.** In the effect
  policy's sense: a bundle changes no application state, touches no position
  and takes nothing from the trader. The threat model already places evidence
  capture in the observer's *Allowed* list for the same reason.

**`visual-qa` / `trader-ux-review` were not run: BLOCKED.** No authorization to
launch the desktop app in this session. The screenshot notice, the surface
threat-model O-18 requires, is asserted headlessly instead — pixels enter the
control plane through one function and it raises the notice every time, which
is what lets the test assert it at all.
`QUANTICK_CONTROL_EVIDENCE=all,screenshot` paired with
`QUANTICK_CONTROL_PANEL=1` reaches it for a capture run.

## Verification

```
cargo fmt --all -- --check                                   exit 0
cargo clippy --workspace --all-targets -- -D warnings        exit 0
cargo build --workspace                                      exit 0
cargo test --workspace                                       36 binaries ok
```

`quantick-app` 1735 passed / 0 failed / 6 ignored; `quantick-control` and
`quantick-mcp` green; `language_guard` 4/4.

`quantick-feed-mt5`'s `the_bridge_paging_tests_pass` fails on this machine for
an environmental reason unrelated to this branch — the `python3` Windows Store
alias intercepts the interpreter. Running the script directly:
`python bridge/mt5/tests/test_paging.py` -> *all checks passed across 21
tests*, exit 0. Unchanged from `main`; CI runs it under a real `python3`.

Schemas and the capability catalog regenerated and committed; the snapshot
tests and the no-egui-on-the-wire guard are green over 46 documents. The MCP
tool list is fixed at ten for the observer profile, guarded in `tools.rs`,
`server.rs` and the STDIO smoke test.

## What this closes

Plan §18's last open criterion, *an evidence bundle can reproduce an
investigation*. With 5.1, 5.2 and 5.3 on `main`, the MVP is complete when this
merges.

https://claude.ai/code/session_01LEwtrMfe2cLDidGkD7PHyL


crates/app/src/control/evidence.rs
```diff
@@ -0,0 +1,1892 @@
...
+            .iter()
+            .map(|chunk| raw_sha256(chunk))
+            .collect::<Vec<_>>();
+        let chunk_count = chunks.len();
+
+        let retention_ms = self.store.retention_ms();
+        let expires_at_unix_ms = self
+            .captured_at_unix_ms
+            .saturating_add(i64::try_from(retention_ms).unwrap_or(i64::MAX));
+        self.store.insert(
```
milocaetano, line 1150: **Medium — a bundle can be retained *after* access was withdrawn.**

`forget_evidence()` (gateway.rs) calls `EvidenceStore::clear()` on the application thread when the trader disables local agent access, but `EvidenceCapture` carries its own clone of the store handle and this `insert` has no guard against a store that has since been emptied.

Concrete: the trader clicks *Disable access* while an 8 MiB screenshot bundle is being PNG-encoded and canonicalised on the response worker. `clear()` runs first on the UI thread; this `insert()` runs after, on the worker. The bundle — screenshot included — then sits in memory for the full `CONTROL_EVIDENCE_RETENTION_MS` (15 minutes), and `request_enable()` never clears the store, so it survives a re-enable too.

The module doc ("gone the moment access is withdrawn") and the contract both promise otherwise, and `withdrawing_access_forgets_every_retained_bundle` only exercises `clear()` in isolation, so nothing catches it. A generation/epoch counter on the store bumped by `clear()` and captured into `EvidenceCapture` would make the late insert a no-op.

crates/app/src/control/evidence.rs
```diff
@@ -0,0 +1,1892 @@
...
+fn walk_unavailable(
+    prefix: &str,
+    root: &Value,
+    budget: usize,
+    found: &mut Vec<EvidenceUnavailableField>,
+    truncated: &mut bool,
+) {
+    let mut pending = vec![(prefix.to_owned(), root)];
+    while let Some((pointer, value)) = pending.pop() {
+        if found.len() >= budget {
```
milocaetano: **Low — `truncated` can be set when nothing was actually dropped.**

The budget check runs at the top of every iteration, for every pending node: scalars, arrays, and objects with no `available: false` marker alike. So a scope whose walk finds exactly `budget` unavailable fields and still has any ordinary node left on the stack sets `*truncated = true` and returns.

`into_manifest` then appends `{subject: "coverage.unavailable_fields", reason: "truncated_at_named_limit"}` to `not_captured` even though the list is complete — a false statement inside the one section whose job is to be honest about what is missing. Moving the check to the point where a field is about to be pushed (or only tripping it when `pending` still holds an object) fixes it.

crates/app/src/control/evidence.rs
```diff
@@ -0,0 +1,1892 @@
...
+/// Each entry is a decision recorded where a reader will meet it, rather than
+/// a silence they would have to interpret.
+fn fixed_gaps() -> Vec<EvidenceGap> {
+    vec![
+        EvidenceGap {
+            subject: "diagnostic_logs".to_owned(),
+            reason: "not_captured_in_this_tier".to_owned(),
+        },
+        EvidenceGap {
+            subject: "user_authored_text".to_owned(),
```
milocaetano: **Low — `user_authored_text: redacted_by_projection_policy` is asserted unconditionally, but the embedded event page is not redacted.**

Every bundle always carries `events: EventPage`, and journal payloads carry free text verbatim: `create_mark` (crates/app/src/control/actions.rs) writes `"note": input.note` into the `attention.mark.created` payload, and the annotate module writes `{"annotation": payload}`, which for a label includes its `text`.

Concrete: take a mark with a note (`QUANTICK_CONTROL_MARK=<note>`, or any annotator-profile client), then capture a bundle while that event is still in the journal's newest page — the reassembled document contains the note, while `coverage.not_captured` states the bundle carries no user-authored text.

`no_token_user_path_user_text_or_redacted_config_key_reaches_an_evidence_bundle` plants a *drawing* note only, which the projection layer already strips, so the journal path is untested. Either scope the gap's wording to the projections, or redact free-text payloads on the way into the bundle.

crates/app/src/control/gateway.rs
```diff
@@ -1043,20 +1241,181 @@ impl ControlAccess {
...
+        self.screenshot_armed = false;
+        self.screenshot = Some(raw);
+        app.show_agent_toast(SCREENSHOT_NOTICE.to_owned());
+    }
+
+    /// Run the captures that were waiting for an image, or give up on them.
+    ///
+    /// Returns before touching anything when none is waiting, which is every
+    /// frame of every session where no client asked for a picture.
+    fn serve_awaiting_screenshot(
```
milocaetano, line 1356: **Low — deferred captures run outside the frame's admission guard.**

`serve_awaiting_screenshot` runs before `drain_bounded_since` and executes up to `CONTROL_MAX_SCREENSHOT_WAITERS` (= `CONTROL_UI_MAX_REQUESTS_PER_FRAME`, 4) full `execute_on_ui` passes without consulting either the elapsed-time budget (`CONTROL_UI_BUDGET_US`) or the per-frame request count. The drain then admits its own four on top.

Concrete: four clients call `evidence.capture` with `screenshot: true` in the same frame. On the next frame the application thread runs eight full projection passes (each up to 32 scopes) instead of the documented ceiling of four, and because the work never goes through the drain, no budget violation is recorded for it either. Counting these against the same frame budget the drain uses would keep the ceiling one number rather than two.

crates/app/src/app.rs
```diff
@@ -2582,6 +2604,118 @@ impl QuantickApp {
...
+                    .readable_scopes()
+                    .into_iter()
+                    .map(|scope| scope.to_string()),
+            );
+        }
+        if wants_screenshot {
+            // Harvests as well as arms: the frame service that normally takes
+            // the pixels runs only while the gateway is enabled, and this hook
+            // is meant to work without a client on the socket.
+            access.service_screenshot(self, ctx);
```
milocaetano, line 2678: **Low — the screenshot notice fires before anything checks that `observe.screenshot` is granted.**

`service_screenshot` sends the viewport command and, once the frame arrives, `accept_screenshot` raises `SCREENSHOT_NOTICE` through `show_agent_toast`. Only afterwards does `invoke_local_read` reach `contract.prepare`, which refuses the capture with `control.scope_denied` when the screenshot scope is missing from `configured_scopes`.

Concrete: launch with `QUANTICK_CONTROL_EVIDENCE=all,screenshot` and `QUANTICK_CONTROL_SCOPES=all-reads,observe.evidence`. The trader is told "Your assistant captured a picture of this window", the hook logs `CONTROL_EVIDENCE_HOOK_REFUSED`, and no picture ever left the process. It fails safe, but `visual-qa` now treats that toast as the evidence the O-18 indicator works, so a notice that can appear with no capture behind it weakens exactly the assertion it was added for. Checking the configured grant for `observe.screenshot` before arming would keep the notice one-to-one with a real capture.

milocaetano (commented):
Correctness pass over the diff — 5 findings, none of them blockers for the feature's happy path. The highest-value one is the retention race at `EvidenceStore::insert`.

milocaetano:
## Second review pass — all five findings fixed (`5d0f872`)

Every one confirmed, none deferred. CI green on the new head.

**1. The event page carried the operator's own words.** The one that matters.
The projections strip user text before it reaches a snapshot, and the bundle
said so — but a bundle also embeds a page of the *journal*, and the journal
records prose verbatim: the note a trader types at a mark (`actions.rs:371`),
the message an assistant sends (`notify.rs:398`), the words on a label. So
`coverage` claimed user text was redacted while the trader's typed note sat in
the same document.

The event page is now stripped on its way into the bundle, unconditionally —
`events.read` still serves prose to a client holding `observe.events`, but a
bundle must not become the one durable object where every scope's text is
gathered together. The marker says redacted rather than empty, so a reader can
tell a withheld note from a mark that never had one.

The original canary could not find this because it planted a *drawing* note,
which the projections already strip. The test now plants a mark note too, and
I verified it fails without the fix:

```
thread '...no_token_user_path_user_text_or_redacted_config_key_reaches_an_evidence_bundle'
panicked at crates\app\src\app.rs:30770:17:
the trader's mark note reached the bundle
```

**2. A capture still encoding when access was withdrawn was retained anyway.**
Real race, as described. The store now carries an epoch that `clear()` bumps;
each capture records the epoch it was collected under and an insert from an
older one is refused with `control.resource_gone`. Test:
`a_capture_that_finishes_after_access_is_withdrawn_is_not_retained`.

**3. Coverage could claim a truncation that never happened.** The budget was
checked at the top of the walk rather than where a field is added, so a scope
that fitted exactly but still had a scalar queued reported dropped fields.
Test: `a_coverage_walk_that_fits_exactly_does_not_claim_it_truncated`.

**4. The screenshot waiters ran outside the frame budget.** They now count
against `CONTROL_UI_MAX_REQUESTS_PER_FRAME` and check `CONTROL_UI_BUDGET_US`
before each pass; a waiter the frame cannot serve stays queued for the next
one. Both halves now read the elapsed time through one shared helper, so the
waiters and the drain are measuring the same thing.

**5. The launch hook raised the notice before the scope check.** `wants_screenshot`
is now cleared against `grants_screenshot()` before anything arms a rasterise,
and the refusal is logged. The indicator only means something if it is never
wrong.

Thanks — findings 1 and 2 were both things the existing tests were structurally
unable to catch.

crates/app/src/control/evidence.rs
milocaetano: Fixed in `5d0f872`. `StoreState` now carries an `epoch` that `clear()` bumps; `EvidenceCapture` records the epoch it was collected under and `insert` refuses an older one with `control.resource_gone`. Test: `a_capture_that_finishes_after_access_is_withdrawn_is_not_retained` — it clears between collection and insert, which is exactly the race described.

crates/app/src/control/evidence.rs
milocaetano: Fixed in `5d0f872`. The budget is now spent where a field is about to be pushed, not at the top of the walk, so a scope that fits exactly no longer claims it truncated. Test: `a_coverage_walk_that_fits_exactly_does_not_claim_it_truncated`.

crates/app/src/control/evidence.rs
milocaetano: Fixed in `5d0f872`, and this was the most valuable of the five — the leak was real, not just the wording. Took the second option: the event page is redacted on its way into the bundle (`note`, `message`, `text`), unconditionally. `events.read` still serves prose under `observe.events`; what a bundle must not become is the one durable object where every scope's text is gathered together. The redaction test now plants a mark note as well as a drawing note, and I verified it fails without the fix: `the trader's mark note reached the bundle`.

crates/app/src/control/gateway.rs
milocaetano: Fixed in `5d0f872`. `serve_awaiting_screenshot` now counts against `CONTROL_UI_MAX_REQUESTS_PER_FRAME` and checks `CONTROL_UI_BUDGET_US` before each pass; a waiter the frame cannot serve stays queued for the next one. Both halves read the elapsed time through one `elapsed_us_since` helper, so the ceiling is one number rather than two.

crates/app/src/app.rs
milocaetano: Fixed in `5d0f872`. `wants_screenshot` is cleared against a new `ControlAccess::grants_screenshot()` before anything arms a rasterise, and the refusal is logged as `CONTROL_EVIDENCE_HOOK_SCREENSHOT_NOT_GRANTED`. The notice is now one-to-one with a real capture, which is the property `visual-qa` leans on.

crates/mcp/src/tools.rs
```diff
@@ -192,6 +203,14 @@ pub fn tools(profile_ceiling: &str) -> Vec<Tool> {
             output_schema: Some(capability_output_schema(parse_schema(SNAPSHOT_CAPTURE_SCHEMA))),
             annotations: ToolAnnotations::observer_read("Scene"),
         },
+        Tool {
+            name: CAPTURE_EVIDENCE.to_owned(),
```
milocaetano, line 207: **P1 ? Request evidence scopes in the MCP handshake**`n`nWhen `quantick-mcp` is the client, `serve` builds its handshake request solely from `OBSERVER_SCOPES` (`crates/mcp/src/main.rs:43-64`), which includes neither `observe.evidence` nor `observe.screenshot`. The gateway intersects requested scopes with the trader's grant, so even after both are enabled this advertised tool always returns `control.scope_denied`, and `evidence.read` through `quantick_invoke` is unusable too. Request both scopes here, as the adapter surface requires at `CLAUDE.md:18`.

crates/app/src/control/gateway.rs
```diff
@@ -1043,20 +1253,198 @@ impl ControlAccess {
         };
 
         self.emit_semantic_changes(app);
+        // Before the drain, and deliberately: an image that arrived this frame
+        // is of the frame just painted, and the projections about to be taken
+        // still describe that frame. Harvesting after the drain would pair a
+        // scene with a picture of the screen before it.
+        self.harvest_screenshot(app, ctx);
+        self.serve_awaiting_screenshot(app, generation, ctx, frame_started);
```
milocaetano: **P1 ? Capture state from the frame that produced the screenshot**`n`nWhen a live feed or replay advances between frames, the `Event::Screenshot` found here is the reply produced after painting the frame in which the command was sent, but `draw_frame` has already called `drain_tabs()` and replay/annotation hooks before reaching `begin_frame` (`app.rs:9811-9842`). `execute_on_ui` therefore captures newer projections and merely stamps the older pixels with their new revision, so the image and bundle can describe different state. Retain the photographed frame's DTO or capture before those mutations, per `CLAUDE.md:40`.

crates/app/src/control/evidence.rs
```diff
@@ -0,0 +1,2103 @@
...
+/// trader has set up. A bundle that carried them while requiring neither would
+/// be exactly the aggregation-as-a-way-in this module says it refuses — and
+/// the read-time recheck could never catch it, because a scope missing from
+/// the manifest is a scope the recheck is not looking for.
+const ALWAYS_AGGREGATED_PERMISSION_IDS: &[&str] = &[
+    // The event page, and its cursor.
+    "observe.events",
+    // The effective configuration: feed IDs, symbol catalogues, the
+    // MetaTrader port.
+    "observe.market",
```
milocaetano, line 1674: **P2 ? Require `observe.system` for every evidence bundle**`n`nFor a capture naming only `feed.status`, a client can hold `observe`, `observe.market`, `observe.events`, and `observe.evidence` while lacking `observe.system`; this list still authorizes the call, although every manifest and document embeds `system_snapshot()` at `evidence.rs:1022-1024`. Add `observe.system` to the always-aggregated set so capture and every chunk read enforce that scope, as required by `docs/control-plane/observer-threat-model.md:192-196`.

crates/app/src/control/gateway.rs
```diff
@@ -1043,20 +1253,198 @@ impl ControlAccess {
...
+    /// Whether this request has to wait a frame for the window to be
+    /// rasterised, arming the rasterise if so.
+    fn defer_for_screenshot(&mut self, request: &UiRequest, ctx: &eframe::egui::Context) -> bool {
+        if !request.prepared.dispatch.needs_screenshot() || self.screenshot.is_some() {
+            return false;
+        }
+        if self.awaiting_screenshot.len() >= CONTROL_MAX_SCREENSHOT_WAITERS {
+            return false;
+        }
+        self.arm_screenshot(ctx);
```
milocaetano, line 1438: **P2 ? Recheck authority before arming a screenshot**`n`nIf a screenshot request is queued or parked when the user revokes that connection, this method runs before `execute_on_ui` performs its generation/revocation check (`gateway.rs:1164-1171`), so it keeps issuing `ViewportCommand::Screenshot` and may show the capture notice after revocation even though the request is ultimately denied. Reject revoked or stale requests before this deferral and remove their waiters, because revocation must apply immediately to in-flight work (`docs/control-plane/observer-threat-model.md:105-110`).

crates/app/src/control/gateway.rs
```diff
@@ -1043,20 +1253,198 @@ impl ControlAccess {
...
+    ///
+    /// One function, called by both teardown paths, so the two cannot disagree
+    /// about what a withdrawal clears — which is exactly the kind of omission
+    /// nobody notices until a screenshot flag outlives the door it came
+    /// through.
+    fn forget_evidence(&mut self) {
+        self.evidence.clear();
+        self.screenshot = None;
+        self.screenshot_armed = false;
+        self.awaiting_screenshot.clear();
```
milocaetano, line 1467: **P2 ? Clear evidence when the gateway stops unexpectedly**`n`nIf the gateway exits through `LifecycleEvent::Failed` or an unexpected `Stopped`, `poll_lifecycle` changes the state to `Disabled` but never calls this helper; frame service then stops, leaving retained bundles, armed pixels, and screenshot waiters behind while access is off. Call this cleanup on every terminal lifecycle transition, not only explicit disable and exit, as required by `docs/control-plane/observer-threat-model.md:94-103`.

crates/app/src/control/gateway.rs
```diff
@@ -1043,20 +1253,198 @@ impl ControlAccess {
...
+        let mut waiting = std::mem::take(&mut self.awaiting_screenshot);
+        while let Some(request) = waiting.pop_front() {
+            // These captures spend the same frame budget the drain does, and
+            // they run before it. Left uncounted, four waiters plus the
+            // drain's own four would put eight projection passes in one frame
+            // against a documented ceiling of four, and the drain would find
+            // its budget already gone. A waiter this frame cannot serve stays
+            // queued and is served by the next one.
+            if served >= CONTROL_UI_MAX_REQUESTS_PER_FRAME
+                || elapsed_us_since(frame_started) > CONTROL_UI_BUDGET_US
```
milocaetano, line 1393: **P2 ? Share the waiter count with the ordinary frame drain**`n`nOn a frame where one ready screenshot waiter executes, `served` becomes 1, but the subsequent `drain_bounded_since` initializes its own `processed` counter to zero and can execute four more requests; with four give-up waiters the total can still reach eight. This local check therefore does not enforce the documented four-request frame ceiling. Pass the consumed count or remaining quota into the ordinary drain.

crates/app/src/control/evidence.rs
```diff
@@ -0,0 +1,2103 @@
...
+    .map(redact_event_prose)
+}
+
+/// Payload keys under which an operator's own words reach the journal.
+///
+/// `note` is what a human types at a mark, `message` what an assistant puts in
+/// a notification, `text` what either of them writes on the chart. A new
+/// action that records prose adds its key here — and the doc comment is the
+/// contract, because nothing else can find prose in a free-form payload.
+const USER_PROSE_PAYLOAD_KEYS: &[&str] = &["note", "message", "text"];
```
milocaetano: **P2 ? Redact the event fields that actually contain object names**`n`nWhen an annotation create call receives `name`, the journal serializes it as `annotation.label` (`annotate.rs:438-459`), and `script.attach` similarly records `script.name`; neither key is in this list, so `recent_events` embeds the operator's prose while coverage says events were redacted. Redact the actual journal payload shapes, including string arrays such as declared inputs, in keeping with `CLAUDE.md:40`.

crates/app/src/control/evidence.rs
```diff
@@ -0,0 +1,2103 @@
...
+        EvidenceGap {
+            subject: "diagnostic_logs".to_owned(),
+            reason: "not_captured_in_this_tier".to_owned(),
+        },
+        EvidenceGap {
+            subject: "user_authored_text".to_owned(),
+            // Two mechanisms, one promise: the projections strip it before it
+            // reaches a snapshot, and the event page is stripped on its way
+            // into the bundle.
+            reason: "redacted_from_projections_and_events".to_owned(),
```
milocaetano: **P2 ? Redact indicator prose before claiming it is absent**`n`nWhen callers include `analysis.indicators`, that projection deliberately emits user-authored script kinds, titles, and plot/input names under the `observe.user_text` grant (`analysis.rs:11-18` and `analysis.rs:484-490`). The evidence document copies the projection verbatim, so this unconditional `redacted_from_projections_and_events` entry is false. Either strip those fields for bundles or report the retained text accurately, per `CLAUDE.md:40`.

crates/app/src/app.rs
```diff
@@ -2582,6 +2604,131 @@ impl QuantickApp {
...
+                self.control_access = Some(access);
+                return;
+            }
+            tracing::warn!(
+                target: "quantick::control",
+                event_code = "CONTROL_EVIDENCE_HOOK_GAVE_UP_ON_IMAGE",
+                frames = CONTROL_EVIDENCE_HOOK_FRAMES,
+                "the window never delivered a frame to rasterise; capturing without one"
+            );
+            wants_screenshot = false;
```
milocaetano: **P2 ? Preserve the screenshot request after hook timeout**`n`nWhen `QUANTICK_CONTROL_EVIDENCE=...,screenshot` exhausts its 120-frame wait, changing this flag to false sends `screenshot:false` to `capture_prevalidated`, which records `screenshot/not_requested`; the promised `frame_not_delivered` gap is produced only when the request remains true and no frame is present. Preserve the request, and cancel any outstanding arm separately, so the logged manifest honestly reports the timeout as required by `CLAUDE.md:40`.

crates/app/src/control/evidence.rs
```diff
@@ -0,0 +1,2103 @@
...
+    }
+}
+
+impl StoreState {
+    fn expire(&mut self, now_unix_ms: i64) {
+        while let Some(front) = self.bundles.front() {
+            if front.expires_at_unix_ms > now_unix_ms {
+                break;
+            }
+            self.drop_front();
```
milocaetano: **P2 ? Expire stale bundles beyond the queue front**`n`nExpiration order is not guaranteed to match this deque: concurrent response workers can insert bundles out of capture order, and wall-clock rollback can also reorder deadlines. An expired bundle behind a live front is never removed here, so a later count or byte eviction drops the live front while the stale bundle consumes its slot. Enforce a monotonic ordered deadline or scan all entries before applying bounds.

.claude/skills/ui-harness/SKILL.md
```diff
@@ -282,6 +283,73 @@ Landing with the one-week candle default (`fix/frvp-candles-window-and-history`)
...
+is a complete client, no extra tooling. It discovers the running instance
+itself and never starts one:
+
+```powershell
+$lines = @(
+  '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{},"clientInfo":{"name":"ui-harness","version":"1"}}}',
+  '{"jsonrpc":"2.0","method":"notifications/initialized"}',
+  '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"quantick_get_scene","arguments":{}}}'
+) -join "`n"
+$lines | & "$target\debug\quantick-mcp.exe" --profile observer
```
milocaetano: **P2 ? Define and build the MCP executable used by the skill**`n`nThis PowerShell block never defines `$target`, and the preceding workflow builds only `quantick-app`, not `quantick-mcp`; on a clean target directory the command therefore resolves to a missing `debug/quantick-mcp.exe` or runs an unrelated stale binary. Define the target directory and build `-p quantick-mcp` before invoking it so the documented validation path is reproducible.

crates/mcp/src/tools.rs
milocaetano: Fixed in `eb794f2`. Both scopes added to `OBSERVER_SCOPES`. This was the worst of the eleven and entirely mine: the comments directly above that list explain this exact failure mode for `observe.paper` and `observe.user_text`, and I added the tool without adding the ask — so `quantick_capture_evidence` would have answered `control.scope_denied` forever, even after the trader ticked both boxes. Rather than just add two strings, the list is now held against the committed capability catalogue by `the_adapter_asks_for_every_scope_an_observer_capability_needs`, which names the capability and the scope it needs. I verified it fails with the scopes removed: `observe.evidence (needed by "evidence.capture")`.

crates/app/src/control/gateway.rs
milocaetano: Fixed in `eb794f2` — reported rather than papered over. You are right that `drain_tabs()` runs before `begin_frame`, so the pixels are of the painted frame and the projections come after the next drain. Moving the capture ahead of the drain would reorder the replay-trace re-injection and the annotate hooks that deliberately run first, which is not a change I would make at this point in the branch. So the bundle now records `screenshot.state_skew / pixels_precede_projections_by_one_drain` whenever it carries an image, and `EvidenceScreenshot::capture_revision` documents exactly what the shared revision ties together: controls do not move in a drain, so every region still names what is under it, but a price elsewhere can be one step newer. That is the honesty rule you cited applied to a skew I cannot currently remove.

crates/app/src/control/evidence.rs
milocaetano: Fixed in `eb794f2`. `observe.system` joins `ALWAYS_AGGREGATED_PERMISSION_IDS`, so both the capture and every chunk read enforce it. Test: `a_bundle_requires_the_system_scope_it_always_embeds`.

crates/app/src/control/gateway.rs
milocaetano: Fixed in `eb794f2`. `serve_awaiting_screenshot` now checks the grant generation and the revoked set before anything else and answers the waiter with `control.permission_denied`, so a revoked connection cannot keep the window rasterising or raise the capture notice.

crates/app/src/control/gateway.rs
milocaetano: Fixed in `eb794f2`. `forget_evidence()` is now called on the `Failed` arm and on the `Stopped` arm as well as on explicit disable and exit — four call sites, all terminal transitions.

crates/app/src/control/gateway.rs
milocaetano: Fixed in `eb794f2`, and you are right that my previous fix was only half of it. `serve_awaiting_screenshot` now returns how many requests it served and `drain_bounded_since` takes that as `already_processed`, so the two share one counter and one ceiling instead of two.

crates/app/src/control/evidence.rs
milocaetano: Fixed in `eb794f2`. `label`, `name` and `title` added to `USER_PROSE_PAYLOAD_KEYS`, which is how an annotation's own name and a script's were reaching the page. The list's doc comment now says explicitly that it is the contract and that a new action recording prose must add its key.

crates/app/src/control/evidence.rs
milocaetano: Fixed in `eb794f2` — took the "report accurately" branch. Stripping those fields inside evidence would duplicate the projection's own policy, and `analysis.indicators` publishes them on purpose so an agent can address an indicator by name. So the single blanket claim is now two derived ones: `user_authored_text_in_events / redacted_by_payload_key` always, and `user_authored_text_in_projections`, whose reason is `retained_under_observe_user_text` or `redacted_by_projection_policy` depending on the scopes the bundle actually aggregated.

crates/app/src/app.rs
milocaetano: Fixed in `eb794f2`. The request stays `true` on give-up, so the capture finds no image and records `frame_not_delivered` — what actually happened — instead of `not_requested`, which was a lie about a run that asked for a picture and waited two seconds for it.

crates/app/src/control/evidence.rs
milocaetano: Fixed in `eb794f2`. `expire` now sweeps every entry and returns the swept bytes to `total_bytes`, instead of stopping at the first live bundle. Test: `an_expired_bundle_behind_a_live_one_is_swept_and_gives_its_bytes_back`, which inserts in an order the clock disagrees with and checks the live neighbours on both sides survive.

.claude/skills/ui-harness/SKILL.md
milocaetano: Fixed in `eb794f2`. The block now defines `$target`, runs `cargo build -p quantick-mcp`, and invokes the resolved path, so the documented validation path works on a clean target directory instead of finding a missing or stale binary.

PR #253, opened 2026-08-29
feat(app): a load-older press that reaches the previous session, and venue candles on a chart cut by trades

milocaetano:
## Summary

A chart could not show yesterday, and three separate things were stopping it.

**The press fetched a page, not a span.** `Tab::request_older_history` sent one
`load_older` of `history_step` prints and stopped; on a liquid contract that is
minutes. Every transport serving trade history takes a count and a cursor — a
venue answers nothing else — so reaching a *span* has to be a run of requests,
one at a time, with the reply sending the next. `crates/app/src/history_reach.rs`
owns that: a `HistoryReach` the trader picks (`page`, the press every release
before this one had and still the default, or `previous-session`) and a
`Campaign` that spends pages until the reach is met or a budget stops it. It
holds no channel and no clock — it is told what the chart holds and answers what
to do next — so every stop condition is a unit test rather than a session with a
live venue.

Where the session's open comes from had to be decided rather than looked up: not
a calendar, because quantick has no venue session table and inventing one is a
second source of truth about every exchange's hours, wrong the first time a
holiday moves. The tape says it. A stretch with no prints at all wider than the
configured gap is the market having been closed, the print on its older side is
that session's last, and `previous-session` continues a further configured lead
past it so the day before is on screen to compare against rather than merely
touched.

**A chart cut by trades got no prefix at all.** `refold_history_prefix` handed a
pane with no `time_interval_ms()` an empty vector, so a tick, volume, dollar or
imbalance chart never carried a venue candle and `NoChartCutByTime` greyed the
*older candles* control out on it. That was honest while the only offer was a
fold, and a fold is exactly what cannot happen here — a minute is not a tick bar.
Installing the base *unfolded* is a different offer and a true one: the venue's
own minutes, in front of bars cut by trades, already counted apart from built
bars on the status bar. Behind a switch that is **off by default**, because a
candle appearing in front of a trader's prints unasked would be this change
taking a decision that is theirs.

**A replay refused to ask.** `request_ohlcv_history` returned early while
`replay.is_some()`, yet `feed::replay` answers `FetchOhlcv` from the context file
downloaded beside the recording — the run-up that file exists to carry. A
recording opened at a wall and only picked its context up if the trader happened
to press *load older*. The guard is gone; the capability already reports `false`
for a recording with no context, so one without context is still never asked.

Closes #

## Verification loop

All four exit 0 on the current head, which merges `origin/main` at `3ba1711`.
CI passed on `75439b4` (the head before that merge).
`crates/app/tests/language_guard.rs` passes. `bridge/mt5/` and `tools/mt5/` are
untouched, so `ruff` and the exporter test are not in scope for this change.

## Evidence

Driven end to end against the real MetaTrader socket with a scripted bridge
speaking `bridge/mt5/PROTOCOL.md` schema 1 (hello declaring `history_paging`,
today's session as the opening backfill, `load_older` answered by walking back
into the previous day). **One press of `+ older`:**

```
HISTORY_REACH_SETTLED tab=0 symbol=WINJ26 pages=11
  anchor_ms=1774008002500 reached_ms=1773944503000 action="reach_met"
```

Eleven pages from one press. The anchor is today's oldest print; the run crossed
the overnight gap on its second page (the bridge's cursor jumped 14.9 hours) and
stopped 3.05 hours before the previous session's close — the configured lead. The
chart's time axis then reads `18:06 … 18:23` (yesterday) followed by `09:04 …`
(today), and the status bar goes from ~96 bars to `535+0 bars`.

The lead-in, on a replay with a context file beside it:

| | status bar | candle request |
| --- | --- | --- |
| off (today's behaviour) | `0+14 bars` | none made |
| on | `240v+0+14 bars` | `OHLCV_RECEIVED bars=240 action="install_prefix"` |

The default really is unchanged: the control run issued zero `FetchOhlcv`.

## Performance, declared by rate

Nothing per-trade, per-depth or per-frame is touched.

- `Campaign::advance` and `last_close_before` — **rare**, once per history reply.
  `last_close_before` binary-searches the anchor and walks *back* from it,
  stopping at the first break, so the usual cost is the page that just arrived.
- `refold_history_prefix` / `trim_to_seam` — **rare** (a candle slice, a
  prepended page, a switch). `trim_to_seam` keeps its owning signature so the
  common time-pane path stays a `retain` over a vector `fold` just built and is
  about to drop; the borrowed variant beside it serves the lead-in, and one
  `seam_bucket_ms` owns the rule both read.
- One bar rebuild per page, which is the pre-existing cost of a press now paid
  per page. Measured over the eleven-page run: `frame_cpu_ms` flat at ~11 ms,
  `frame_worst_ms` 64 ms on the first prepend then 42–48 ms. (Captured under
  llvmpipe software rendering, so the ~37 ms `frame_avg_ms` is the renderer, not
  quantick.) Bounded by `MAX_CAMPAIGN_PRINTS`, and the loading overlay is up
  throughout.

## Blast radius (`new-extension`)

One file added — `crates/app/src/history_reach.rs` — and twelve edited, every
edit registration-shaped: a field, a mirror line, a menu block, a snapshot field,
a config section, a `mod` line. `HistoryReach::ALL` is the port: adding a reach
is one variant plus its arms in that file, and `runs_a_campaign()` is what `tab.rs`
asks so no `== PreviousSession` grows there. Defaults preserve today's behaviour
exactly — `HistoryReach::Page` and `venue_lead_in: false` — and the control run
above proves it rather than asserting it.

## Drivable without a mouse

- **Act** — `Tab::request_older_history(config)`, `QuantickApp::set_history_reach`,
  `Tab::set_venue_lead_in`. The click and the hook take the same door.
- **Read** — the workspace snapshot carries `history_reach`, `history_reach_running`
  and `venue_lead_in`; each pane already carried `venue_prefix_present`. All three
  are `#[serde(default)]`, since a required field is a breaking change to a v1
  schema.
- **Discover** — `HistoryReach::ALL` backs the menu and `QUANTICK_HISTORY_REACH`
  alike; an unknown token is refused out loud rather than silently defaulted.

New harness hooks, registered in `.claude/skills/ui-harness/SKILL.md` in this
change: `QUANTICK_HISTORY_REACH=<token>` and `QUANTICK_VENUE_LEAD_IN=1`.

## Reviews

`arch-review` ran over `git diff main...HEAD`. **step 0: `code-review` at high,
14 findings, 13 confirmed on re-argument** — 13 fixed, one resolved by
disclosure, one outside this session's reach. The two it would have blocked on:

- **A run that never stopped.** Only MetaTrader ever lowers `history_paging`;
  Binance's is a compile-time `true` answering an empty block to a rate-limited
  fetch exactly as to "nothing older", so one press became 64 back-to-back REST
  calls — how a 429 becomes an IP ban. Now bounded by `MAX_IDLE_PAGES` (three,
  not one: an empty page is *ordinary* while a bridge crosses a weekend, and
  both halves are tests).
- **The span cap fired before the reach across every weekend.** Monday's open is
  ~62 h after Friday's close and the cap was 48 h, so the run stopped at the gap
  on exactly the mornings the reach is named for. The cap now applies only while
  the tape has shown no close at all.

`trader-ux-review` found two Should-fixes, both fixed: the `+ older` button
stayed lit during a run and swallowed the press, and nothing said that
withdrawing the reach is how a run is called off.

## Notes for the reviewer

**One test moved that this feature does not touch**, and it is worth reading why.
`gateway_wait_for_change_sees_a_human_mark_and_does_not_delay_a_concurrent_read`
turned CI red on `8c42742`. I had called it a pre-existing environmental flake in
an earlier draft of this description and expected it to pass on a clean runner.
It did not, so I root-caused it rather than re-running, and it is a real defect
in the test. It means to assert that a *parked wait blocks nothing*; it also
asserted, silently, that **one frame** suffices to drain the gateway's queue —
scheduling, not contract. A long frame left the reply unwritten, so `read()`
blocked until the request's own five-second deadline and returned
`control.timeout`: a slow frame rendering as a wrong answer. It now runs frames
until the answer is on the socket, bounded, which is the loop the parked waiter
twenty lines below already uses. Measured: **3/15 failures on this branch and
7/15 on `origin/main`** in the same container before the fix, **0/25 under
deliberate CPU load** after. No test was skipped, disabled or quarantined.

Three things deferred deliberately, none of them silent:

1. **The branch name is Portuguese and carries none of `CLAUDE.md`'s three
   prefixes.** It was assigned by the session harness, which forbids pushing to
   any other branch, so it could not be fixed from here. Worth renaming on merge
   — the merge commit carries it into history otherwise.
2. **With the lead-in on, an indicator running across the seam averages venue
   minute candles with bars cut by trades.** On a time pane the prefix is folded
   to that pane's interval so the series stays one population; here it cannot be.
   Resolved by disclosure — the View hover and the field doc both say it —
   rather than by hiding the prefix from a trade-cut chart's indicators, which
   would make one prefix mean two different things depending on the pane.
3. **No registered control-plane *action* for the reach.** The state is
   readable and the named setter exists; a `workspace.history_reach.set`
   capability with its descriptor and schemas is a separate change.

Two boundary cases worth eyes on: `SESSION_GAP_MS` / `PREVIOUS_SESSION_LEAD_MS`
now default a `[history]` config section (documented in `feeds.toml`), and I
sized them against B3's index future — a broker whose instrument has a real
lunch break will read that pause as a close and want a longer gap. And
`MAX_CAMPAIGN_PRINTS` at 250 000 is my judgement of "one press's worth of
re-cutting"; a run that stops there is not a failure, since the anchor moves and
the next press continues, but it is the number most likely to want tuning after
a real session.

https://claude.ai/code/session_017g9WeMbATa7SkEoL8jR2qK


milocaetano:
CI went red on `8c42742` with one failure, in a test this diff does not touch:

```
---- app::tests::gateway_wait_for_change_sees_a_human_mark_and_does_not_delay_a_concurrent_read ----
assertion failed: matches!(snapshot.outcome, ResponseOutcome::Success { .. })
test result: FAILED. 1886 passed; 1 failed
```

I flagged this in the PR body as a pre-existing flake and expected it to pass on a clean runner. It did not, so I root-caused it instead of re-running — and it turned out to be a real defect in the test rather than luck.

**What was wrong.** The test means to assert that a *parked wait blocks nothing*: a read arriving on another connection is served while one is registered, and the assertion above it already proves the parked wait holds no UI request slot. But it also asserted, silently, that **one frame is enough** to drain the gateway's queue. That is scheduling, not contract. When the frame ran long the reply had not been written yet, so `read()` blocked until the request's own five-second deadline and came back `control.timeout` — a slow frame rendering as a wrong answer.

**The fix** (`75439b4`) runs frames until the answer is on the socket, bounded — the same loop the parked waiter twenty lines below already uses. Nothing about what the test proves changes; it stops proving something it never meant to.

**Measured, because "flake" is not a root cause.** The failure reproduced **3/15 on this branch and 7/15 on `origin/main`** in the same container, so it predates this branch. After the fix: **0/25 with four CPU burners running** — the load the old version needed to fail.

No test was skipped, disabled or quarantined, and no empty commit was pushed. The four checks pass locally on the new head (`fmt`/`clippy -D warnings`/`build`/`test --workspace` all exit 0).

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #256, opened 2026-08-29
fix(app): the load-older press says what it reached

milocaetano:
A `+ older` press must either put older history on the chart the trader selected, or say on screen — once, quietly, without a dismiss — that there is none to fetch. The *previous session* reach must never end in silence.

## What was wrong

Reported as "the previous-session loading is a facade, shipped without testing". It was, in two separate ways.

**Nothing ever told the trader what a press reached.** `CampaignEnd` names seven endings and six of them mean *the chart did not move*. Every one settled into a `tracing::info!` line and nothing else, so on screen a run that gave up after three empty pages was indistinguishable from a button that does nothing. The default one-page reach was equally silent.

**A market replay claimed it could page trades.** `feed/replay.rs` declared

```rust
// A recording pages *candles*, never trades: the tape in the file
// is the whole day, and there is no venue behind it to ask for more prints.
history_paging: request.session.context.is_some(),
```

— a fact about a *candle* file answering the *trade* capability, directly under a comment saying the opposite. So a recording put the reach chips, the page-size box and a live `+ older` button in front of the trader, answered every request with `HistoryPrepended(vec![])`, spent its three idle pages and stopped. That is exactly the reported session: a replay of WINV26, `0 trades backfilled so far`, nothing on screen.

## What changed

- **`CampaignEnd::notice`** gives every ending its sentence and `ReachMet` none — yesterday on the chart is a better answer than a line about it. The sentences carry the one distinction a trader acts on (press again, or the record is spent) and name no budget's size.
- **The sentence lands in the loading lane**, in the row the spinner just left: the trader watched `loading history…` appear when they pressed, and the answer to that press has no business arriving anywhere else. Muted, no spinner, no dismiss; it leaves after `HISTORY_NOTE_LINGER` on its own.
- **`Tab::empty_page_verdict`** reads the same two facts `Campaign::advance` reads first, in the same order, so a page with no run behind it can never tell a different story about one empty block: a withdrawn capability is `Exhausted`, an empty chart is `NothingCharted`, and only what is left is this request's own.
- **A recording declares what it can serve.** `history_paging: false`, so the trade half of the history control is not offered over one. The "pick up a run-up downloaded since this session opened" behaviour moved onto `FetchOhlcv` — the request that actually answers with candles — gated on the file's modified time, because it now runs on the playback thread whose next step charges elapsed time to the playhead.
- **`HistoryPagingOff`** replaces the pair of bools the refusal grew into, and reads the candle entry's own `OlderCandles` state, so the refusal never ends at a control that is itself greyed out. On the order-flow default — a tick chart, no time pane — it names the venue lead-in, which is the switch that actually unblocks the trader.
- **`QUANTICK_HISTORY_NOTE=<ending>`** photographs the surface through the same call a settled run makes, and the control plane carries the verdict beside `history_trade_count`, so an operator without a mouse learns the outcome in the same words.

## Evidence

**Screenshots** (all at 60 fps; blank or clipped captures would be the environment, and these are neither):

| Frame | Shows |
| --- | --- |
| `note-nothing-coming-back.png` | A replay with the note up — *"nothing older came back — the venue has run out, or is refusing for now; give it a moment before pressing again"* — and `+ older` honestly greyed |
| `note-control-none.png` | The same launch without the hook: no note, `+ older` still greyed |
| `note-live.png` | The note over a live Binance chart |

**Four checks**: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo build --workspace` and `cargo test --workspace` all clean — 1936 app tests, 36 test binaries green. The single failure is `the_bridge_paging_tests_pass`, environmental on this machine (`python3` resolves to the Windows Store alias); `python bridge/mt5/tests/test_paging.py` passes all 21 checks and this branch does not touch the bridge.

**Performance**: the settle is *rare* (once per history reply); `expire_history_note` is *per-frame* over a `Copy` `Option` and one duration compare, skipped entirely while there is no note; the lane allocates one galley per frame while a note is visible, as its sibling rows already do. `reload_context` went from every candle request to only a file that changed. No per-trade or per-depth work added.

**Tests written first** — the missing test is what let this ship, so they were written and run red before the fix. Among them: `a_run_that_reaches_nothing_says_so_where_the_trader_is_looking`, `a_run_that_meets_its_reach_says_nothing`, `the_settled_reach_paints_its_sentence_over_the_chart` (reads the frame's own shapes, so the sentence is proved onto the glass and not just into a field), `a_recording_never_claims_to_page_trades`, `a_refusal_never_points_at_a_control_that_is_also_off`.

## Review

**Step 0** — `code-review` at `xhigh`, **14 findings**. One was already fixed in flight (the agent read a mid-edit tree); one was refuted (`CampaignEnd::ALL` was said to want `LoadingTask`'s guard, which only exists because `LoadingTask::index()` is `self as usize` — `CampaignEnd` already follows `HistoryReach::ALL`'s established pattern in the same file, and a stronger exhaustive-match guard was added anyway). The other **twelve are fixed in `9c685a5`**, including one promoted to Blocker:

> `history_reach_running` landed in the observer feed-status `required` list while `SCHEMA_VERSION` stayed 1, which §5.6 of the control-plane plan calls breaking — a v1 payload recorded by an earlier build would stop deserializing. `#[serde(default)]` keeps it optional, so the addition is additive as intended.

**`trader-ux-review`** — one Should-fix, fixed: `NothingComingBack` told the trader to "press again to retry", inviting exactly the request burst `MAX_IDLE_PAGES` exists to refuse (that budget is sized against a venue answering a rate-limited request with the same empty block a closed market gives). It now asks them to give it a moment, and the budget's own doc says why the sentence reads that way.

**Nothing deferred.** No Blocker or Should-fix is left open.

## Not in scope

A recording opened with **no** context file has no candle capability for the rest of the session — `fixed_capabilities` never updates — so a run-up downloaded mid-replay cannot be picked up there. The old `LoadOlder` path could not reach that case either (`history_paging` was the same expression), so this is not a regression, and the test says plainly which half is reachable. Making that capability live is its own change.


milocaetano:
Rebased onto `main` after #258 (`feat/replay-day-before`) and #259 landed. Three conflicts, all in the same territory:

- **`loading.rs`** — `main` scoped the overlay per surface (`overlay_scoped(.., scope)`). The note now rides along and `overlay_scoped` is the single place that decides which surface carries it, derived as `scope == LoadingTask::History.scope()` rather than pinned to `Whole` by name: move `History` to another surface and its verdict follows it, instead of the answer to a press landing on a surface the press was never about.
- **`app.rs`** — took `main`'s four scoped draws wholesale; each is handed the note, and the filter inside decides. Deciding at the call sites as well would be the same coupling written twice.
- **`tab.rs`** — took `main`'s restructured source-switch resets; kept the clause about the verdict going with the run.

**No semantic conflict with #258**, which is the one that could have had one. "The day before" joins the session through the *opening backfill*, once, never through a request — so a recording still cannot page an older trade and `history_paging: false` stays the honest answer. The two are complementary: #258 puts yesterday on the chart, this stops the button claiming it can fetch more. Pinned as a third case in `a_recording_never_claims_to_page_trades` so the interaction cannot regress unnoticed.

Four checks re-run green at `8a7c088` — 36 test binaries, CI passing.

PR #264, opened 2026-08-30
feat(app): carve the Surface port, and a ratchet that keeps the trunk shut

milocaetano:
Stop the `app` crate's trunk from absorbing every new capability: add a mechanical size ratchet and an `arch-review` accumulation dimension, then carve the `Surface` port and move exemplary surfaces onto it.

**step 0: code-review at xhigh, 14 findings — 11 fixed, 1 fixed in part, 2 deferred (listed at the bottom).** One was a Blocker in this branch's own guard; see *What the review caught*.

## Why

`arch-review` dimension 1 asks whether a capability can dock as a new file plus one registration line. Recent features answered yes *honestly* — `pointer_compass.rs` really is a new file — and `app.rs` still went from 108 lines to over 36,000 in five weeks, monotonically, never once shrinking.

The review measures the leaf and never the trunk. Nothing asked **where the registration lines accumulate**. They accumulate in `QuantickApp` (133 fields) and `new_with_workspace` (1,149 lines), because a capability with no port is docked by hand in four places — a field, an initialiser, a draw call, a hotkey — and every one lands in the same file. `grep` for "line count | file size | god object | monolith" in `arch-review/SKILL.md` returned zero hits.

Measured coupling says the design underneath is sound, not rotten: **9 of 21 `draw_*` surfaces touch 1–2 fields** of `QuantickApp`; only `draw_frame` (14) and `draw_menu_bar` (17) are genuinely tangled. This is wiring debt, not architecture debt — which is why it is mechanical to repay.

## What is in it, in rising order of risk

**1. `crates/app/tests/size_guard.rs`** — a ratchet over *production* lines, meaning every line outside a top-level `#[cfg(test)]` item. Two thirds of `app.rs` is test code, and dimension 4 asks for exactly that: a guard counting total lines would fire on a well-tested change and teach the author to write fewer tests. It fails on growth past a recorded ceiling, and also when a file sits more than 200 lines *below* one, because unclaimed slack is only headroom for the next feature to refill silently. Raising a ceiling stays legal — it just has to be a signed line in the diff instead of an invisible +400 inside a 36,000-line file.

**2. `arch-review` dimension 9 and the `new-extension` rule** — the judgement half of that guard. Both name the shape they exist to stop: `ChartLayer` is a registry in name only, with 21 variants across 264 `ChartLayer::` sites in six files, so adding a layer reopens `app.rs`, `pane.rs`, `tab.rs` and `toolbar.rs`. `new-extension`'s blast radius now counts lines, not only files — one new file beside 2,000 lines poured into thirteen others reads healthy on a file count alone.

**3. The `Surface` port** — `SurfaceEnv` and `SurfaceResponse`, copied from `dock.rs`'s `DockEnv`/`DockResponse` rather than invented, so the repo keeps one way to do this. The response is a **struct, not an enum**: a new request is an added field defaulting to "did not ask", never a `match` arm that reopens every caller. The trait also carries `apply_env_hook`, so a surface that forgets its `ui-harness` hook shows up as an empty override instead of compiling silently. The agent popup, the acknowledgement toast and the Save-as box move onto it with their tests and their hooks — and `app.rs` now drives none of the three.

The registry is a typed struct rather than `Vec<Box<dyn Surface>>`, deliberately: the host must reach a specific surface to command it (`show_agent_popup`), and the routes back from a trait object are downcasting — which dimension 1 calls a broken port — or a command enum, which is the growth this port exists to stop. The trade is documented at the top of `surfaces/mod.rs`.

## What the review caught

Worth stating plainly, because it is the most useful thing in this branch: **step 0 found a Blocker in the size guard itself.** The first version stopped counting at the first `#[cfg(test)]` line of any kind, not at the test module. Fifteen files carry a top-level `#[cfg(test)] use` / `const` / `fn` above their tests, so the guard scored `control/gateway.rs` at **72 lines of its 4,142**, `drawings/mod.rs` at 221 of 2,283 and `toolrail.rs` at 96 of 2,114 — blind on the largest files in the repo, and one `#[cfg(test)] use` at the top of `app.rs` would have switched it off there too.

Fixed by counting every line outside a top-level `#[cfg(test)]` *item*, pinned by four cases including both spellings that caused it. The tracked set went from 13 files to 18, and `paper_trading.rs`'s ceiling from a false 7,611 to its real 8,861. Both skills were asserting a coverage the counter did not have; those sentences now say what it counts and how it was wrong.

## Evidence

| Criterion | Result |
| --- | --- |
| Ratchet fails on growth | **Verified both ways.** +5 lines in the production region → `FAILED … 11879 production lines, ceiling 11874 (+5)`. +5 lines inside `mod tests` → passes, by design. |
| Ratchet fails on unclaimed slack | Fired on its own author: `down to 11650 from 11874 — good news, tighten the entry to 11650`. |
| Ratchet is not blind | 18 files tracked, including the five the first version could not see. |
| `arch-review` dimension 9 | Added; severity list, `ReportFindings` categories and the closing verdict extended. Numbered 9, not renumbered, because `mission` and eight `GOAL-archive-*` files cite dimension 8 by number. |
| `new-extension` UI-surface rule | Added, with the `Surface` row in the port table. |
| Port + second implementation | `a_second_implementation_needs_only_the_trait` drives an outside type through the same fold `draw_all` uses and asserts its request survives; `a_shipped_surface_and_a_fake_take_the_same_path` holds the fake to the contract the product is held to. |
| 3 surfaces moved | `QuantickApp` **133 → 130 fields**; `app.rs` **11,874 → 11,651** production lines; `draw_frame` hand-calls none of them, and neither does `new_with_workspace`. |
| Behaviour preserved | `cargo test --workspace`: **2,497 passed** across 38 binaries. |
| Visual | All three surfaces captured twice — before and after the review fixes — at fps 56–59 and read: popup (title, message, attribution, Dismiss), toast (message + Undo, `Foreground` layer, unobstructed), Save-as (hint text, focus ring, Save correctly disabled on an empty name). **PASS.** |
| Per-frame performance | Flat. See below. |
| Blast radius | **5 new files, 1,253 lines; existing files +168 / −285 — net −117 lines of existing code.** A package, not a patch. |
| Four checks | fmt 0, clippy 0, build 0, test — one known-environmental failure, below. |

### Performance

Per-frame path, so it is measured, not asserted. `frame_avg_ms` is vsync-bound at ~16.7 ms on a 60 Hz desktop, so `frame_cpu_ms` is the number that moves. Three passes, with the order reversed in the second to separate the change from drift in market density and desktop load; the third is the code as shipped, after the review fixes.

| Pass | control (`origin/main`) | branch |
| --- | --- | --- |
| 1 — control first | 2.050 ms | 2.239 ms |
| 2 — branch first | 2.071 ms | 1.827 ms |
| 3 — final code | 2.066 ms | 2.021 ms |
| **mean** | **2.062 ms** | **2.029 ms** |

The ordering flips between passes and the branch's own runs span 0.41 ms — ten times the 0.03 ms difference in the means. The effect is below the noise floor, which is what the code predicts: production dispatch is **static**, `dyn Surface` appearing only inside `#[cfg(test)]`, so three direct calls replaced three direct calls. 19 health samples per run, dense frame (`QUANTICK_BOOK_AUTOSTART` + `QUANTICK_BUBBLES_AUTOSTART` + `QUANTICK_LIVE_STRIP_AUTOSTART`).

### Known-environmental test failure

`quantick-feed-mt5 :: the_bridge_paging_tests_pass` fails on this machine only: it tries `python3` first, which here is the Microsoft Store alias stub. Running the script the test wraps passes outright — `python bridge/mt5/tests/test_paging.py` → **all checks passed across 31 tests**. Unrelated to this branch (nothing here touches `feed-mt5` or the Python bridge) and green on CI's real `python3`.

## Deferred, deliberately

- **Toasts raised after `draw_all` paint one frame late.** `draw_all` sits where `draw_agent_popup` was, early in `draw_frame`, so an acknowledgement raised by the object manager or by `save_named_workspace` — both later in the frame — appears on frame N+1 rather than N. The call site was chosen to preserve the popup's `Middle`-order layering exactly; moving it late instead would put the popup above every dialog it currently sits below. A 16 ms deferral of a toast against a visible z-order change is the better trade, and no test detects it — but it is a real difference, recorded here rather than in a comment nobody reads.
- **A workspace-wide policy lives in `quantick-app`'s integration tests.** A `pine`-only branch that shrinks `compile.rs` past the slack has to edit an app-crate file to land. CI runs `cargo test --workspace`, so nothing escapes, but the altitude is wrong; the right home is an `xtask` or a tiny crate of its own. Kept here because moving it is its own change and this branch already carries three.
- **`paper_trading.rs` still has its own `Toast`** — per tab, on a 4-second clock, with its own 96 px lift. Pre-existing, so not a leftover of this extraction, and its doc claim of being "the window's one acknowledgement channel" has been corrected rather than left false. It is the obvious next implementation of the port, and converging it would be the strongest proof the port is real.
- **`Surfaces` is not enumerable.** `Surface::id()` gives each surface a stable id, and each paints under `egui::Id::new(self.id())` so the name cannot drift from the layer — but there is no iterator for the control plane to list what is on screen. Adding one now would ship an API with no caller (clippy's `dead_code` already caught that once here). It belongs with the first consumer that needs it.
- **The indicator legend paints over the agent popup.** Visible in the capture. **Not a regression** — `draw_agent_popup` sat at `draw_frame` line 83 and `draw_indicator_legends` at 91 on `origin/main`; on this branch they are 83 and 106. Same relative order, so the overlap predates this work.

### DIFF
--- a/.claude/GOAL-archive-surface-port-and-size-ratchet.md
+++ b/.claude/GOAL-archive-surface-port-and-size-ratchet.md
@@ -1,0 +1,64 @@
+# Mission
+
+Stop the `app` crate's trunk from absorbing every new capability: add a
+mechanical size ratchet and an `arch-review` accumulation dimension, then
+carve the `Surface` port and move exemplary surfaces onto it.
+
+## Why (the diagnosis this mission answers)
+
+`arch-review` dimension 1 asks whether a new capability can dock as a new file
+plus one registration line. Every recent feature answered yes honestly — and
+`app.rs` still grew from 108 lines to 34,030 in 37 days, monotonically. The
+review measures the leaf and never the trunk: nobody asks *where the
+registration lines accumulate*. They accumulate in `QuantickApp` (130 fields)
+and `new_with_workspace` (1,022 lines), because there is no port for "being a
+surface of the app" — so each of the 68 modules is wired by hand in four
+places: a field, an init, a draw call, a hotkey.
+
+Measured coupling says the underlying design is sound, not rotten: 9 of 21
+`draw_*` surfaces touch 1-2 fields of `QuantickApp`; only `draw_frame` (14)
+and `draw_menu_bar` (17) are genuinely tangled. This is wiring debt, not
+architecture debt, which is why it is mechanical to repay.
+
+`grep` for "line count | file size | god object | monolith" in
+`arch-review/SKILL.md` returns zero hits. The repo already knows how to turn a
+rule into a test (`language_guard.rs`, `source_encoding_guard.rs`); size has no
+such guard. A rule enforced only by a skill's judgement drifts.
+
+## Classification
+
+Code change + hot path (per-frame draw dispatch) + user-visible + adds a
+capability (the port). All four gate sets apply.
+
+## Acceptance criteria
+
+1. `crates/app/tests/size_guard.rs` fails when any tracked file grows past a
+   versioned baseline, and passes at today's baseline.
+2. `arch-review` gains an accumulation dimension: it judges where registration
+   lines pile up, not only whether a new file was added.
+3. `new-extension` states that a UI surface docks through the `Surface` port,
+   never through a new `QuantickApp` field.
+4. A `Surface` port and registry exist in the `app` crate, with a fake second
+   implementation covered by a test (`new-extension`'s rule).
+5. At least 3 surfaces move onto the port; `QuantickApp` loses at least 3
+   fields (before/after counts recorded); `draw_frame` no longer hand-calls
+   them.
+6. Behaviour identical: `visual-qa` PASS on every moved surface, or defects
+   explicitly accepted.
+7. Per-frame performance flat or better: `APP_HEALTH_SUMMARY` fps/frame_avg
+   against a `main` control run, numbers in the PR body.
+8. Four checks green, run as separate commands; `arch-review` run with every
+   Blocker/Should-fix resolved or deferred in the PR body; PR opened.
+9. Every artifact in English.
+
+## Order of work (risk-ascending, so the gate survives a failed port)
+
+Items 1-3 are additive and carry zero production risk; they stop the bleeding
+on their own. The port (4-5) comes second and starts with the surfaces that
+touch a single field. If the port fails, it is dropped and the gate still
+ships.
+
+## Done means
+
+The PR is open with green CI and the evidence in its body. Merging is the
+trader's call, never part of the mission.
--- a/.claude/skills/arch-review/SKILL.md
+++ b/.claude/skills/arch-review/SKILL.md
@@ -122,7 +122,7 @@ Read the neighbouring code before judging any of it. The repo's existing
 pattern is the standard; a change that invents a second way to do something
 already solved is a finding, even when the new way is prettier in isolation.
 
-## The eight dimensions
+## The nine dimensions
 
 ### 1. The docking test — modularity and extensibility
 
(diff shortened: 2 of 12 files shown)


milocaetano:
CI note: the first run failed on a single test, `app::tests::the_trade_paint_layer_switch_stops_the_marks`, and passed on a re-run of the same commit with nothing changed. It is a flake, not a regression — the branch touches no code that test reads (its diff is `surfaces/`, the size guard, four skills and the edits in `app.rs` that move the three surfaces out), and `main` itself went red on a different single test earlier the same day (run 33324338457, `gateway_client_reads_the_running_application_and_wrong_tokens_fail_closed`). The test names its scratch directory after the process id, which is the shape of flake this repo already sees on PID reuse.

PR #305, opened 2026-09-05
refactor(app): the launch hooks leave the constructor, the paper wiring leaves app.rs

milocaetano:
Move the 721-line launch-hook sequence out of `new_with_workspace` into
`crates/app/src/app/launch_hooks.rs`, and the paper wiring into
`crates/app/src/app/paper_wiring.rs`, leaving `app.rs` as the window's
definition and nothing else.

**Mission tier: `medium`.** One receiver rename across 721 lines carrying an
order invariant, plus a ~380-line pure move.

`app.rs` **1,932 → 789 production lines** (1,961 → 824 total).

## What moved

| | before | after |
| --- | --- | --- |
| `app.rs` | 1,932 production lines | **789** |
| `new_with_workspace` | 932 lines (`:494`-`:1425`) | **213** (`:503`-`:715`) |
| the hook sequence | 721 lines inside the constructor | `launch_hooks.rs`, 10 appliers |
| the paper wiring | 9 methods in `app.rs` | `paper_wiring.rs`, 392 lines |

What is left in `app.rs`: the mods and consts, the struct, `new`,
`new_with_workspace`, `attach_surface`, the `eframe::App` impl, `mod tests`.

### The ten appliers, called in the block's original order

```
apply_launch_hooks             12 lines   apply_indicator_hooks       22
  apply_book_and_strip_hooks   11         apply_layout_hooks          85
  apply_control_hooks          53         apply_replay_hooks         124
  apply_rail_hooks             59         apply_dock_and_report_hooks 110
  apply_history_hooks          66         apply_workspace_hooks       73
  apply_tape_hooks            140
```

The module's doc comment states the order rule the sequence depends on —
day-before read before anything loads a session, `seed_new_panes` before any
autostart, the env hooks after the config defaults and the restored workspace,
the workspace writers last — because reordering them changes what a launch
opens with and no compiler catches it.

## Proof that only the receiver changed

The brief's recipe (`sed 's/\bapp\./self./g'`) does not work as written: three
multi-line receivers end a line with a bare `app` (`:1071`, `:1263`, `:1287`),
and a broader `\bapp\b` substitution would corrupt 18 `target: "quantick::app"`
string literals. The transform actually applied is **`app.` → `self.` plus
those three bare receivers, string literals untouched** — 74 of 722 lines
touched, 15 `quantick::app` literals left alone.

```
old block lines (renamed): 721
new bodies lines         : 721
identical                : True
```

The receiver-renamed old block is **byte-identical** to the ten new function
bodies concatenated in order.

```
old block : 68 sites, 52 distinct
new module: 68 sites, 52 distinct
lists identical, in order: True
```

### Every non-move hunk, quoted

There is exactly one, and it is rustfmt's:

```rust
-            self.active_tab_mut().tape_mut().set_live_lane_window(window);
+            self.active_tab_mut()
+                .tape_mut()
+                .set_live_lane_window(window);
```

`self` is one character longer than `app`, which pushed that line past 100
columns. Nothing else in the 721 lines changed.

### The paper wiring is a pure move

365 non-blank lines, multiset-identical between the removed region and the new
file, plus the new `impl QuantickApp {` block's own closing brace. Six private
methods widened to `pub(super)` because their callers are now the parent.
`git diff --color-moved=zebra` shows them as moves.

## Two things the brief predicted wrong

**1. The generated registry is *not* byte-identical, and could not be.**
`hooks::OWNERS` names the file a reader should open to find each hook, and a
guard checks the registering file is the reading file. The reads moved, so
`declare_hooks!` had to move — this is forced, not the optional choice the
brief called it. Fifty *Declared in* cells changed, `app.rs` →
`app/launch_hooks.rs`, +650 bytes. The hook **names** are byte-identical, which
is the half that matters (68 sites, 52 distinct, same order). The +650 is paid
inside the context budget by cutting 650 bytes of history and reasoning from
`ui-harness/SKILL.md` — what the file used to say about `harness.rs`, what a
misspelt hook used to look like, the field count behind the `Harness` argument,
the four edits the `Surface` port removes, and `ChartLayer`'s variant count.
None of them decided an outcome, which is the test `CLAUDE.md` sets for a line
in a `SKILL.md`. Signed in `context-baseline.txt`.

**2. `site.cfg_test` moved by 3, not 0** (622 → 625). The moved code took the
last *production* read of five names, and `app::tests` still reaches them
through `use super::*`. `app.rs` already documents this exact case and answers
it with `#[cfg(test)]`-gated imports rather than edits to `app/tests/`; three
attributes were added on that precedent, one removed. Every other `--report`
line moved only for `app.rs` and the two new files.

## Verification

- [x] `cargo fmt --all -- --check` — exit 0
- [x] `cargo clippy --workspace --all-targets` — exit 0
- [x] `cargo build --workspace` — exit 0
- [x] `cargo test --workspace` — exit 0

Each run separately, exit codes read individually.

`cargo test -p quantick-app`: **1,899 passing**, same as `origin/main` — no
test file is touched by this branch and the `#[test]` count is identical (1,113
on both). Guards: 138 + 16 + 5 passing, including
`the_generated_indexes_match_the_code_they_describe` and
`the_committed_registry_is_what_the_generator_emits`.

Baselines tightened: `app.rs` 1932 → 789, size `!budget` 52,139 → 50,996 — a
1,143-line reduction in the trunk budget, signed with its arithmetic.
`launch_hooks.rs` (880 production lines) and `paper_wiring.rs` (392) are both
under the 1,500 threshold, so neither takes an entry.

### Performance

Every touched path is **rare** — launch-time, once per process, before the
first frame. The sequence runs at the same point in the same order; the change
adds one non-inlined call per process. No per-trade, per-depth or per-frame
path is touched, so no hot-path measurement is owed.

### The launch hooks still reach their surfaces

Three scenes, each launched under one hook with the trader's stores redirected
to a scratchpad, then asked over the control plane what the app believes is on
screen. Hashes are over the snapshot scopes with volatile fields (timestamps,
instance id, revisions) removed, so a reviewer can reproduce them.

| hook | what the app reported | sha256 of the readout |
| --- | --- | --- |
| `QUANTICK_LAYOUT=flow` | **one pane** ("Flow chart"); the launch default is the two-pane time+flow, so the hook is proven by the difference | `a17f8b5e06c73bcc…` |
| `QUANTICK_CONTROL_PANEL=1` | `CONTROL_GATEWAY_ENABLED`; the **Local agent access** panel on screen, "Status: On — reading only" | `53b2f95d31a6cb02…` |
| `QUANTICK_REPLAY_AUTOSTART=paused` | `session.replay` → `replaying: true`, session `2026-08-17`, `playing: false` | `907a41d676996… ` |

No `UNKNOWN_HOOK`, no `AUTOSTART_UNKNOWN`, no `HOOK_REFUSED` in any of the
three runs. Health line at capture: 60 fps / 16.7 ms, so the frames are real.
Screenshots kept as attachments, unhashed — a PNG hash varies with the clock
and the live tape and would prove nothing.

## Reviews

**`arch-review`** — step 0: `code-review` at `low` (effort-first, no reuse
notice), **0 findings**. It independently checked the hook-read order is
byte-identical between the removed block and the new module, and that the call
site sits where the block began.

Shape pass, all nine dimensions (tier `medium`), graded against `46a6676`:
docking improved (the ui-harness gains a named address); performance flat
(launch-time only); trunk shrank by 1,143 lines with both baseline changes
signed; language clean by guard and by reading.

One **Consider**, deferred as a follow-up: the order the module doc calls "the
contract" is documented but unguarded — reordering the ten applier calls
compiles and passes every test. The invariant was equally unguarded before this
branch (it was implicit in the linear block), and a test for it is beyond a pure
move. Concrete fix for the follow-up: a test that asserts the ten
`apply_*_hooks` calls appear in `apply_launch_hooks` in the block's original
order, or that `seed_new_panes` precedes the first autostart read.

**`delivery-review`** — completeness pass (the `medium` mode), run inline.
Nothing `UNLEDGERED`: all 19 asks in the request map to a ledger line. No
criteria pass ran at this tier, so **this is not a full conformance review** —
the two criteria not literally met (A8, A11) are the ones written up above,
recorded rather than quietly dropped.

Assumption audit: **S1 should have been a check, not an assumption.** The brief
said the `declare_hooks!` placement was free because the registry would be
byte-identical either way; `hooks::OWNERS` in fact forces the declaration to
follow the reads, and that is what produced the +650-byte context raise. One
grep before assuming would have caught it.

## Follow-ups

- Guard the applier order with a test (the `Consider` above).
- The struct's 56 fields — the next cut, and deliberately out of scope here.


milocaetano:
CI note: the first `ci` run dropped `indicator_worker::tests::dropping_the_lane_clears_the_samples_already_published`. That file is not in this branch's diff, the test passes locally, and the re-run is green — the known random single-test drop under CI contention, not a regression from this change.

PR #306, opened 2026-09-05
feat: trades bars for MetaTrader B3, recorded from the venue's deal counter

milocaetano:
Add a `trades` bar kind for MetaTrader B3 that cuts a bar every N exchange deals: the venue's session deal counter, exact at each of its ~31-second readings and estimated per print between them, recorded to disk behind an explicit REC control so a recorded day reopens as the same chart and a change of N recuts it.

**Mission tier: `high`** ? bridge protocol, feed-mt5, engine, app chrome, persistence, control plane and harness all change; `delivery-review` ran in full. Goal file: `.claude/GOAL-archive-trades-bars-b3.md`.

## What the counter turned out to be ? and what the chart does about it

Running the branch on the live tape for a day measured the counter itself: the terminal refreshes `SYMBOL_SESSION_DEALS` about **every 31 seconds**, not every poll (592 readings on 2026-09-04, median interval 31.2 s, p10 30.3 s, p90 31.9 s, 1 500 to 10 000 deals apart). A bar cut from the readings alone closes once per reading whatever N is ? which is what the trader saw when changing N: nothing. The trader chose the per-tick estimate over stopping or cutting at the counter's resolution (goal archive, *D-late*). The engine now credits each print its contracts times the last completed window's deals-per-contract rate and re-anchors the total to the exact reading every time one arrives: the day's total and the bar count are the venue's, each boundary is an estimate within its 31-second window, the chip and the REC hover say *estimated*, and ? the rule using only what came before the print ? the live edge and a rebuild cut identical bars, so a change of N recuts the whole day from the recorded readings. A reading holds for ten minutes; REC turns amber after three missed readings. Proof: `another_n_recuts_the_same_day`, `interleaved_samples_cut_where_a_rebuild_cuts`, `a_reading_corrects_the_estimate_without_cutting_twice` (engine), `a_deal_count_change_recuts_the_bars` (app, through the toolbar's settle path).

## Why a tick bar is not the Trades chart

Measured on 2026-09-03 for `WINV26` against the running terminal: `SYMBOL_SESSION_DEALS` 5 821 205, trade ticks in `CopyTicks` 1 774 869, volume matching to within 61 contracts. MetaTrader folds ~3.3 B3 deals into one tick and keeps no count per tick, so `tick:2000` and ProfitChart's `2000T` cut in different places. The session counter, read by the bridge every poll and stamped on live ticks, is the only deal count there is ? which is why recording is part of the feature. `docs/deal-recording.md` has the full matrix.

## What ships

- **Engine** ? `DealBarBuilder` (`crates/engine/src/deals.rs`): joins prints to counter readings by time, closes at the session's multiples of N (ProfitChart alignment even when connecting mid-session), reports prints before the first reading as *uncounted*, never folds them. `BarBuilder` exposes an optional `DealCounterInput` port plus generic diagnostics; print-only builders implement no deal-counter operation.
- **Bridge + feed-mt5** ? `deals` stamp on live ticks and `deal_counter` in the hello (Python bridge and EA; `PROTOCOL.md`); `DealSampler` reduces stamps to one sample per change; `Mt5Event::DealCounter` ? `FeedEvent::DealCounter`; `FeedCapabilities::deal_counter`.
- **App** ? `trades` bar kind offered only where the feed carries a counter; a REC control beside the symbol with popover, chart-corner chip and status cell saying one of *recording / stale / recorded / no deal count*; recording per asset to `Documents/Quantick/deals/<SYMBOL>/<day>.deals` (delta text), resumed after a restart, recorded days in the popover and the history menu; record-by-default via `record_deals = true` on `metatrader-b3` with a Tools-menu override saved in the workspace; `trades:N` config spec, refused on any non-MetaTrader feed.
- **Control plane** ? `feed.status` carries `deal_recording`; cockpit capability `feed.deal_recording.set` starts, stops and loads a recorded day by name; catalog regenerated (40 capabilities); typed `layout.pane.set_bar_spec` drives the same recut path as the UI; recorder failures return structured control errors and refused compound calls leave the durable default unchanged.
- **Harness** ? `QUANTICK_DEAL_RECORDING=on|off|menu`, `QUANTICK_DEALS_DIR`.

## Docking (new-extension)

Ports: the engine aggregator trait, the `FeedEvent` channel, `FeedCapabilities`, the toolbar model, the action registry. New files: `engine/deals.rs`, `feed-mt5/deals.rs`, `app/deal_recording.rs`, `app/deal_recording_ui.rs`, `app/deal_recording_tab.rs` (an `impl Tab` beside the tab), `app/app/deal_recording_wiring.rs` (an `impl QuantickApp` child module), `app/control/deal_recording.rs`. Trunk diff: `app.rs` +24 (a `mod`, a field, restore/save, one call per surface), `tab.rs` +8, `pane.rs` +7, `stream.rs` +9 ? every raise is signed in `size-baseline.txt` and paid by tightening `app.rs`, 31 lines in all. Defaults: a config without `record_deals` records nothing; a feed without a counter shows no REC and no `trades`. Fake second implementations: the tick builder ignoring readings (engine test), a Binance-capability tab in the control test. Blast radius: `Trade` unchanged; every `FeedCapabilities` literal gains one field; `Mt5Status::Connected` gains one field; `BarKind` gains a variant (three exhaustive matches).

## Performance

Per-print: one compare in the feed and one push per *changed* reading. Per-poll: one buffered append, flushed once a second. Per-frame: one string compare per tab, three small view clones. Rare: a directory scan on start/stop/market switch, a rebuild on loading a day.

Same session tape (1 774 869 prints), tick(5000), idle after load, maximised, same hour:

| Run | fps | frame avg | frame cpu |
| --- | --- | --- | --- |
| `main` | 59?60 | 16.67 ms | 6.2?6.4 ms |
| branch, REC off | 59?60 | 16.67 ms | 4.5 ms |
| branch, REC on | 59?60 | 16.67 ms | 5.6?6.3 ms |
| branch, trades(2000), 2910 bars | 59?60 | 16.67 ms | 5.5?6.0 ms |

## Verification

- [x] `cargo test -p quantick-guards`; `python bridge/mt5/tests/test_*.py` (4 suites); `ruff check --select F`
- [x] Merged `origin/main` at `d3cf317` (the `quantick-feed` extraction, the generated hook registry and capability inventory, the owned test scratch); the branch follows each, and the merge commit says how ? and again at `7dc2e15` (#294, the history reach into the feed crate; one import conflict) ? and again at `cc4c92f` (#295?#304: app.rs, pane.rs and tab.rs split into sidecars, the inline tests moved out of the six largest hosts, the paper policy out of the ticket). The branch's thirteen app.rs docking lines re-homed into the sidecars, the spec switch kept the shared recut path, both ratchets re-signed on the measured numbers. All four checks green on each merged head, and on the final one (`final-checks-<sha>.log` in the review dossier).
- [x] `arch-review` ? see below
- [x] `delivery-review` ? PASS with G8's ordering clause deferred by the trader; see below
- [x] `visual-qa` ? all cells PASS except two accepted below; `trader-ux-review` ? see below

Live check over `quantick-mcp` (branch build): `feed.deal_recording.set {load_day: 2026-09-03}` ? `loaded_days [2026-09-03]`, flow pane `trades/2000`, `closed_bar_count 2910` = 5 821 205 / 2000 ? the candle count ProfitChart shows for `2000T`.

## Accepted / deferred

- The disabled `trades` entry is photographed through the new `QUANTICK_BARS_MENU` hook (`cap_bars_menu_disabled_trades.png`, Binance tab); its hover reason stays unit-tested rather than photographed, because the pointer hook reaches only the candle area.
- The narrow-window cell: `QUANTICK_WINDOW_SIZE=1000x700` did not produce a 1000 px window under a restored workspace, so the toolbar's collapse plan with the extra REC width is not photographed.
- The market was closed during every capture, so the `recording` state shown is the pre-first-reading form; the counted form is exercised by tests, and the first live session is the proof that remains.
- The MQL5 bridge carries the same stamp but was not run.
- Follow-ups (out of scope by the trader's decision): Binance's exact per-print deal ids and pairing `.deals` sidecars with replay/export-session tapes. The codec, writer and scanner themselves now live in headless `quantick-replay::deals`, reachable by backtest and MCP without linking the app.

### delivery-review

Full mode at tier `high`, over the final head `6a5b439` (checklist source: the goal archive on the branch; completeness pass run inline on the strong model, nothing UNLEDGERED ? the one phrase outside the earlier table, the purpose clause about conflicts being easy to resolve, is R2's; every `R` COVERED, R10 and R11 delivered in the session before the branch as the ledger records). Criteria pass from a fresh-context `sonnet` subagent over the dossier (the diff, its stat and log, the check logs per head, the visual-qa and trader-ux reports, the captures, the markers): 17 DELIVERED, **G8 PARTIAL** on exactly the clause the trader deferred on 2026-09-04. Escalation on the strong model over G8 alone: **DELIVERED** ? the ordering clause reads as satisfied by the deferral, and the remainder holds on the shipped rule: the fixture is fixed and its expected bars are derived by hand in the test's doc comment from the estimate rule (rate 75 deals per contract, cuts at 1 004 000 and 1 006 000), `golden_cuts_at_the_sessions_multiples_of_n` asserts exact times, counts, prices, reading, rate and progress, `same_input_same_bars` and `interleaved_samples_cut_where_a_rebuild_cuts` guard determinism two ways, and nothing of the earlier reading-only rule survives in the file. Its note for the record: the estimate rewrite (`2ae5fbd`) also landed fixture and code in one commit ? the same clause, the same deferral. **PASS**; the branch did not move between the first dispatch and the last verdict (HEAD, working tree and both markers checked before and after). Two remarks from the pass, neither a grade: the goal archive's evidence record for A9 quotes stale line counts for `tab.rs` and `pane.rs` (the real stat is app.rs +7, tab.rs +145, pane.rs +41 ? a bookkeeping error in the archive, and `app.rs`'s diff is exactly a `mod`, a field and an init line); and A1's evidence bullet still describes the pre-estimate cut rule, which the archive's `## Decided on the live tape` section corrects. Neither was fixed after the verdict: a commit would stale both markers for a comment. The earlier passes over `47cf9a2` and `f2a521a` are in the goal archive's evidence record; their assumption audit flagged `S1` (what a bar does across a stretch with no counter readings) as the question the mission should have asked up front ? it was answered by the cadence measurement and the trader's decision above.
### trader-ux-review

Blocker 1 (the stale state compared a wall-clock epoch with an arrival delay and could never fire) and Should-fixes 3 (`no deal count` on a pane that was counting), 4 (REC's width moved BUY/SELL), 5 (REC missing from the collapse budget), 6 (popover stretched by the full path), 7 (Stop?Start rebuilt tick panes too), 10 (`since` was the first reading, not the file's start) and 11 (`N trades` beside `trades`) are fixed in the follow-up commit. Deferred, with the reason:

- **Should-fix 2** ? a stretch without readings (a restart, a stopped counter) folds into the bar the last reading was forming, drawn as an ordinary N-deal bar. The follow-up commit first cut it as *uncounted*; the fourth round removed that rule on purpose, because it made live and rebuilt bars differ (the one thing the engine must never do). The honest fix is the recorder writing an explicit end-of-coverage marker on stop ? a file-format addition, the same follow-up issue arch-review names below.

- **Should-fix 8** ? an empty trades pane explains itself only in the corner chip (`no deal count ? N prints before have no count`). A centred pane notice like the offline one needs the empty-pane note surface to take a second author; follow-up.
- **Should-fix 9** ? a loaded day's row says "loaded" before that day's tape is paged in. Paging back on load is a history-reach change; follow-up. The chip and `feed.status` say which days are loaded.
- **Consider** ? the Tools toggle reads the active feed's config even on a Binance tab; a symbol round trip forgets a hand that stopped REC; the `recorded` chip also shows on tick panes; the view is cloned three times per frame.

### arch-review

Shape pass at tier `high` (all nine dimensions): no Blocker, seven Should-fix. Fixed in the follow-up commit: readings loaded behind the live tape land through one sort (`ChartState::observe_deals_batch`) instead of an insert per reading; the day scan caches per file by size and mtime so a stop re-parses only the file it closed; the recorded-day list is shared (`Rc<[RecordedDay]>`) so a view is a refcount, not a copy; `STALE_AFTER_MS` is the engine's `READING_MAX_AGE_MS`, one rule stated once; `since_unix_ms` is described as what it holds and `first_reading_unix_ms` joins it; `chart.summary` carries `uncounted_prints` so the chip's number exists as data. Deferred:

- **Should-fix (test-coverage)** ? two claims proven by reading only: the stream sends a reading *ahead* of its print (`feed-mt5/src/stream.rs`), and a market switch stops and flushes the old recorder before the new one starts (`app/deal_recording_wiring.rs`). Both need a TCP-harness or a tab-switch fixture this branch does not add; follow-up, named here.
- **Consider** ? `deals_n = 2_000` as a named default; `ProviderKind::may_count_deals()` as the validator's one home; the popover's "? 3 to 12 deals" resolution is a WIN number; `ShowAsTrades` is click-only (no capability sets a flow pane's bar kind for any rule ? pre-existing).

### arch-review ? the review chain, and where it stopped

Three rounds plus a confirmation pass, findings shrinking every round (shape: 7 ? 2 ? 1 Should-fix; bugs: 8 ? 8 ? 0 new). Round 4's confirmation over the round-3 fixes found every prior finding fixed, no Blocker, and one Should-fix ? `Recording::open` judged a file *fresh* before cutting a torn last line away, so a crash mid-header left a day that never got one ? fixed in `b9d37aa` with a test that fails without it, together with the Consider that a reload's flush error went unrecorded. The merge of `origin/main` after that was verified by the guards and the four checks rather than a fifth round: the chain's budget is spent, and the resolution is mechanical (a field re-applied on the moved `FeedCapabilities`, two hooks declared, both registries regenerated, tests moved onto the scratch module). Deferred Consider items from that pass: `READING_MAX_AGE_MS` lives in the engine while only the app reads it; the `counter stuck at 0` label pair has no test of its own; one baseline comment wraps past 80 columns.

**Round 5** (step 0 at `medium` over the head that recorded G8's deferral, `f2a521a`) came back with eight confirmed findings, seven real after reading the code ? four of them in the recorder's state machine written in round 4, which is the shape the budget rule calls a design problem rather than a finish problem. The trader chose one more round over deferring open Blockers. Fixed in `c889d6d`: the engine joins a print to the newest reading *strictly before* it, never at its own millisecond, because two readings can share one (ticks do across poll rounds) and the at-or-before join placed live and rebuilt bars one reading apart there ? a test feeds both orders and compares the cuts, and the fixtures move their readings one millisecond ahead of the prints they stamp, as the feed sends them; a series reset (Reload, a replay seek) keeps the readings the panes held, so a morning counted with REC off is not uncounted afterwards; a recorded day on disk is listed and opens with no bridge connected; a resumed day is the recording, not a loaded one, so Stop after a restart reads Off; the recorder is keyed by feed *and* symbol; the open file keeps the day it was named for across a display-timezone change; the no-counter chip is reachable. Deferred: a corrupt terminated line refuses the day (a round-3 decision ? a whole bad line is not a torn tail) and the button then reads Off rather than an error state; the popover and `feed.status` carry the reason, and an error state for the button is a follow-up.

**Round 6** (step 0 at `medium` over `c889d6d`): eight findings again, and the trader ran the branch on their own workspace ? footprint layer on ? which found a ninth: the footprint series asserts on the first deal bar, because a print a deal bar leaves uncounted was still folded into the ladders. Fixed in the sixth commit: one helper feeds the ladders and skips what the builder left uncounted (a refold feeds the readings first, as a rebuild does); the reset that keeps readings was also the reset a symbol switch and a replay open took, so the old market's readings cut the new one's prints ? the pane now has a plain reset and a keeping one, and only the Reset event takes the latter; with the panes holding the readings, the reseed from disk on every Reset (a file parse per replay seek) is gone; a header torn before its fields starts the day over instead of refusing it; the wire's `state` documents `unsupported`, which a counter-less tab with a recorded day on disk now answers. Deferred from this round, each a recorded follow-up: after a session rollover, a later reading in the same join that already reaches the new session's first multiple closes that bar one print late (live and rebuilt agree; the fix needs a push that can return two bars); the day scan parses every line of every day file on the UI thread once per launch and market switch, because the delta encoding has no footer to read the last reading from ? a footer or sidecar is a file-format addition; the *Record deals by default* toggle has no control-plane call or readable field; the day file takes no lock, so two writers on one symbol (a second process on the shared folder) could interleave lines; an out-of-order live reading is retained but not fed to the live builder; a transient `0` from the terminal reads as a rollover; `day_of`/`fmt_hms`/`fmt_count` duplicate helpers that already exist elsewhere.

**Round 7** (step 0 at `medium` over `568e6a1`): eight findings, three of them real Blockers in the previous round's code. Fixed in the seventh commit: opening or closing a replay went through the reset that keeps the counter readings, so the live market's counter cut the replay's prints ? the reset now takes a flag, and only the Reset event and a feed reload keep them; the REC control is not drawn over a replay; the rebuild after a resumed file or a loaded day goes through the pane and does the bookkeeping a spec switch does (pagination revision, indicator replay); a reading older than the newest held re-cuts the series at once through the batch rule; the validator asks the provider whether it may count deals instead of naming it; on a market switch the old recorder stops before its scan cache moves on; the corner chip takes the offline chip's geometry and clamp. Deferred, recorded above or here: the rollover-then-boundary edge in the engine; the per-Stop rescan of the day file just closed; a spec of `trades:N` on a MetaTrader session that turns out to have no counter stays on `trades` with the chip saying so (a runtime downgrade or feed notice is a follow-up); the corner-chip painter is still a second copy of the offline chip's, on the same numbers.

**Round 8** (step 0 at `medium` over `44748db`): eight findings, seven real and all in the two previous rounds' code. Fixed in the eighth commit: a replay opening puts the live market's readings aside and its closing brings them back (a morning counted with REC off survived a symbol round trip but not a replay one); a recorder still writing says so across a feed reload, which resets the capabilities until the next hello; a start reached through the control plane on a source with no counter is refused with its reason; a reading left over from last night does not name today's file; closing a day lists it from what the recorder wrote, under the flushed file's stamp, instead of reading the day back; a reading older than the newest held is kept for the next rebuild rather than re-cutting the live edge per reading (a clock that went backwards delivers one per poll for an hour), and is held once whatever shares its millisecond. Deferred: a one-poll dip of the counter reads as a session rollover and cuts two bars the venue did not ? a debounce is a rule the engine would have to state, and the recorded file would carry the dip either way; the per-pane copy of the reading series (the same shape as the per-pane trade retention).

**Round 9** (step 0 at `medium` over `76bf058`): eight findings, five fixed in the ninth commit ? a live reading at the newest held millisecond with a lower count (a regression the sampler forwards) was fed to the builder and read as a session rollover, cutting a bar the venue never cut; it is held at its sorted place for the next rebuild; a replay opened over a replay overwrote the readings put aside with the first replay's empty panes; a start refused before the hello disarmed the config's default for the session; `feed.deal_recording.set` on a replaying tab answered as if the feed had no counter, and now refuses with its reason; the file's time delta saturates. Declined or deferred: the record-by-default fires again on the next market a tab opens after a hand-stop ? by design and documented ("the new market's REC starts on its own default"); *Show as trades* is click-only because no capability sets a pane's bar kind for any rule (pre-existing, recorded in round 1); the rollover-then-boundary edge in the engine (recorded).

**Round 10** (step 0 at `medium` over `c54cbb1`): eight findings, three fixed in the tenth commit, and one of them changed a rule. A morning Reload kept last night's readings, today's unstamped backfill joined the last of them, and the session's first (lower) reading closed the whole morning as one "2000-deal" bar ? so a reading now holds for four seconds of tape (`READING_MAX_AGE_MS`), and a print further behind the newest reading before it is *uncounted*, like a print before the first reading, never folded into a bar the counter did not cut. Still one rule and still order-free (the age is against the reading in force, which both orders hold), unlike the between-two-readings hole rule round 4 removed; it also answers the catch-up-round case (a 30 s stall caught up in one poll no longer folds into one oversized bar). The chip names a stuck or stale counter with REC off too, and never says "recording" over a replay. Deferred, recorded: the rollover-then-boundary edge (reproduced by the pass, still one print late and identical live and rebuilt); a bridge restarted with a larger clock offset writes nothing for an hour with no label (needs an operator override to trigger); the day scan's cost grows with the recordings folder (footer or sidecar, a file-format addition); every pane keeps its own copy of the reading series (a tab-level series is the follow-up).

**Rounds 11?12** (step 0 at `medium` over `d0e2f12` and `5704439`). Round 11 found the four-second age rule of round 10 dropping the first print after every lull longer than four seconds ? the reading arrives stamped at that print's own millisecond, which "strictly before" cannot see ? and that finding, with the trader's live report that changing N changed nothing, led to measuring the counter itself (above). The eleventh commit fixed the rest of that round: a dip smaller than a bar's worth of deals is a late poll, not a rollover; the footprint ladders stay aligned when a rollover ends a bar on an uncounted print; a rotation into an existing day file hands its readings back; the day being recorded cannot be loaded as a recorded day; *record by default* is a control-plane field and call. Round 12, over the estimate, found its edges: an unchanged reading re-emitted by a reconnect set a rate of zero; a window holding an uncounted print inflated the next rate; a multiple reached by a reading right after a close got no bar; the print that opens the bar after a rollover could not close its own; the feed dated a reading at the first tick of its round rather than the last tick the previous reading covered; a start did not write the reading already in hand; the counter's health was judged in two places. All fixed in the twelfth commit, each with a test. Deferred from these rounds: the same-millisecond duplicate held twice on the live path (identical output for any N ? 2); the raw `== BarKind::Trades` comparisons where `needs_deal_counter()` exists; the replay gates living in the tab rather than the recorder.

**Round 13** (step 0 at `medium` over `140a345`): eight findings, seven fixed in the thirteenth commit. The one that mattered most was UX: the rebuild after a resumed file or a loaded day did none of the index bookkeeping a spec switch does, so a level drawn on a 14:00 bar stayed on its index while three hundred morning bars grew in front of it ? a spec switch and readings under folded bars now re-cut a pane through one path (`Tab::recut_pane_with`: view, marks, indicators, strategies). In the engine: a reading two or more multiples past the boundary gives each multiple its bar, one per print, so the bar count stays the venue's over N, while the multiples crossed during a stall get none; a drop by more than half is a restart, so a count still below N can restart. The feed dates a reading at its own tick across a gap longer than a reading holds for (a connection kept alive overnight). The recorder writes a reading stamped behind the file's last line instead of skipping it. The stale hover says what the builder does. The day's name comes from the paper journal's calendar. Deferred, recorded: the chart holds an older live reading for the next rebuild where the builder drops it ? the live edge is never re-cut per reading (round 9's decision), so a bridge restarted with another clock offset moves bars at the next rebuild rather than live.

**Round 14** (step 0 at `medium` over `4f3da20`): five confirmed, five plausible; six fixed in the fourteenth commit ? the retained series sorted by time alone and stably, so a same-millisecond dip replays as the live edge saw it; a refold with a held reading is a rebuild; the recut gates on the rule the builder cuts under, not the selector; an unchanged reading after a halt renews the age; the countdown counts from where the bar opened and never past a bar; the symbol's folder takes the paper journal's mapping. Deferred, recorded: a session whose hello says `deal_counter: false` under a pane already on `trades` renders an empty pane with the chip's reason and no fallback block (the runtime downgrade named earlier); one corrupt terminated line refuses the day until the file is removed by hand (round 3's decision; a quarantine is a follow-up).

**Round 15** (step 0 at `medium` over `8703991`): eleven confirmed. The one that mattered undid a round-12 mistake: dating a reading at the *previous* round's last tick had the engine credit the round's prints again on top of a total that already held them ? the bridge reads the counter after the fetch, so the reading covers its own round; the feed now dates it at the round's last tick (ticks share their send time within a round) and emits it when the next round begins. Also fixed: a reading past the boundary while a bar formed before an empty window left that bar open past its multiple; the footprint refold rebuilt the bars behind the pane's bookkeeping (held readings now stay out of the refold instead); a loaded day beside a live counter read "recorded" for the rest of the session; the day file could rotate backward on a behind-stamped reading; a recut after readings disarmed strategies for a "spec change"; a placeholder recorder scanned the working directory. Deferred, recorded: the day scan on the interface thread (a per-file summary or a worker is the follow-up); readings dated behind the newest are held for the next rebuild rather than re-cut live (round 9's decision ? the chip reads *counter stale* meanwhile).

**Round 16** (step 0 at `medium` over `119eed6`): eight findings, seven fixed in the sixteenth commit ? a restart no longer inherits the catch-up the old session skipped on an empty window; a restart is a drop by more than half only, since a drop by a bar's worth is what a refresh one poll behind looks like at the open; the sampler releases the reading still waiting for its round when the heartbeat says no tick has come for a second (the last reading of a connection waited for a round that never came); an older reading is inserted after its millisecond's run, the order the stable time sort keeps; *record by default* is applied after every refusal, so a refused call changes nothing; the close-replay path that keeps the replay attached does not bring the live readings back under it; an empty `QUANTICK_DEALS_DIR` is no folder rather than the working directory. Declined with a measurement: a throttle on the bridge's `symbol_info` read (0.02 ms per read from the client cache). The same commit adds `QUANTICK_BARS_MENU`, the hook the delivery review's visual cell was missing.

**Round 17** (step 0 at `medium` over the merged head `d411a83`): eight findings, six fixed in the seventeenth commit, each with a test where one can prove it ? the sampler ends a round on a new counter stamp as well as on a new send time (an in-process bridge can send two rounds inside one millisecond), and dates a pending reading under the offset its ticks were mapped with, so a heartbeat restating the offset cannot move it; the connection's end flushes the reading a round was still waiting on; the bar-kind combo's disabled reason has one owner (`bar_kind_reason.rs`) and offers `trades` wherever a count is in hand, as *Show as trades* already did, instead of calling it impossible before the hello; a footprint refold with a held reading is a rebuild, which retires the per-sample filter and the held list (a flag now); `quantick-backtest --bars trades:N` refuses with the true reason and lists the kind; the engine's module doc states the half rule the code applies. Deferred, recorded: the restart heuristic itself ? a reading lower by more than half is a restart, which a late poll at the very open could mimic (no regression was seen in the 592 readings measured; the follow-up is a yardstick from the last window's size); a pane seeded mid-session cuts with readings the source pane holds for its next rebuild (round 9's live-edge decision); a control-plane call setting the flow pane's bar spec, missing for every kind (recorded in round 1, now also in `docs/deal-recording.md`). Round 17's fixes are verified by their tests and the four checks rather than an eighteenth pass: the chain's budget is long spent, the findings have shrunk to edges of hypothetical bridges and wording, and the trader's goal for this branch is the PR.

**From the trader's own run.** Two reports while the branch ran on the live tape: the MT5 data lagging in bursts (measured: `symbol_info` costs 0.02 ms per read, so the counter stamp is free; `copy_ticks_from` for one second of WIN tape costs 44 ms median and 149 ms worst at 11:00, so the bridge's 20 ms poll falls behind on bursts and catches up ? the terminal's cost, shared with `main`; the history paging ran only at the opening); and a change of *N deals* with no visible change ? the path is the one *N ticks* takes, and `a_deal_count_change_recuts_the_bars` walks it (three bars every two deals become one every four); the egui field commits typed text on Enter, and a drag moves 50 per pixel.

**The shape of the chain.** Seventeen bug passes, each capped at eight findings and each filling its cap: rounds 1?3 over the original code, rounds 5?17 mostly over the previous round's fix ? and round 11's finding is the one that, with the trader's live report, exposed the counter's cadence and changed the design. Blockers shrank from design-level (a join rule that differed live and rebuilt) to edge-level (which reset keeps which readings) ? converging, but not closed by the reviews alone: the panic the trader's own workspace found (round 6) came from running the branch, not from a pass. The trader chose each extra round over deferring open Blockers.

### arch-review step 0

`step 0: code-review at medium (effort-first, no reuse notice), 8 findings, 8 confirmed, 3 plausible noted` over `4ecddcb`. Fixed in the round-2 commit: the deal sampler now follows the heartbeat's server offset like the tick mapper; the engine judges a hole only *between* two readings more than four seconds apart, never at the live edge (a round fetched after a stall spans minutes under one valid reading taken after it), and a rollover print passes the hole check before opening a bar; two readings at one millisecond are both written (the live chart joined prints to both); a failed file open is reported once, not retried per reading; a context pane opened later is seeded with the flow pane's readings; the bar-kind combo lists `trades` disabled with its reason where the feed has no counter instead of hiding a selected kind; *Show as trades* acts on the focused pane it describes; the EA declares and stamps the counter only when the terminal reports the property (three-argument `SymbolInfoInteger`); a torn last line ? a crash mid-write ? is cut away on resume instead of refusing the day. Deferred:

- **Round 3** (shape pass over the round-2 head): six of seven round-1 Should-fixes confirmed fixed, one new Blocker (a recorded day batched over a live series that already held two readings at one millisecond folded nothing, so the rebuild read a rollover ? fixed: sort by time *and* reading), one Should-fix (a whole bad line was treated as a torn tail ? fixed). Deferred: the hole rule differs between the live edge and a rebuild for prints under a stopped counter or inside a post-stall round (live: joined to the reading in force; rebuild: uncounted once the next reading is in hand). The honest fix is the recorder writing an explicit end-of-coverage marker on stop, which is a file-format addition ? follow-up issue. Also deferred: tests for a late context pane's seeded readings and for *Show as trades* on the focused pane.
- **Round 3 bug pass** (over the round-2 head) came back with eight confirmed findings ? the same count as round 1, which the review budget reads as "go to the trader", so this PR is opened with them resolved and the shape stated plainly. Fixed in the fourth commit: the hole rule is gone ? one join rule, a print joins the newest reading at or before it, identical live and rebuilt (the divergence both reviews had flagged), with a stretch without readings folding into the bar it interrupts; a feed Reload or replay seek re-seeds every pane with the recorder's readings; a failed file open turns recording off with its reason instead of a red button over dropped readings; an empty day file is fresh again and a header-only one keeps one header; the Python bridge documents that a counter at zero is stamped as zero (the chart shows a counter that never moves as `REC ? counter stuck at 0`); the EA evaluates the tape probe before the counter check; tests never touch `Documents/Quantick` (the stray `deals/TESTUSDT` files they had written on this machine were removed); a trades pane on a feed with no counter carries the `no deal count ? this source has no deal counter` chip; the day-scan cache survives a market switch. Still open, by design: the explicit end-of-coverage marker (follow-up issue), and the cosmetic amber flash for one poll after a catch-up round.
- **`quantick-backtest --bars trades:N`** is refused with the kinds it knows. The backtest has no source of counter readings (a `.deals` reader reachable from a headless crate is the same follow-up as replay's), so the one-engine promise is not yet kept for this rule; recorded here rather than shipped as a runner that cuts nothing.

?? Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01GnTCTYs72iWQk95PKRoFV5

## Refactor and current-main integration (2026-09-11)

- Integrated current `origin/main` and resolved all conflicts at `e8f591cb`.
- Reused `SpecSelector`, split `Tab` owners, and `ChromeState`; removed parallel toolbar state and avoided widening `QuantickApp`.
- Kept v1 observer additions backward compatible and regenerated schemas/catalog from code.
- Passed formatting, workspace Clippy, workspace build, workspace tests, 225 guardrail scenarios, all MT5 bridge Python suites, and Ruff undefined-name checks.
- Preserved the documented behavior and accepted deferrals from this PR. No visible interaction/layout changed during the refactor, so existing UI evidence remains applicable.

Evidence is committed under `docs/evidence/pr-306-refactor/`.

## Final AI-review repairs (2026-09-12)

The five open AI-review threads were fixed on final head `09e8d1aa`: recorder start/load failures are typed and propagated; validation precedes `record_by_default`; `Show as trades`, the toolbar and `layout.pane.set_bar_spec` share one recut path; the `.deals` codec/writer/scanner moved below app into `quantick-replay`; `DealCounterInput`, centralized replay seeding and named `BarInputAvailability` replace the defaulted deal methods, repeated loops and positional booleans. Focused error-path tests and the full fmt/clippy/build/test workspace loop pass. The follow-up AI report is https://github.com/milocaetano/quantick/pull/306#issuecomment-5643835311 and reports 5 fixed, 0 open, no new FAIL/WEAK.

Latest base `origin/main` is `57767f25`; it was merged cleanly before the immutable review head. Local verification is recorded under `docs/evidence/pr-306-refactor/`. Current GitHub Actions run: https://github.com/milocaetano/quantick/actions/runs/34675967819.


milocaetano:
CI note: the first `ci` run dropped one app test, `the_trade_paint_layer_switch_stops_the_marks` ("the marks kept painting with their layer off (309 vs 224)"), with the other 1,946 green and `windows` green. The branch's only touch on that file is a `deal_counter: false` capability literal; the test passes locally three times in a row and is on the measured list of CPU-contention flakes that `main` drops at the same rate. The failed job was rerun and both jobs are green.

crates/app/src/control/deal_recording.rs
```diff
@@ -0,0 +1,347 @@
...
+                .position(|recorded| recorded.day == *day)
+                .ok_or_else(|| {
+                    ControlError::invalid_request(format!(
+                        "no recorded day '{day}' for {}; feed.status lists the recorded days",
+                        tab.symbol
+                    ))
+                })?;
+            tab.apply_deal_recording(DealRecordingAction::LoadDay(index));
+        }
+        if let Some(enabled) = input.enabled {
```
milocaetano: **FAIL (3. AI-ready): a failed start comes back as `Ok` with the failure in `error: Option<String>`**

`set` calls `apply_deal_recording(Start)` (line 258) and always answers `Ok(DealRecordingResult)`. When `DealRecorder::start` refuses (`crates/app/src/deal_recording.rs:691` placeholder, `:700` no counter declared, `:756` file open failed), it writes a sentence into `self.error` and returns. The caller only finds out by string-matching `recording.error` (line 102). `load_day` does the same thing on a read failure (`deal_recording.rs:876`).

Failure scenario: a script calls `{ "enabled": true }` on a tab whose day file cannot be opened (disk full, locked folder). It gets a success response with `state: "off"` and a prose `error`. It cannot tell "this call failed" apart from "an earlier flush failed and the message is still there": `error` is sticky, because `stop()` never clears it and a flush error from any frame stays until the next successful start or load.

Fix: have `start`/`load_day` return `Result<_, DealRecordingError>`, an enum (`NotConfigured`, `NoCounter`, `Open(io::ErrorKind)`, `Read { line }`), and map it to a `ControlError` in `set`. Keep the snapshot's `error` for the recorder's background state only. While there, make `DealRecordingSnapshot::state` (line 70) a serde enum so the schema lists the five words.

crates/app/src/deal_recording_tab.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// What the REC control asked for.
+    pub fn apply_deal_recording(&mut self, action: DealRecordingAction) {
+        match action {
+            DealRecordingAction::Start => self.start_deal_recording(metrics::wall_clock_ms()),
+            DealRecordingAction::Stop => self.deal_recorder.stop(),
+            // The selector's own path, on the pane the popover described —
+            // the focused one, like every bar control; the change is picked
+            // up a frame later by the same spec sync a toolbar click goes
+            // through.
+            DealRecordingAction::ShowAsTrades => {
```
milocaetano, line 129: **FAIL (3. AI-ready): "Show as trades" and the N-deals count can only be set by clicking**

Two things this diff adds can only be reached from the UI: `DealRecordingAction::ShowAsTrades` (here, from the popover's click at `deal_recording_ui.rs:237`) and the `N deals` DragValue (`toolbar.rs:601`). No capability puts the flow pane on `trades:N` or changes N. `layout.pane.set_interval` refuses pane 0 (`control/layout.rs:735`, "set by the toolbar's BARS group") and only knows `Time`. `feed.deal_recording.set` takes `enabled` / `load_day` / `record_by_default`, not a spec. The one test of the recut, `a_deal_count_change_recuts_the_bars`, drives `pane.spec` directly.

Failure scenario: an agent reads `feed.status`, sees `deal_count_available: true`, and wants the tab to cut every 2 000 deals. The only way is a mouse on the combo, or a config edit and a restart.

Fix: add a named call that sets a pane's bar spec from the canonical string (`trades:2000`, and in effect every kind), gated by the same `bar_kind_reason::disabled_reason`, refusing with that reason as a typed error. Map `ShowAsTrades` onto it.

crates/app/src/control/deal_recording.rs
```diff
@@ -0,0 +1,347 @@
...
+    // A replay is another tape: the live market's recorder is not reachable
+    // over it, and the answer says so rather than reading as "no counter".
+    if tab.replay.is_some() {
+        return Err(ControlError::invalid_request(format!(
+            "tab {} is replaying; the live market's deal recorder is reachable again when the \
+             replay closes",
+            tab.id
+        )));
+    }
+    if let Some(on) = input.record_by_default {
```
milocaetano: **FAIL (3. AI-ready): a refused call has already changed and saved the default**

`record_by_default` is applied here, before the `load_day` lookup at line 245 can refuse. The comment says "after every refusal", but only the replay refusal comes before it.

Failure scenario: `{ "record_by_default": false, "load_day": "1999-01-01" }` returns `Err("no recorded day '1999-01-01' …")`. `chrome.record_deals` has already become `Some(false)` on every tab and is saved with the workspace. A client that retries after fixing the day sees the default it asked for, but the first "failed" call already applied it, and a client that gives up on the error keeps a change it was told did not happen.

Fix: check every input first (tab, replay, `load_day` resolves to an index), then apply. Or move the `record_by_default` block below the `load_day` lookup. Add a test in `control_plane_tests` that sends this pair and asserts the default is unchanged.

crates/app/src/deal_recording.rs
```diff
@@ -0,0 +1,1762 @@
...
+    /// than after the torn line.
+    pub complete_bytes: u64,
+}
+
+/// Read one recording. A line the format does not describe is an error
+/// naming the line, never a sample guessed around it — except a torn last
+/// line, which is what a crash mid-write leaves and carries no sample: the
+/// file reads back to the line before it, and [`DealFile::complete_bytes`]
+/// says where a writer resumes.
+pub fn read_file(path: &Path) -> io::Result<DealFile> {
```
milocaetano: **WEAK (1. Modular / 2. Decoupled / 4. Agent-tested): the `.deals` file format is inside `app`, so nothing else can read a recording**

`deal_recording.rs` has four jobs in one file: the directory/env resolution and harness hook (`:90`, `:110`), the on-disk v1 codec (`Recording`, `read_file` here, `parse_sample_line`, `scan_days`), the recorder state machine (`DealRecorder`), and the surface wording (`button_label`, `headline`, `status_cell`). The codec and the day scan are headless, deterministic and have no egui in them, yet they sit in the one crate that nothing may depend on.

Breaking variant: `backtest trades:N` over a recorded day. `crates/backtest/src/bars.rs:110` already refuses `trades` because "no exported session carries" the counter, and the recording that does carry it lives in `quantick-app`, which `backtest` and `mcp` cannot depend on. So one-engine-three-consumers gets a feature only the chart can cut. It also means the error-path fixtures for the format (`a_broken_file_is_refused_with_its_line_named`, `a_torn_last_line_is_cut_away_rather_than_refusing_the_day`) only run under `cargo test -p quantick-app`, not below `app`.

Fix: move `DealFile`, `read_file`, `parse_sample_line`, the writer half of `Recording`, and `scan_days`/`RecordedDay` into a crate below `app` (`replay`, which already owns recorded tapes, or a `deals` module in `engine` next to `DealSample`), with their tests. `app` keeps `DealRecorder` and the view. Flips 1, 2 and 4 to PASS.

crates/engine/src/builder.rs
```diff
@@ -60,4 +60,24 @@ pub trait BarBuilder {
...
+
+    /// Hand the builder one reading of the venue's session deal counter.
+    ///
+    /// Only a rule that counts what the venue counts listens
+    /// ([`crate::DealBarBuilder`]); every other rule is fed by prints alone
+    /// and ignores the reading, which is why the default does nothing. A
+    /// consumer feeds every sample it holds through this one method whatever
+    /// rule is running, so switching rules never has to know which of them
+    /// wants the counter.
+    fn observe_deals(&mut self, _sample: DealSample) {}
```
milocaetano: **WEAK (5. Extensible): the venue's counter was added to the core bar-building path instead of plugging in as a separate input**

The deal counter reaches the builder as a new defaulted method on the core trait (`observe_deals` here, plus `uncounted_trades`). `ChartState` holds its own retained `deal_samples` series and `readings_held` flag. It replays them in three places: `rebuild`, the footprint refold in `set_footprint_group`, and `reset_series`. `observe_deals_batch` has its own sort and dedup rule. `bar_kind_reason::disabled_reason` takes one positional bool per capability (`traded_volume, deal_counter, deal_count_available`).

Breaking variant: the next bar kind fed by a second venue input, say open-interest bars on a futures feed, or a rule that also listens to depth. It adds another `observe_*` default to `BarBuilder`, another retained `Vec` and replay loop in each of those three `ChartState` sites, another `FeedEvent` variant, and a fifth bool on `disabled_reason`. That is five existing sites edited, when adding a new kind should only mean adding code.

Fix: give the builder one side-input type (`enum VenueReading { Deals(DealSample) }`, or a trait object) through a single `observe(&VenueReading)`. Have `ChartState` retain one ordered `Vec<VenueReading>` and replay it once in a shared `replay_into(builder)` helper that all three sites call. Pass `&FeedCapabilities` plus the in-hand flag to `disabled_reason` instead of positional bools.

milocaetano:
## AI review - PR #306 @ 3955718d7542a60b655e4e987510775daea02071
Identity: HEAD 3955718d7542a60b655e4e987510775daea02071; PR branch `feat/trades-bars-b3` (reviewed from the worktree `quantick-worktrees/refactor-pr-306-merge-ready`, local branch `refactor/pr-306-merge-ready`, which tracks `origin/feat/trades-bars-b3`); base ref `origin/main`; base tip e8eb23e543dcc827c16c7c552478e7ebf2150a4f (= merge-base); review key a92b506e6e7a905ad946682d630833f27bbb65e0
Scope: full diff, round one (no earlier ai-review report or thread on this PR); 92 files, +7198/−137. 5 findings, 5 unresolved / 0 accepted / 0 fixed. Threads: `PRRT_kwDOTfuoRs6htNL5`, `…htNRP`, `…htNSk`, `…htNY5`, `…htNZ0`.

1. Modular:      WEAK - `crates/app/src/deal_recording.rs` has four jobs in one file: env/dir resolution and the harness hook (:90, :110), the v1 on-disk codec (:219–:523), the recorder state machine (:527–:949), and the surface wording (:985–:1060). The other new files each do one thing: `engine/src/deals.rs`, `feed-mt5/src/deals.rs`, `deal_recording_tab.rs`, `deal_recording_ui.rs`, `control/deal_recording.rs`, `bar_kind_reason.rs`. Blast radius: 75 pre-existing files edited; the most edited is `crates/app/src/state.rs` (+571/−20, about 330 of those lines tests). Thread `…htNY5`.
2. Decoupled:    WEAK - the engine and feed side keep the right direction: `DealSample` lives in `engine` (`engine/src/deals.rs:76`), `feed-mt5` sends it through `Mt5Event::DealCounter` (`stream.rs:374`), `feed` forwards `FeedEvent::DealCounter` (`feed/src/lib.rs:161`), gated on `FeedCapabilities::deal_counter` and not on the provider name (the test at `control_plane_tests` builds a Binance handle with the capability on). But the `.deals` codec and `scan_days`, which are headless, sit in `app`, so `backtest`/`mcp` cannot reach a recording (`backtest/src/bars.rs:110` refuses `trades` for exactly that reason). Thread `…htNY5`.
3. AI-ready:     FAIL - `feed.deal_recording.set` is registered (`control/actions.rs:194`) with typed input and output and is safe to call twice (`start` returns early when already enabled, `load_day` dedups through `observe_deals_batch`). But: a failed start or load comes back as `Ok` with a sticky `error: Option<String>` (`control/deal_recording.rs:257`, thread `…htNL5`); "Show as trades" and N deals are reachable only by clicking (`deal_recording_tab.rs:110`, `toolbar.rs:601`, thread `…htNRP`); and a refused call has already saved `record_by_default` (`control/deal_recording.rs:231`, thread `…htNSk`).
4. Agent-tested: WEAK - the engine is exemplary: `cargo test -p quantick-engine interleaved_samples_cut_where_a_rebuild_cuts`, `another_n_recuts_the_same_day` and `a_reading_corrects_the_estimate_without_cutting_twice` fail without the diff, sit below `app`, have expected values worked out by hand in the fixture doc (`deals.rs:424`), and cover the error paths (`rejects_zero_size`, `a_sample_going_back_in_time_is_ignored`, uncounted prints). `feed-mt5` `deals::tests` does the same at the sampler. Hunt: no `HashMap`, `rand` or thread timing. `SystemTime` appears only as a file-mtime cache key (`deal_recording.rs:479`). `wall_clock_ms` is read in `app` and passed into the recorder (`deal_recording_tab.rs:20`). The env read is guarded by `cfg!(test)` (`deal_recording.rs:91`). WEAK because the file-format error fixtures (`a_broken_file_is_refused_with_its_line_named`, the torn-line tests) only run under `-p quantick-app`, and no capability test covers a failed start. Thread `…htNY5`.
5. Extensible:   WEAK - `BarKind` stays a closed enum over the domain (`state.rs:38`) and the capability is appended to the name-keyed registry (`actions.rs:194`), both fine. But the side input reaches the core through defaulted methods on `BarBuilder` (`builder.rs:72`), a retained series that `ChartState` replays in three places, and one positional bool per capability in `disabled_reason` (`bar_kind_reason.rs:12`). The next bar kind fed by another venue input would edit all of them. Thread `…htNZ0`.
6. Scalable:     PASS - per trade: `fold_print` adds two `uncounted_trades()` reads (`state.rs:181`) and `DealBarBuilder::push` is O(1) amortised (`deals.rs:304`); per tick in `feed-mt5`: `DealSampler::observe` is O(1) with no allocation (`feed-mt5/src/deals.rs:97`); per reading (the counter refreshes about every 31 s): one append per pane (`state.rs:249`), one buffered line (`deal_recording.rs:299`), a flush once a second (`:328`), and one `symbol_info` per bridge pump round. Rare: `observe_deals_batch` sorts and dedups once per load (`state.rs:264`), and a rebuild replays readings plus prints (`state.rs:913`). An out-of-order reading costs one `Vec::insert` and holds the recut until the next rebuild (`state.rs:235`), so the live edge is never recut per reading. Buffers are bounded by the day's readings (~600–1 200), which the doc comments state.
Top fix: `crates/app/src/control/deal_recording.rs` - make `start`/`load_day` return a typed `DealRecordingError` mapped to `ControlError`, validate every input before applying `record_by_default`, and add a `pane bar spec set` call that `ShowAsTrades` routes through - flips: 3 → PASS (then moving the `.deals` codec below `app` flips 1, 2 and 4).

milocaetano:
## AI review - PR #306 @ 09e8d1aa0b5661da8b554b94f1c579b8884ae3f0

Identity: HEAD `09e8d1aa0b5661da8b554b94f1c579b8884ae3f0`; PR branch `feat/trades-bars-b3`; review worktree `C:/src/quantick-worktrees/refactor-pr-306-merge-ready`; local branch `refactor/pr-306-merge-ready`; base ref `origin/main`; base tip `57767f25a63d7849c07663e599e1c2942613cbe2`; review key `162f1aab409341e7bb3e2db64a7f3dab1072cc1f`.

Scope: follow-up over the five open finding IDs from the full review; 0 unresolved / 0 accepted / 5 fixed. The narrow regression hunt found no new FAIL introduced by the fixes.

1. Modular: PASS - the headless `.deals` codec, writer, and scanner now have one owner in `crates/replay/src/deals.rs:1`; the app delegates storage mechanics and retains recording/UI policy. The repair's largest edited pre-existing file is `crates/app/src/deal_recording.rs`, which shrank by 333 lines while the extracted owner is independently testable.
2. Decoupled: PASS - `quantick-replay::deals` is publicly reachable at `crates/replay/src/lib.rs:59`, below the desktop app and already available to headless consumers. `DealCounterInput` is a separate optional port at `crates/engine/src/builder.rs:10`, so print-only builders implement no deal-counter no-op.
3. AI-ready: PASS - recorder start/load return `DealRecordingError` through checked methods (`crates/app/src/deal_recording.rs:476`, `:646`) and the capability maps them to a stable `ControlError`; input-dependent validation precedes the durable default update (`crates/app/src/control/deal_recording.rs:229-274`). The registered typed `layout.pane.set_bar_spec` action lands in `Tab::set_pane_bar_spec` (`crates/app/src/control/layout.rs:303`, `crates/app/src/deal_recording_tab.rs:152`), the same path used by Show as trades, and repeated identical calls report no change.
4. Agent-tested: PASS - `cargo test -p quantick-replay a_broken_file_is_refused_below_the_application` covers the extracted codec's error path below app (`crates/replay/src/deals.rs:298`). `cargo test -p quantick-app checked_start_names_an_unavailable_counter_instead_of_reporting_success` proves the typed start refusal, and `feed_status_carries_the_deal_recorder_and_the_capability_moves_it` proves atomic default handling plus the registered bar-spec action. No new randomness, iteration-order output, or timing wait was introduced; file mtime remains only a rare scan cache key.
5. Extensible: PASS - venue readings attach through `DealCounterInput`; all retained-sample rebuilds use the single `seed_deal_counter` helper (`crates/app/src/state.rs:24`); availability is carried by named `BarInputAvailability` fields (`crates/app/src/bar_kind_reason.rs:6`) instead of positional booleans. A future side input can add a port and diagnostics without patching every builder with a pretend operation.
6. Scalable: PASS - per trade, the optional-port observation and diagnostics reads remain O(1), with no allocation or history scan. Per counter reading, the engine appends once. Rare rebuild seeding is O(retained counter samples), centralized and unchanged in asymptotic cost. `.deals` scanning is rare and caches by length/mtime; writing remains buffered and append-only.

Top fix: none - all prior findings are fixed; no new FAIL or WEAK verdict remains.

milocaetano:
## AI review follow-up - PR #306 @ d9938a3d6f8609632e6ae9a3906920deb3cfc807
Identity: branch `feat/trades-bars-b3` reviewed from `refactor/pr-306-merge-ready`; base `origin/main` at `57767f25a63d7849c07663e599e1c2942613cbe2`; review key `7d8ef992b55663b44107bb426593b673e237abc0`.
Scope: traceability-only delta from `09e8d1aa`; the archive removes self-referential G5 and classifies delivery review under C1, as required by the mission contract. Existing five findings remain fixed; zero open threads; no source, test, runtime, capability or generated-contract change.
1. Modular: PASS - no product module changed.
2. Decoupled: PASS - no dependency or ownership change.
3. AI-ready: PASS - prior typed action/error fixes unchanged.
4. Agent-tested: PASS - prior proofs unchanged; final-head workspace loop passed.
5. Extensible: PASS - prior ports unchanged.
6. Scalable: PASS - no runtime path changed.
Top fix: none - no new FAIL or WEAK.

PR #307, opened 2026-09-05
refactor(app): tab.rs sheds its sidecars — history, feed, strategies, canvas, tests

milocaetano:
Move the four cohesive groups that lie away from `Tab`'s struct and layout code — venue-history paging, feed lifecycle, strategy runner, canvas painter — plus the four inline test modules, out of `crates/app/src/tab.rs` and into sibling modules under `crates/app/src/tab/`, the way `app/` and `pane/` already hold theirs.

**Tier: `medium`.** `arch-review` and `delivery-review` (completeness pass) both ran and both closed.

## What moved

| File | Lines | Holds |
| --- | --- | --- |
| `tab/history.rs` | 1,003 | `merge_older_candles`, `trim_to_seam`, `trim_borrowed_to_seam`, `seam_bucket_ms`, `OlderCandles`, the venue-candle methods and the trade-history page run |
| `tab/feed.rs` | 1,007 | capture generations, feed switching and presets, `drain_feed*`, `drain_book_feed*`, replay open/close, reconnect, reload, gap marking |
| `tab/strategies.rs` | 265 | `run_strategies*`, `any_alarm_armed`, `apply_strategy_cleanup`, `reset_market_state`, `close` |
| `tab/canvas.rs` | 629 | `CanvasChrome`, `trading_pane`, `draw_canvas`, shared picks and interactions, focus rule, collapsed rail, divider |
| `tab/tests/mod.rs` | 697 | `shared_routing_tests`, `move_pane_tests`, `collapse_path_tests`, `trading_pane_tests`, names kept, bodies at their original indentation |

`tab.rs` 4,472 → 1,661 production lines. No sidecar imports another: each names only `Tab` and a const or two, so the four concerns read independently.

## Bodies unchanged — how that is proven

Not by reading the diff. A line multiset over the original `tab.rs` against the new five files returns **only** these differences, and no body line at all:

- **Seven visibility widenings** — a private method in a child module is not visible to `tab.rs`, a sibling, or `tab::tests`:
  `trading_pane`, `apply_shared_interactions` (canvas); `request_ohlcv_history`, `take_ohlcv_history`, `settle_history_page`, `abandon_history_run` (history); `run_strategies` (strategies).
- **One call qualified** — `feed::demo_gap_ms()` → `quantick_feed::demo_gap_ms()`, because `mod feed;` now occupies the name `feed` inside this module.
- **Import lines**, redistributed from `tab.rs` to the sidecars.

`git diff --color-moved=zebra` renders the rest as moves.

## Re-exports (no import changes elsewhere)

```rust
pub use canvas::CanvasChrome;
pub use history::OlderCandles;
```

`crate::tab::CanvasChrome` (`app/frame.rs`) and `crate::tab::OlderCandles` (`toolbar.rs`, app tests) still resolve. The only files this branch touches outside `crates/app/src/tab*` are `crates/guards/size-baseline.txt` and the archived goal file.

## `BOOK_DRAIN_BUDGET` — the call the brief asked for

It **stays in `tab.rs`, unchanged and un-widened**, together with `BOOK_GENERATION_STRIDE`, `FOCUS_RULE_PX` and the rail constants. A child module sees its ancestor's private items, so no widening was needed at all — and `BOOK_GENERATION_STRIDE` is `pub` and imported as `crate::tab::BOOK_GENERATION_STRIDE` by the app tests, so moving it would have cost a re-export for nothing.

## Hooks

Nothing moved. `tab.rs` has exactly one real env read (`QUANTICK_PANE_COLLAPSED`, inside `new`) and one `declare_hooks!` site; both stayed. The generated registry and capability inventory are byte-identical. (The brief's ledger #3 claimed eight reads live in the feed and history groups — those eight are doc-comment mentions, not reads.)

## Verification

- [x] `cargo fmt --all -- --check` — exit 0
- [x] `cargo clippy --workspace --all-targets` — exit 0
- [x] `cargo build --workspace` — exit 0
- [x] `cargo test --workspace` — exit 0, no failures
- [x] `cargo test -p quantick-guards` — 159 tests, green
- [x] `cargo test -p quantick-app` — **1899 passed, 4 ignored**, identical to the branch point

**Ratchet.** Ceiling tightened 4,472 → 1,661; `!budget` 50,996 → 48,185, a fall of 2,811. No ceiling raised. Every new file is far below the 1,500 threshold, so none becomes tracked. `--report` before/after differs only in the `tab.rs` line, the ratchet numbers, `crate.lines.app` (+55, the module docs and `impl` wrappers) and `site.cfg_test` (625 → 622: the four `#[cfg(test)]` markers moved into an untracked tests path, leaving one). `footprint_render.rs` newly *appears* under `file.largest` — pre-existing and untouched, promoted into the top-N only because `tab.rs` left it.

**Performance.** Move-only, so flat by construction rather than by belief: no body byte changed and nothing crossed a crate boundary. Per-trade (`ingest_live_trade_at`, `drain_feed`), per-depth (`drain_book_feed`), per-frame (`draw_canvas`, `run_strategies`) and rare (paging, feed switch) all keep their instruction sequence.

## Review findings

One Should-fix, raised and resolved in-branch (`ddc4481`): the sidecars first opened with `use super::*;`, which no production sidecar under `app/` or `pane/` does. They now name their imports. That exposed what the glob was hiding — the three `egui::Rect` fields left in `Tab` are all `#[cfg(test)]`, so its `egui` import is now `#[cfg(test)]` too. `tab/tests/mod.rs` keeps `use super::*;`, the same line the modules had inline and what `pane/tests/mod.rs` does.

One **Consider**, deliberately not acted on: the four nested test modules keep a redundant `#[cfg(test)]`. Removing them would edit moved bodies, and the brief's ledger #9 explicitly sanctioned nesting inside `tests/mod.rs`.

No deferrals. Nothing open.

## Notes for whoever lands second

`refactor/gateway-rs-sidecars` meets this branch only at the `!budget` line of `size-baseline.txt` — re-run `cargo run -p quantick-guards -- --tighten` after merging.

Two corrections to the brief, both re-verified against the branch point `62c8730` (the brief measured `cc4c92f`, before #305 landed): `!budget` was 50,996 rather than 52,139, and every `file:line` in the brief still matched exactly. `history_reach_running` also travels to `history.rs` — it sits inside the range's gap and reads only the paging run's own state.

https://claude.ai/code/session_01GBS1HyoNMGjJC1Fqr9iVkP


milocaetano:
First `ci` run failed on a single test, `app::tests::layers_tests::the_trade_paint_layer_switch_stops_the_marks` ("the marks kept painting with their layer off (309 vs 224)"), with the other 1,896 green and the `windows` job passing. Re-ran the failed job; both are now green.

Recording why that was a re-run and not a cover-up:

- The test lives in `crates/app/src/app/tests/layers_tests.rs`, which this branch does not touch — `git diff origin/main...HEAD --name-only` lists only `crates/app/src/tab*`, `crates/guards/size-baseline.txt` and the archived goal file.
- This branch changes no body byte, so the paint path it exercises is instruction-for-instruction what `origin/main` runs.
- Locally it passed 3/3 in isolation, and in two full `cargo test --workspace` runs.
- It is one of the names on the known list for this repo's CPU-contention flake, measured on 2026-09-04 to fail on `main` at the same rate as on a branch: quiet machine 12/12 green both sides; under load, 3 reds in 9 branch runs and 3 reds in 6 `main` runs.

Shape matches exactly: one test, everything else green, unrelated to the diff.

PR #309, opened 2026-09-05
chore: throwaway fixture for the ai-review thread mechanics

milocaetano:
Throwaway. Proves that ai-review can post one resolvable review thread per finding, and that a second run opens none on unchanged code. Closed as soon as the evidence is captured.

### DIFF
--- a/fixture/ai_review_probe.rs
+++ b/fixture/ai_review_probe.rs
@@ -1,0 +1,30 @@
+//! A throwaway probe for the ai-review thread mechanics. Not in the workspace.
+
+use std::collections::HashMap;
+
+/// The next alert kind means editing this enum and every match on it.
+pub enum AlertKind {
+    Price,
+    Volume,
+}
+
+/// Owns a map and reports it in iteration order, which is not stable.
+pub struct AlertBook {
+    alerts: HashMap<String, AlertKind>,
+}
+
+impl AlertBook {
+    /// The only entry point is this method; nothing registers it anywhere.
+    pub fn render_labels(&self) -> Vec<String> {
+        self.alerts.keys().cloned().collect()
+    }
+
+    /// A failure is a String, so a caller cannot match on what went wrong.
+    pub fn arm(&mut self, name: String) -> Result<(), String> {
+        if name.is_empty() {
+            return Err("empty name".to_owned());
+        }
+        self.alerts.insert(name, AlertKind::Price);
+        Ok(())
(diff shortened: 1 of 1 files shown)


fixture/ai_review_probe.rs
```diff
@@ -0,0 +1,30 @@
...
+    /// The only entry point is this method; nothing registers it anywhere.
+    pub fn render_labels(&self) -> Vec<String> {
+        self.alerts.keys().cloned().collect()
+    }
+
+    /// A failure is a String, so a caller cannot match on what went wrong.
+    pub fn arm(&mut self, name: String) -> Result<(), String> {
+        if name.is_empty() {
+            return Err("empty name".to_owned());
+        }
```
milocaetano, line 26: FAIL - AI-ready: `arm` returns `Result<(), String>`, so a caller cannot match on what went wrong; an empty name and a full book are the same value. Fix: a typed error enum.

fixture/ai_review_probe.rs
```diff
@@ -0,0 +1,30 @@
...
+
+/// Owns a map and reports it in iteration order, which is not stable.
+pub struct AlertBook {
+    alerts: HashMap<String, AlertKind>,
+}
+
+impl AlertBook {
+    /// The only entry point is this method; nothing registers it anywhere.
+    pub fn render_labels(&self) -> Vec<String> {
+        self.alerts.keys().cloned().collect()
```
milocaetano, line 19: FAIL - Agent-tested: `render_labels` returns `HashMap` keys, so two runs over the same book return two orders and no test can assert an expectation. Fix: BTreeMap, or sort before returning.

fixture/ai_review_probe.rs
```diff
@@ -0,0 +1,30 @@
+//! A throwaway probe for the ai-review thread mechanics. Not in the workspace.
+
+use std::collections::HashMap;
+
+/// The next alert kind means editing this enum and every match on it.
+pub enum AlertKind {
+    Price,
```
milocaetano, line 7: WEAK - Extensible: `AlertKind` is a closed enum where the variants are the domain. It holds today; the breaking variant is a spread alert carrying two symbols, which edits the enum and every match on it. Fix: a trait object keyed by name.

milocaetano:
Evidence captured: three threads posted, one resolved by the fix, none opened on unchanged code by the second run.

Issue #364, opened 2026-09-10
test(app): the gateway handshake test is load-sensitive under a busy machine


milocaetano:
Second of two load-sensitive tests found while running #360's checks. The first is #361.

## What fails

`app::tests::control_plane_tests::gateway_client_reads_the_running_application_and_wrong_tokens_fail_closed`, during `cargo test --workspace`.

## Rate, and the condition

| Condition | Failures |
| --- | ---: |
| full suite, other `cargo` jobs competing for CPU | 2 / ~20 |
| full suite, run alone | 0 / 16 |
| campaign base `98f1882`, full suite | 0 / 6 |

It only appeared while this machine was running concurrent builds and test suites.

## Why it is not a product defect

The test is handshake-only. It discovers the instance, connects, then reconnects three times with a wrong bearer token, a wrong instance id and a wrong process nonce, asserting `control.auth_failed` each time. It never sends a capability request, so it never reaches `prepare`, `dispatch_prepared` or anything below them.

Discovery and connect both run against a real loopback socket under `CONTROL_HANDSHAKE_TIMEOUT_MS`. On a machine loaded enough, a handshake that would have been refused for the right reason can instead miss that deadline, and the assertion sees a different error code than the one it names.

Honest limitation: **the failing assertion was never captured.** Two attempts to reproduce it with output capture both came back clean (10 runs and 6 runs), and the two observations that did fail were seen only as the summary line. So the mechanism above is read from the test and its timeouts, not from a captured failure — it is the most likely explanation, not a demonstrated one.

## What a fix probably needs

The same shape as any timeout-bearing assertion under load: assert the refusal *reason* separately from the transport outcome, so a handshake that timed out is reported as a timeout rather than as a wrong-code failure. Whoever owns the local transport should say whether the handshake deadline is the right knob or whether the test should retry a transport failure before asserting.

Related: #361, the other load-sensitive test, in `worker_progress`.

milocaetano:
## Correction: the machine was out of disk, and that is the better explanation

The issue as filed attributes the failures to CPU contention. That attribution is probably wrong, and I am recording why rather than quietly editing it.

While reviewing the same branch, a reviewer reported `cargo test` aborting with:

```text
LLVM ERROR: IO failure on output stream: No space left on device
```

Checking the container:

```text
Filesystem      Size  Used Avail Use% Mounted on
/dev/vda        252G   38G  1.3M 100% /
```

[...]

Issue #361, opened 2026-09-10
test(app): the delayed-producer progress test asserts one interleaving nothing forces


milocaetano:
## What fails

`orderflow_worker::progress_tests::delayed_producer_bookkeeping_does_not_block_real_flush_or_sample_recovery`, at `crates/app/src/worker_progress/tests/protocol.rs:92`:

```text
assertion `left == right` failed
  left: Some(2)
 right: Some(1)
```

The second panic in the log, `fixture releases worker: Disconnected` at `crates/app/src/worker_progress/tests.rs:81`, is teardown unwinding after the first — not a separate defect.

Observed in CI on [run 34528139191](https://github.com/milocaetano/quantick/actions/runs/34528139191/job/103041998612).

## Why it races

My first reading — that the phase gate does not fence the write — is wrong, and worth recording so the next reader does not repeat it. In `ProgressHandle::applying` the write and the signal are already ordered:

```rust
{
    let mut state = self.lock();
    let mut sample_slot = self.try_sample();
    // ...
    if let Some(slot) = sample_slot.as_mut()
        && let Some(sample) = slot.sample
        && sample.ticket <= state.admitted
    {
        // ...
        state.last_sampled_ticket = Some(sample.ticket);   // worker_progress.rs:325
    }
    state.phase = Phase::Applying;
}
self.clock.phase(Phase::Applying);                          // :332 — fires `Hold::reached`
```

The write happens before the signal, so by the time `second.reached()` returns the worker has already decided. What is unforced is the **condition**: whether the producer has filled the sample slot yet. `try_sample()` reads a slot the producer thread writes asynchronously, and the test's name says the producer's bookkeeping is meant to still be delayed at that point — but nothing holds the producer there. When it happens to have landed, ticket 2 is consumed and `last_sampled_ticket` is `Some(2)`.

So the assertion pins one interleaving of producer and worker that the fixture does not actually establish.

## Rate

Filtered runs (`cargo test -p quantick-app delayed_producer_bookkeeping`):

| Revision | Failures |
| --- | ---: |
| `campaign/architecture-a` (`98f1882`) | 0 / 80 |
| `fix/idempotency-contract` (`7542f7d`) | 1 / 80 |

One against none at that size is noise around a rate near 1 %, not a difference between revisions. A filtered run under-reproduces: CI runs the whole 1,971-test binary in parallel, and that contention is what makes it appear.

## What a fix needs

[...]

PR #382, opened 2026-09-12
test(app): force the interleavings the two load-sensitive tests assert

milocaetano:
## Summary

Campaign child of #367 (Q1); closes on integration: #361, #364. Mission tier: **medium**.

Two tests made CI red at random, each by asserting an outcome that only holds when the machine is quiet. No production code changes: every hunk is inside a `#[cfg(test)]` module.

**#361 — the delayed-producer progress schedule.** The S2 step admitted ticket 2 on the worker thread while the producer was still recording its acceptance on the fixture thread. Whichever ran first decided whether the sample slot still held the unobserved ticket 1, so `last_sampled_ticket` came out `Some(1)` or `Some(2)` by luck. Every other send in that schedule is made against a worker parked in a phase hold; S2 was the one that was not. It now parks the worker in the `Phase::Idle` callback that closes the previous cycle, with `Gate::hold` — the mechanism the issue named.

A note on the reading. The issue asks for a hold on *the producer's* sampling, but the producer's sampling is `WorkerProgress::record_send`, which the fixture calls on its own thread and therefore already controls completely. The unforced party is the worker, and it reads the sample slot inside `SharedProgress::begin` *before* it signals `Phase::Applying`, so no existing callback can fence that read — and adding one would be the production change this mission was told not to make. Parking the worker establishes the same interleaving with no production change. The assertion is not weakened.

Parking there also lets the schedule state the S1 property outright instead of implying it: at the parked point the cycle has applied, published, retired and counted while the producer has recorded no acceptance at all.

**#364 — the gateway handshake test.** The failing assertion had never been captured; this mission captured it. It is **not** a handshake timeout:

```text
panicked at crates\app\src\app\tests\control_plane_tests.rs:852:5:
assertion `left == right` failed: the application frame must drain the queued gateway request
  left: 1
 right: 0
```

`drain_bounded_since` (`crates/app/src/control/gateway/server.rs:95`) stops the frame's drain once `CONTROL_UI_BUDGET_US` — 250 microseconds — has elapsed, and `begin_frame` spends part of that budget before the drain starts. On a loaded machine the prelude consumes the whole budget, the drain admits nothing, and the queue still holds the request. The product is behaving as designed; the test asserted that one frame is always enough. It now runs frames until the queue is empty, which still proves that application frames drain it and the socket thread never does.

A third point surfaced once that was repaired: `disable_test_gateway` ("test gateway did not stop cleanly") and its two sibling waiters counted 400 sleep iterations rather than elapsed time, so a loaded machine ran out of iterations before the gateway thread had its chance. They are bounded by wall-clock time now, and say so when they give up.

The issue's own recommendation is applied as well, because it costs little and the handshake path carries the same class of exposure: the refused handshakes report their refusal code through a helper that retries a transport-level failure (`control.instance_gone`) a bounded number of times and, if it never clears, reports the transport failure rather than a wrong refusal code. `control.auth_failed` is still asserted for the wrong bearer token, the wrong instance id and the wrong process nonce.

### Rate tables

One filtered test per process, `--test-threads=1`, while a loop of `cargo build --workspace` competed for all twelve cores.

| Test | Build | Runs | Failures | Failure captured |
| --- | --- | ---: | ---: | --- |
| #361 `delayed_producer_bookkeeping…` | before | 200 | **1** | `protocol.rs:92`, left `Some(2)` right `Some(1)` — the issue's own assertion |
| #361 `delayed_producer_bookkeeping…` | after | 200 | **0** | — |
| #364 `gateway_client_reads…` | before | 200 | **1** | `control_plane_tests.rs:852`, frame did not drain |
| #364 `gateway_client_reads…` | after the drain repair only | 200 | **1** | `mod.rs:1970`, "test gateway did not stop cleanly" |
| #364 `gateway_client_reads…` | after, final head | 200 | **0** | — |
| #361 `delayed_producer_bookkeeping…` | after, final head | 200 | **0** | — |

Whole `quantick-app` test binary, 26 runs of each build on the same machine: 3 failures each, none of them either of these two tests.

### Performance impact

Test-only, rare. Every touched loop is `#[cfg(test)]`: the drain helper runs per call site, the three waiters once per test setup or teardown, and `delayed_bookkeeping` once per test. No per-trade, per-depth or per-frame production path is reached and nothing grows with session length.

## Verification loop

Run locally at the code head, each command on its own, none chained:

- [x] `cargo test --workspace` — 1983 passed, 0 failed in the `quantick-app` binary
- [x] `cargo test -p quantick-guards`, and `cargo run -p quantick-guards -- --report` with every ratchet at its recorded value

Plus the rate tables above (local), and required CI at the final head.

## Reviews

**arch-review — step 0: `code-review` at `low` (effort-first, no reuse notice), 0 findings, twice.** Both invocations declined the diff with "every hunk lands in test or goal-doc paths, which this level does not review", so the bug pass was run independently instead, as the skill provides for. That pass read `worker_progress.rs`, `worker_progress/tests.rs` and `orderflow_worker.rs::run` against the new schedule and returned **clean**: the `Phase::Idle` hold is pushed while the Gate's FIFO is empty and the worker is blocked inside the Publishing hold, so it cannot be missed; the parked worker holds no ledger, sample or mailbox lock, so no deadlock; the parked assertions are deterministic; moving `acknowledge(a1)` cannot hang because `run()` sends flush acknowledgements after `finish()` and before the next `recv()`; and S2 is forced more strongly than the comment claims, since `record_send`'s `pending` guard short-circuits before the slot is even locked.

Verdict at `01a001c7`:

- **Correctness** — clean; nothing open.
- **Docking** — no port involved; the change adds a helper beside the helpers it belongs with.
- **Performance** — flat. Every touched path is `#[cfg(test)]`.
- **Operability** — no surface; the change adds no user-facing capability.
- **Proof** — the two tests themselves, measured: each fails on the base under contention and passes 200 of 200 at this head.
- **Accumulation** — trunk flat. No ratchet-tracked file moved; `--report` shows the size and context ratchets at their recorded values, and goal files are excluded from the context ratchet by design.
- **Language** — the guard passed inside `cargo test --workspace`; the prose, the branch name, the commit messages and this body were read here and are English.

**ai-review** — 6 of 6 PASS, 0 findings, 0 threads. Report: https://github.com/milocaetano/quantick/pull/382#issuecomment-5642156391

**delivery-review** — completeness pass (tier `medium`), inline. One gap found and repaired in the record: the request's A4 asks for the campaign merge to be read back, which a child never performs, so it had no line on the mission's checklist. It is now closing step **C4**, owned by the coordinator. An independent follow-up graded that prose delta and returned CARRY FORWARD.

## Notes for the reviewer

**Out of scope, and reported rather than fixed.** Two sibling control-plane tests carry the same class of exposure, and this mission's scope line excludes other tests:

- `gateway_a_client_that_never_reads_does_not_stall_another` — `control_plane_tests.rs:2183`. Measured under contention at **4 failures in 100** filtered runs on the base build and **0 in 100** at this head, failing on the base at the same assertion CI hit. It is what made the first CI run on this PR red, and it is pre-existing.
- `a_retry_that_races_its_own_first_call_is_refused_rather_than_acted_on` — `control_plane_tests.rs:5255`, one `run_frame` then `client.read().unwrap()`. The same 250 µs mechanism as #364; the one-line repair is the `drain_gateway_requests` helper this PR adds. Measured at 3 failures in 26 whole-binary runs on the base build.

Neither was touched, because fixing them is a scope change the campaign has to grant. **CI on this PR is exposed to them**, so a red run whose only failure is one of those two is not a finding against this diff.

**A deliberate choice worth an eye.** `protocol.rs` now asserts `!unaccounted.valid` at the parked point — it enshrines a transient invalid ledger. It is deterministic today and the comment beside it says why, but if the ledger ever learns to represent "accounting outstanding" without flipping `valid`, this line fails for a good reason.

### DIFF
--- a/.claude/GOAL-archive-load-sensitive-tests.md
+++ b/.claude/GOAL-archive-load-sensitive-tests.md
@@ -1,0 +1,274 @@
+# Mission: force the interleavings the two load-sensitive tests assert
+
+Two tests make CI red at random: `delayed_producer_bookkeeping_does_not_block_real_flush_or_sample_recovery`
+(#361) pins one producer/worker interleaving that nothing in the fixture
+establishes, and `gateway_client_reads_the_running_application_and_wrong_tokens_fail_closed`
+(#364) asserts a refusal code where a loaded machine can instead produce a
+transport timeout. A+ gate 3 requires default-branch CI green at the assessed
+revision, and every child PR of campaign #367 inherits the exposure.
+
+**Tier:** `medium`. Campaign child Q1 of #367, assigned at `medium` by the
+coordinator: two independent test harnesses in two subsystems, each needing a
+design decision about where the ordering belongs, but no product change.
+
+## Request ledger
+
+- **R1** — #361: give the fixture a way to establish the delayed-producer
+  interleaving instead of hoping for it. Verbatim: *"so the delayed-producer
+  interleaving is established, not hoped for"*. Weakening the assertion to
+  `Some(1) | Some(2)` is forbidden. *(issue #361 "What a fix needs"; Q1
+  adoption comment, Scope bullet 1; D1, D3)*
+- **R2** — #364: assert the refusal reason separately from the transport
+  outcome, so a handshake that timed out under load reports as a timeout and
+  not as a wrong-code failure; or retry a transport-level failure a bounded
+  number of times before asserting the code. The test must still assert
+  `control.auth_failed` for each wrong credential. *(issue #364 "What a fix
+  probably needs"; Q1 adoption comment, Scope bullet 2; D1, D4)*
+- **R3** — No production change to worker-progress semantics, handshake
+  deadlines or authority, and no other tests touched. Verbatim: *"Out of
+  scope: any change to worker progress semantics, handshake deadlines in
+  production, or other tests."* *(Q1 adoption comment, Scope; D2)*
+- **R4** — `cargo test -p quantick-app delayed_producer_bookkeeping` passes
+  200 of 200 filtered runs under a concurrent `cargo build --workspace`, with
+  the property under test (sample recovery after a delayed producer) still
+  asserted. *(Q1 adoption comment, A1; D5)*
+- **R5** — `gateway_client_reads_the_running_application_and_wrong_tokens_fail_closed`
+  passes 200 of 200 runs under the same contention, still asserting
+  `control.auth_failed` for each wrong credential. *(Q1 adoption comment, A2;
+  D5)*
+- **R6** — Both issues' rate tables are extended with the after numbers, in
+  the PR body and as a comment on each issue. *(Q1 adoption comment, A3; D5)*
+- **R7** — The change ships through a PR whose base is exactly
+  `campaign/lean-a-plus`. *(Q1 adoption comment, A4; integration contract)*
+- **R8** — Purpose: default-branch CI stops going red at random on these two
+  tests, so gate 3 can be assessed. *(Q1 adoption comment, opening paragraph)*
+
+## Decisions (from the coordinator)
+
+- **D1** — Both tests are fixed in this one PR.
+- **D2** — No production change to worker-progress semantics, handshake
+  deadlines or authority. Only test/fixture/harness code changes. A genuinely
+  required production change stops the mission and is reported instead.
+- **D3** — #361 is fixed by giving the fixture a way to hold the producer's
+  sampling (the same shape `Gate::hold` gives the worker's phases), so the
+  delayed-producer interleaving is established. Weakening the assertion to
+  accept `Some(1) | Some(2)` is forbidden.
+- **D4** — #364 is fixed by asserting the refusal reason separately from the
+  transport outcome, or by retrying a transport-level failure a bounded number
+  of times before asserting the code; `control.auth_failed` stays asserted for
+  each wrong credential when the handshake completes.
+- **D5** — Proof of A1/A2: 200 filtered iterations of each test while a
+  concurrent `cargo build --workspace` competes for CPU, before and after the
+  fix; both tables recorded in the PR body and on both issues.
+- **D6** — Tier `medium`: `arch-review` with `code-review` at `low` and the
+  full shape pass; `delivery-review` completeness pass only, inline;
+  `ai-review` completion required.
+
+## Assumptions
+
+- **S1** — *Wanted to ask.* D3 asks for a hold on the producer's sampling. The
+  producer's sampling is `WorkerProgress::record_send`, which the fixture calls
+  on its own thread: its timing is already fully under the test's control, so a
+  producer-side hold cannot remove any interleaving. The unforced party is the
+  worker consumer, which reads the sample slot inside `SharedProgress::begin`
+  before it signals `Phase::Applying`, so no existing phase callback can fence
+  that read, and adding one would be the production change D2 forbids. The
+  fixture therefore establishes the interleaving with the mechanism D3 names —
+  `Gate::hold` — applied to the worker: the worker is parked in the `Phase::Idle`
+  callback that ends the previous cycle, so the producer's whole S2 step
+  completes before the worker can admit it. This is exactly the pattern the S3
+  step of the same schedule already uses; S2 was the one step that omitted it.
+  Safe to assume because it satisfies D3's operative requirement (the
+  interleaving is established, not hoped for) with no production change, and
+  because the alternative readings all require touching `worker_progress.rs`.
+- **S2** — *Wanted to ask.* Parking the worker in the `Phase::Idle` callback of
+  the S1 cycle moves the S1 flush acknowledgement after the producer's late
+  `record_send`, because the worker sends flush acknowledgements only after
+  `finish()` returns. The S1 property ("delayed producer bookkeeping does not
+  block the real flush") is preserved and strengthened by asserting, at the
+  parked point, that the worker completed the whole cycle — `Phase::Idle`,
+  `cycles == 1`, `mailbox_replacements == 1` — while the producer had still
+  recorded no acceptance at all. Safe to assume because the observable property
+  is stronger, not weaker.
+- **S3** — A handshake that misses its deadline surfaces as
+  `control.instance_gone` (`instance_gone_error()` in
+  `crates/control-local/src/client.rs`), which is the transport outcome D4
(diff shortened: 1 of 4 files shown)


milocaetano:
## AI review — PR #382 @ 7bb7ecd9e61a5e2ee0b448a2bcec6b2278c7f71e

**Identity:** head `7bb7ecd9e61a5e2ee0b448a2bcec6b2278c7f71e`; branch `fix/load-sensitive-tests`; worktree `C:\src\quantick-worktrees\fix-load-sensitive-tests`; base ref `origin/campaign/lean-a-plus`; base tip `e8eb23e543dcc827c16c7c552478e7ebf2150a4f`; review key `9f58627e454d1bc28957733f59b4a9dba6921023`.

**Scope:** full diff, round one. 0 findings; 0 unresolved, 0 accepted, 0 fixed. No FAIL or WEAK, so no threads were opened.

1. **Modular: PASS** — Three pre-existing files edited, all test modules; the largest edited is `crates/app/src/app/tests/control_plane_tests.rs` (+63/−26), which is the blast radius. Each new unit has one job: `crates/app/src/app/tests/mod.rs:1948` `drain_gateway_requests` runs frames until the queue empties, `crates/app/src/app/tests/control_plane_tests.rs:791` `refused_handshake_code` turns a refused handshake into its refusal code. The placement follows the repo's existing split — a waiter reachable from any test module sits in `tests/mod.rs` beside `wait_for_queued_gateway_requests`; a helper with one caller stays in that caller's file.
2. **Decoupled: PASS** — The new helpers take exactly what they read: `crates/app/src/app/tests/mod.rs:1948` takes `&mut QuantickApp` and `&egui::Context`, the same pair every neighbouring frame-driving helper takes, and owns the deadline itself rather than making each caller carry one. `crates/app/src/app/tests/control_plane_tests.rs:791` takes `&InstanceDescriptor` and `&ConnectOptions` by reference and clones only what `LocalClient::connect` consumes. No new `pub` item; `GATEWAY_TEST_WAIT` (`mod.rs:1890`) is private to the test tree.
3. **AI-ready: PASS (the diff cannot reach the question)** — No product behaviour is added or changed: every hunk is inside a `#[cfg(test)]` module (`crates/app/src/app.rs:660`, `crates/app/src/worker_progress.rs:505`). No capability, registry entry or entry point moves, so there is nothing for MCP to gain or lose. The one `String` the diff introduces (`control_plane_tests.rs:795`) is an error *code* being compared with `codes::AUTH_FAILED`, not a stringly-typed failure: the typed `ControlError` is matched on before the code is taken.
4. **Agent-tested: PASS** — The tests are the change. `cargo test -p quantick-app delayed_producer_bookkeeping` fails on the base under CPU contention (1 of 200 filtered runs, `protocol.rs:92`, left `Some(2)` right `Some(1)`) and passes 200 of 200 at this head; `cargo test -p quantick-app gateway_client_reads_the_running_application` likewise (1 of 200 before at `control_plane_tests.rs:852`, 0 of 200 after). Both run alone. On the hunt list: the diff *removes* thread-timing dependence from `worker_progress/tests/protocol.rs:70` by parking the worker in a phase hold, and *adds* `std::time::Instant` reads at `mod.rs:1893`, `:1922`, `:1950`, `:1979`. Those are ceilings on waiting, never expected values — no assertion is on elapsed time — and they replace an iteration count, which was the very thing that made the waits load-sensitive. The named breaking variant: a runner suspended for more than `GATEWAY_TEST_WAIT` would fail these for its own reason; that is true of any deadline and strictly better than the 400-iteration count it replaces. The tests live in `crates/app` because the code they cover does.
5. **Extensible: PASS** — `GATEWAY_TEST_WAIT` (`mod.rs:1890`) is one named budget that three waiters and the new drain share, so the next waiter appends a reference rather than repeating a magic number, and retuning it is one edit. `Gate` itself is untouched (`worker_progress/tests.rs`): the repair adds a fourth `hold` call to the existing FIFO rather than a new mechanism, so the next schedule step orders itself the same way.
6. **Scalable: PASS — rare, test-only** — Every touched loop is `#[cfg(test)]`: `drain_gateway_requests` runs per call site (twice in one test), the three waiters run once per test setup or teardown, and `delayed_bookkeeping` runs once per test. No per-trade, per-depth or per-frame production path is reached and nothing grows with session length. No buffer is added; the one bounded retry (`control_plane_tests.rs:797`, `ATTEMPTS = 4`) states its cap and panics loudly rather than dropping the attempt in silence.

**Top fix:** none. The diff is clean against all six dimensions.

milocaetano:
## AI review follow-up — PR #382 @ 01a001c7

**Identity:** head `01a001c7` (full: see the PR head); branch `fix/load-sensitive-tests`; worktree `C:\src\quantick-worktrees\fix-load-sensitive-tests`; base ref `origin/campaign/lean-a-plus`; base tip `e8eb23e543dcc827c16c7c552478e7ebf2150a4f`; review key `2ea84df43a7731a3fc0f50044059564b44be1c67`.

**Scope:** follow-up on the delta `7bb7ecd9..01a001c7` plus the full current diff. Prior report: https://github.com/milocaetano/quantick/pull/382#issuecomment-5642156391 (6 of 6 PASS, 0 findings). Findings: 0 open, 0 accepted, 0 fixed, 0 new; no threads exist to resolve.

The delta is one commit adding eight lines of prose to `.claude/GOAL-archive-load-sensitive-tests.md`: closing step **C4**, recording that the merge into `campaign/lean-a-plus` and its read-back belong to the campaign coordinator rather than to this child. An independent reviewer graded it against `docs/campaign/integration.md` and confirmed each claim it makes — that a child hands off and never merges (Authority and lifecycle; item 2, "The coordinator serializes merges"), that the coordinator reads back `mergedAt`, the merge commit and the campaign ref (item 5), and that the child issues are closed explicitly after that proof (item 5) — that its relative link resolves, that every added line and the commit message are English, and that it touches no executable code, test, fixture, schema, dependency, build or runtime configuration, hook or script.

**Verdict: every dimension carries forward unchanged.**

1. Modular: PASS — no unit added or moved.
2. Decoupled: PASS — no call site changed.
3. AI-ready: PASS — no capability, registry entry or entry point moves.
4. Agent-tested: PASS — no test, fixture or assertion changed; the 200-run measurements at the code head stand.
5. Extensible: PASS — no registry, enum or constant changed.
6. Scalable: PASS — rate unchanged; no loop touched, in production or in test.

**Top fix:** none.

Issue #403, opened 2026-09-12
test(app): the trade paint layer switch test races the shape count under load


milocaetano:
## Context

Campaign child of https://github.com/milocaetano/quantick/issues/367 (task key Q8). Base: `campaign/lean-a-plus`. Mission tier: small.

`app::tests::layers_tests::the_trade_paint_layer_switch_stops_the_marks` (`crates/app/src/app/tests/layers_tests.rs:565`) fails intermittently in CI under the parallel test binary and passes on rerun. Observed twice during campaign #367:

- PR #388, head `2aa644e3`: dropped as a shape-count race under contention (reported by the S3 child).
- PR #400, head `3dc0ebc1`, run https://github.com/milocaetano/quantick/actions/runs/34700339682 (`Test` step, 1 failed of 1,998), with nothing under `crates/app/src/app/tests/layers_tests.rs` or the paint layers in that PR's diff.

A+ gate 3 requires default-branch CI green at the assessed revision and criterion SE5 penalises implementation-derived or interleaving-dependent expectations; this is the same class as #361, #364 and #385, which the campaign already fixed.

## Scope

- Find what the assertion at `layers_tests.rs:565` depends on that the fixture does not force (frame count, a worker publish, a shape cache) and make the test establish it, the way #382 and #400 did for the gateway tests.
- Test-only change. Weakening the assertion is not the fix.

## Acceptance criteria

- [ ] A1: the test passes 200 of 200 filtered runs while a concurrent `cargo build --workspace` competes for CPU, before/after table in the PR; the property (switching the trade paint layer off stops the marks) is still asserted.
- [ ] A2: no production file changes.
- [ ] A3: merged into `campaign/lean-a-plus` through a PR whose base is exactly that branch, with the merge read back.

## Campaign assignment

Parent: https://github.com/milocaetano/quantick/issues/367. Stable key: Q8. Class: autonomous. Priority: 1. Risk: low, test-only. Executor: opus. Dependencies: none. Branch/worktree/PR: null until claimed. Counters start at operation=0, repair=0.
Validation plan: first row (test changed): the filtered test under contention, then the full ordered loop before every commit, CI at the final head.

milocaetano:
executor: opus - test-only harness fix in the layers test module (D13: opus for all remaining children); small tier; branch fix/layers-test-load-sensitive

milocaetano:
Q8 before/after evidence (PR #404, head `2e18b1be`, code at `5cc21dbf`).

[...]

PR #396, opened 2026-09-12
refactor(orderflow): split the projection and config files under 1,500 production lines

milocaetano:
## Summary

Split the two order-flow engine files that still carried signed ratchet
entries — `crates/orderflow/src/projection.rs` (1,850 production lines) and
`crates/orderflow/src/config.rs` (1,523) — into owned sibling modules under
each file's own directory, every file at most 1,500 production lines, with no
behaviour change and determinism intact.

Campaign child of #367 (S8); closes on integration: #375. Mission tier: `high`.

`projection.rs` was the whole heatmap pipeline in one file: the read model a
renderer draws, the two projection halves, the print tiers and the budget
fold. `config.rs` was three config families and their bounds in one file.
Each now has a root that owns the pipeline (or the parent config and the
shared validation) and siblings that own one phase (or one family) each. The
two entries have left `crates/guards/size-baseline.txt`, and the `!budget`
line fell by exactly the two ceilings retired — 9,264 → 5,891 on the current
campaign tip (`c1b4002e`, after #392, #395 and #399 landed; it was 12,473 →
9,100 on the first rebase's tip and 20,025 → 16,652 against the tip the branch
was cut from).

### What moved where

| New module | What it owns | Production lines |
| --- | --- | ---: |
| `projection/model.rs` | the read model: `PriceWindow`, `HeatmapCell`, `AggressionPrimitive`, `LiquidityEventPrimitive`, `BEFORE_CAPTURE`, `GapPrimitive`, `HeatmapProjection`, `SettledProjection` (with `empty` and `with_live`), `LiveMarks` — and, since the step-0 batch, the three rules a reader of those primitives shares with the pipeline: `event_cap_key`, `cap_events`, `normalized_log_intensity`, `normalized_area_size` | 513 |
| `projection/tiers.rs` | one tier of the tape: `TierCut`, `TierGrouping`, `TierClusters`, `cluster_tier`, `refine_tier`, `tier_primitives`, `aggression_primitive` | 315 |
| `projection/fold.rs` | the budget fold: `side_key`, `pane_budgets`, `merge_marks`, `MAX_FOLD_IDS`, `settle_ids`, `fold_chunk`, `FoldOrder`, `fold_to_budget`; and the one fold-only test | 312 |
[...]

PR #391, opened 2026-09-12
refactor(app): split the control-plane contract, evidence and gateway files under 1,500 production lines

milocaetano:
## Summary

Split the control plane's three over-ceiling files into owned sibling
modules, every file at most 1,500 production lines, with no behaviour change
and no authority path moved by one code path.

Campaign child of #367 (S5); closes on integration: #372. Mission tier: `high`.

`contract.rs` (1,718 recorded), `evidence.rs` (1,708) and `gateway.rs`
(1,696) each sat just over the ceiling, and the later AP1 and AP4 missions of
the campaign edit them. Each root now holds the part that decides — the
permission ceilings and the refusal sequence, the capture and its manifest,
the gateway lifecycle and the frame service — and a sibling holds one surface
each. The three entries have left `crates/guards/size-baseline.txt` and the
`!budget` line fell from 14,673 to 9,551 (the sum of the six entries that
remain at the campaign tip, `f574857b` after #392, which this branch is rebased onto).

### What moved where

| New module | What it owns | Production lines |
| --- | --- | ---: |
| `contract/reads.rs` | the observer tier's read capabilities: each `prepare_*` handler (payload decode, the checks a schema cannot make, the scopes a request reaches) and the invocation it becomes; `register_capability`, `read_capability`, `decode_payload`, `serialization_failed` and the two constants only they use | 524 |
| `evidence/store.rs` | the retained bundles: `EvidenceStore`, its state and bounds, the epoch that refuses a capture finished after access was withdrawn, and the chunk read that rechecks the grant, the cursor and the retention on every page — with its eight tests | 328 |
| `evidence/image.rs` | the screenshot: `RawScreenshot`/`ScreenshotPixels` as the application thread hands them over, PNG encoding, the scene's control regions scaled into the image, the two constants that say how (`SCREENSHOT_FORMAT`, `REGION_DECIMAL_PLACES`) and the `base64_len` budget helper | 231 |
| `gateway/local_action.rs` | the in-process door: `invoke_local_action` (with its control-trace intent/result lines), `invoke_local_read`, and the trusted actor they mint (`local_actor`, `hook_agent_actor`) | 315 |
| `gateway/trace_replay.rs` | the trace walk: `TraceReinjection`, `ReplayPosition`, `load_trace_for_reinjection`, and the per-frame `service_replay_trace` | 234 |

What stays in each root: `contract.rs` keeps `ObserverContract` whole —
`new` with every permission, profile, module, effect policy and capability
[...]

PR #404, opened 2026-09-12
test(app): force the frame the trade paint layer switch test asserts

milocaetano:
## Summary

Make `app::tests::layers_tests::the_trade_paint_layer_switch_stops_the_marks` deterministic under CPU load by forcing the frame its shape-count comparison depends on.

**Tier:** `small`. Campaign child of #367 (Q8); closes on integration: #403. Test-only: the diff is one `#[cfg(test)]` module and the mission archive. No production file changed.

### Root cause (captured, not guessed)

The CI failure (run 34700339682, attempt 1) was `the marks kept painting with their layer off (309 vs 224)`: the "off" count was **larger** than the "on" count. I reproduced that assertion verbatim on the base (table below), then instrumented the draws on a failing run:

```text
first=185 aggr=0 builds=0 live_end=None     <- worker has not recorded the prints yet
pre=185   aggr=0 builds=0 live_end=None
post=224  aggr=2 builds=1 live_end=Some(..) <- "on" count: lane laid out, projection requested without it
pre=174   aggr=2 builds=1
post=309  aggr=4 builds=2                   <- "off" count: the finished chart
```

On a quiet run the same chart settles at 359 on / 309 off.

The flow pane takes **two** trips through the order-flow worker to finish converging, and each trip lands whenever the worker thread gets a core:

1. `OrderflowView::live_lane` lays the live lane out only once the worker has published the newest print it recorded (`live_end_ms`). Until then the pane draws without a lane and asks for a projection with `lane: false`.
2. A draw paints the projection that the draw **before** it requested (`project_visible` returns the published frame, then sends the new request).

Under load, the "on" count caught the chart with the lane laid out but painting a projection requested without the lane (224 shapes). The "off" count came one round trip later and caught the finished chart (309). So the marks looked as if they kept painting. The product was behaving as designed. The fixture had assumed the worker is always one step ahead of the render thread.

A first fix that forced only the second trip (`draw → flush → draw` per count) made the test **worse**: 118 of 200 failed under the pinned harness, all with `309 vs 224`. That run is what exposed the first trip.

### The fix

- After the fixture's prints are drained, `tape_mut().flush_for_test()` waits until the worker has recorded them and published its live edge. The first draw then lays out the lane.
- Each count is taken as `draw → flush → draw`, so the measured draw paints the projection for the draw just before it.

`flush_for_test` is the existing bounded wait (`OrderflowWorker::flush`: a `Flush` ack with a 10 s `recv_timeout`) that the other order-flow tests already use. Neither step reads a clock. After step 1 no new data reaches the worker, so the `PROJECTION_INTERVAL` cadence in `project_at` cannot hold back a rebuild: a changed layout rebuilds immediately, and an unchanged one hits a clean cache.

Both assertions are unchanged: `marks_off < marks_on`, and the live paper layer is still visible after hiding closed-trade history.

### Before / after

Every run executes the filtered test with `--exact ... --test-threads=1`, one test per process. A loop of `cargo build --workspace` ran throughout in its own target dir, touching `crates/engine` each round. *Before* used the test binary built at the campaign base `86bbe06a`. *After* used the binary built at `5cc21dbf`.

| Harness | Build | Runs | Pass | Fail | Failure captured |
| --- | --- | ---: | ---: | ---: | --- |
| sequential, one process at a time (the issue's harness) | before | 600 | 600 | 0 | none: this 12-core machine never starved the worker |
| sequential, one process at a time | after | 200 | 200 | 0 | none |
| 20 processes at a time, all pinned to one core (`start /affinity 1`) | before | 200 | 190 | **10** | `layers_tests.rs:565`, `309 vs 224`, the CI assertion verbatim |
| same, repeated | before | 200 | 191 | **9** | same |
| same | first fix (second trip only) | 200 | 82 | **118** | same |
| same | **after** | 200 | **200** | **0** | none |
| same, extended | **after** | 1000 | **1000** | **0** | none |

The issue's harness alone does not reproduce the failure on this machine. The pinned harness recreates the contention of the two-core CI runner, and it reproduces the CI assertion exactly. Both are reported. The same table is posted on #403.

### Performance impact

Test-only, rate "once per test run". The only touched code is a `#[cfg(test)]` test body. It adds two worker flushes and two extra draws to one test, about 1 ms on a quiet machine. No per-trade, per-depth or per-frame production path is reached.

## Verification loop

Run locally at `5cc21dbf` (the code head), each command on its own, none chained:

- [x] `cargo fmt --all -- --check`: exit 0
- [x] `cargo clippy --workspace --all-targets`: exit 0, 0 warnings
- [x] `cargo build --workspace`: exit 0
- [x] `cargo test --workspace` (with `QUANTICK_BUBBLES` unset): exit 0; `quantick-app` 2000 passed, 0 failed, 7 ignored
- [x] `cargo test -p quantick-guards`: 164 passed

The archive commit `2e18b1be` is prose only; guards were rerun after it. Required CI at the final head is pending.

Changed files (`git diff --name-only origin/campaign/lean-a-plus...HEAD`): `crates/app/src/app/tests/layers_tests.rs`, `.claude/GOAL-archive-layers-test-load-sensitive.md`.

## Reviews

Graded at `2e18b1be` against `origin/campaign/lean-a-plus` at `f36cc022`, review key `11c116f8`.

**arch-review.** Step 0: `code-review` at `low` (level given as the first token, no reuse notice), 0 findings. Its `low` recipe skips test hunks, so the bug pass was also run here independently, as the skill provides for. That pass came back clean:

- The flush is the existing bounded `OrderflowWorker::flush` ack.
- The flow pane always has a tape, so `tape_mut` cannot panic here.
- Both assertions are unchanged.
- No clock-dependent cadence remains once the fixture's prints are in the worker.

The tier is `small`, so the shape pass read the dimensions the diff reaches: 2 (the declaration), 4 and 8. One Consider, not acted on: the pre-existing single settle draw is now partly redundant with `settled`'s first draw. It is harmless.

- **Correctness**: clean, nothing open.
- **Docking**: no port involved; the change is local to one test body.
- **Performance**: flat; the only touched path is a `#[cfg(test)]` test body, run once per test run.
- **Operability**: no surface.
- **Proof**: the test itself, measured: it fails 10 of 200 on the base under the pinned harness and passes 1200 of 1200 at this head. It is an app unit test.
- **Accumulation**: trunk flat; nothing tracked by the size ratchet moved.
- **Language**: the guard passed inside `cargo test --workspace`; the comments, branch name, commit messages and this body were read here and are English.

**ai-review**: 6 of 6 PASS, 0 findings, 0 threads. Report: https://github.com/milocaetano/quantick/pull/404#issuecomment-5646980800

**delivery-review**: not run, per the `small`-tier exemption (21 of 300 changed lines).

## Notes for the reviewer

- The campaign tip moved to `f36cc022` (#396, #391) after this branch was cut. Per D5 this branch is not rebased; the coordinator rebases at integration. The touched test module is disjoint from those merges.
- `a_hidden_layer_paints_nothing` in the same file solved the same class of problem another way: it hides every flow layer so the pipeline never starts. That was not an option here without changing what the chart under test looks like, so this test waits for the pipeline instead.

### DIFF
--- a/.claude/GOAL-archive-layers-test-load-sensitive.md
+++ b/.claude/GOAL-archive-layers-test-load-sensitive.md
@@ -1,0 +1,129 @@
+# Mission — force the frame the trade paint layer switch test asserts
+
+**Objective:** make `app::tests::layers_tests::the_trade_paint_layer_switch_stops_the_marks`
+deterministic under CPU load by having the fixture establish the state its
+shape-count comparison depends on, so default-branch CI stops going red at
+random on an assertion the product never broke.
+
+**Tier:** `small`. Campaign child of #367 (task key Q8) for issue #403: one
+test in one test module, no production file, no contract, wire or financial
+rule in reach. Nothing here is the trader's call, so no interrogation and no
+`delivery-review` (the small-tier exemption).
+
+**Campaign base:** `origin/campaign/lean-a-plus` at `86bbe06a`. PR base is
+exactly `campaign/lean-a-plus`.
+
+## Request ledger
+
+- **R1** — find what the assertion at `layers_tests.rs:565` depends on that the
+  fixture does not force (a frame count, a worker publish, a shape cache) and
+  make the test establish it, the way #382 and #400 did.
+- **R2** — *"Weakening the assertion is not the fix"*: the property — switching
+  the trade paint layer off stops the marks — stays asserted, and the live
+  paper layer assertion stays.
+- **R3** — test-only: no production file changes.
+- **R4** — proof: 200 of 200 filtered runs pass while a concurrent
+  `cargo build --workspace` competes for CPU, with a before/after table in the
+  PR and on #403.
+- **R5** — merged into `campaign/lean-a-plus` through a PR whose base is exactly
+  that branch, with the merge read back (the coordinator's step).
+- **R6** — the closing purpose: A+ gate 3, default-branch CI green at the
+  assessed revision, no longer at the mercy of this test.
+
+## Decisions (from the coordinator, answering step 3)
+
+- **D1** — test-only; a production change that turns out to be required is
+  reported, not made.
+- **D2** — reproduce first (200 runs under a concurrent build, capture a
+  failing assertion), then make the test establish the missing state with a
+  bounded wait on the observable condition; the property stays asserted.
+- **D3** — the 200-run table before and after, in the PR body and on #403.
+- **D4** — tier `small`: `arch-review` with `code-review` at `low`;
+  `ai-review` completion; no delivery-review.
+- **D5** — `gh pr ready` once, Bash tool, `cd` prefix; never merge, never touch
+  main, never rebase.
+
+## Assumptions
+
+- **S1** — the failing CI assertion (run 34700339682, attempt 1: "the marks
+  kept painting with their layer off (309 vs 224)") is the order-flow worker
+  publishing a projection frame between the "on" and the "off" draw: the
+  "off" count is *larger*, which only something arriving between draws can
+  explain. Safe: confirmed or refuted by the reproduction before any edit.
+  *Outcome:* confirmed, and refined — convergence takes two worker round
+  trips, not one (the live lane waits for the published live edge). A first
+  fix forcing only one trip failed 118 of 200 under the pinned harness.
+- **S3** — the D2 harness (one process at a time beside a concurrent
+  `cargo build --workspace`) did not reproduce on this 12-core machine: 600 of
+  600 passed on the base. The reproduction adds contention of the kind the CI
+  runner has: 20 copies at a time, every copy pinned to one core
+  (`start /affinity 1`), still beside the concurrent build. Safe: it forces
+  the same interleaving harder and reproduced the CI assertion verbatim; both
+  harnesses are reported.
+- **S2** — the shared `mod.rs` helper is left alone unless a second test needs
+  the same settling. Safe: reversible in one edit.
+
+## Acceptance criteria
+
+- [x] **A1** — the root cause is named from a captured failure, not a guess.
+      *Evidence:* the failing run's assertion output under contention.
+      → the PR body's root-cause section. *(R1)*
+- [x] **A2** — the test passes 200 of 200 filtered runs under a concurrent
+      `cargo build --workspace`, against a before run on the base binary.
+      *Evidence:* the before/after table. → the PR body and a comment on #403.
+      *(R1, R4, R6)*
+- [x] **A3** — `marks_off < marks_on` and the live paper layer assertion are
+      still asserted, unweakened.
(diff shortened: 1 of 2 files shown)


milocaetano:
## AI review - PR #404 @ 2e18b1be

Identity: HEAD `2e18b1be81d2d760f5f213f58e3f33a58a8204e3`; branch `fix/layers-test-load-sensitive`; worktree `C:\src\quantick-worktrees\fix-layers-test-load-sensitive`; base ref `origin/campaign/lean-a-plus`; base tip `f36cc0225b1adbdb9a086ec0b41280b01f5d600c`; review key `11c116f8fe7c9bd63524a0d6c4b9a2cf7c2930f1`

Scope: full diff (round one), 2 files: `crates/app/src/app/tests/layers_tests.rs` (+19/-2) and `.claude/GOAL-archive-layers-test-load-sensitive.md` (new, prose). 0 findings, 0 threads posted; nothing unresolved, accepted or fixed.

1. Modular: PASS. The diff edits one pre-existing file, `crates/app/src/app/tests/layers_tests.rs` (blast radius: 21 lines in that file). The change stays inside one test body, `the_trade_paint_layer_switch_stops_the_marks` (`layers_tests.rs:560-590`). The `settled` closure is local, like its sibling in `a_hidden_layer_paints_nothing` (`layers_tests.rs:491`).
2. Decoupled: PASS. It reaches the worker only through the existing crate-private test seam `OrderflowView::flush_for_test` (`orderflow_view.rs:769`), via the tab's `tape_mut()`. It adds no `pub` item and no new dependency on a concrete producer.
3. AI-ready: PASS. The diff cannot reach this question: it adds no behaviour and no call, and only changes how an existing test observes the chart.
4. Agent-tested: PASS. It runs alone with `cargo test -p quantick-app layers_tests::the_trade_paint_layer_switch_stops_the_marks`. It has to live in `app` because the subject is `ChartPane::draw_chart`, and moving it would need a production change, which D1 excludes. The thread-timing hunt is the point of the diff. The old version read the worker's publish timing (reproduced: 10 of 200 failing, `309 vs 224`). The new version waits on the worker's `Flush` ack (`layers_tests.rs:571`, `:574`) and never sleeps or reads an `Instant`. After the fixture flush no new data reaches the worker, so the `PROJECTION_INTERVAL` cadence in `project_at` (`orderflow/src/engine.rs:1092`) has nothing dirty to hold back. The expectation `marks_off < marks_on` (`layers_tests.rs:583`) is relational and not derived from the code under test.
5. Extensible: PASS. The diff cannot reach this question: there is no variant, registry or enum in the diff.
6. Scalable: PASS. Rate: once per test run. It adds two worker flushes (each bounded by the existing 10 s `recv_timeout`) and two draws to one test. No production loop, buffer or per-event path is touched.

Top fix: none; every verdict is PASS.

Issue #425, opened 2026-09-13
fix(control): the gateway writes an answer before it releases the request ID, so an immediate reuse is refused


milocaetano:
## What happens

Every terminal path in `crates/app/src/control/gateway/server.rs` writes the response first and releases the request ID afterwards:

- the worker path in `dispatch_prepared`: `send_response(...)`, then `global_in_flight.fetch_sub`, then `slots.forget(&request_id)`;
- the response worker for UI requests: `send_response(...)`, then `response_slots.forget(...)`;
- the parked wait's `WakeReason::Shutdown` branch, and the woken or timed-out wait, which goes through `dispatch_prepared` again.

A client that reads its answer and reuses the same `request_id` straight away can therefore reach the reader thread before the writing thread has run `forget`. It is refused with `control.invalid_request` / "request_id is already in flight on this connection", and that refusal is `retryable: false`. Contract §5.2 says a connection rejects duplicate request IDs *while the first request is in flight*; the park site's own comment says the ID is in flight "until the read is answered". A client that has read the answer is not violating that rule, yet a loaded machine refuses it.

## Evidence

Found by campaign #367 child Q10 (#418) while fixing #411. `gateway_rejects_a_duplicate_request_id_while_a_wait_is_parked` reuses the parked wait's ID immediately after reading the wait's answer. Instrumented under local contention (24 copies of the test binary on processor mask `0x3`), the reuse came back:

```text
Failure { error: ControlError { code: ErrorCode("control.invalid_request"), message: "request_id is already in flight on this connection", retryable: false, ... } }
```

15 of 65 instrumented runs; the other 50 succeeded.

## Proposed fix (not applied)

Release the ID (and the in-flight counters) before writing the terminal response, on every terminal path. The duplicate check still refuses a second request with the same ID sent before the answer: a client can only learn the ID is free by reading the answer, which the reorder puts after the release.

Q10's PR fixes #411 at the test layer only (it waits until the ID is released, retrying only that refusal) under decision D3. The product change is outside the decisions Q10 was given, so it is recorded as a `human_decision` in Q10's handoff. Once this is fixed, that wait in the test can go.

milocaetano:
[...]

PR #424, opened 2026-09-13
feat(mcp): default quantick_invoke to the newest registered version

milocaetano:
Default the MCP adapter's `quantick_invoke` to the newest version the connected instance registers for each capability when the caller omits `capability_version`.

**Tier:** `medium`. Campaign child of #367 (Q11); closes on integration: #422. Base: `campaign/lean-a-plus` (rebased onto `eb60ed41`). Authorized by the trader's D20 (https://github.com/milocaetano/quantick/issues/367#issuecomment-5650222968, "Sim, padrão = mais nova").

## Why

PR #421 added version 2 of seven `layout.*` capabilities and kept v1 registered; v1 now answers an honest refusal pointing at v2. `quantick_invoke` hard-coded version 1 for a version-less call, so an MCP agent that omitted the version met the refusal instead of the working answer.

## The resolution rule

- **Omitted version** → the highest `version` among the rows the instance's own `control.describe` registry lists for that `capability_id` (one row per registered version). Read at call time; no table in the adapter, no cache. The call is then sent to the instance that answered the describe, so the version and the answer come from one registry.
- **Explicit version** → sent unchanged, routed exactly as before. An explicit version the instance lacks is refused `control.capability_unknown`, never rounded down.
- **ID the registry does not list** → forwarded at version 1 as before, so the instance's own refusal (`control.capability_unknown` / `control.permission_denied`) answers unchanged.
- **Describe itself fails** → version 1, and the call itself meets and reports whatever stopped it; a caller never receives describe's error for a call it did not make.
- The success result now carries `capability_version`, the version that answered, so a caller that left the choice to the instance can read it (the tool declares no output schema; the field is additive).
- The named tools (`quantick_get_snapshot` and the rest) stay pinned at v1: their embedded input/output schemas are the committed v1 documents.

## Tests

- `crates/mcp/tests/fake_gateway.rs` — `an_omitted_version_reaches_the_newest_registered_one_and_an_explicit_one_is_kept`, through the real discovery, `LocalClient`, `LocalLink` and `McpServer`: a capability registered at v1 and v2 invoked without a version reaches v2 and reports `capability_version: 2`; `capability_version: 1` reaches v1; the v1-only `snapshot.read` answers at v1 without one; an explicit v3 is `control.capability_unknown`. The fake gateway now keys its answers on `(id, version)` like the real registry. The existing no-version `paper.order.place` assertion (unregistered ID → `control.capability_unknown`) stays green.
[...]

PR #419, opened 2026-09-12
perf(app): prove hot-path work is independent of session length and bound the forming fold

milocaetano:
## Summary

Prove, with counted work on dense short- and long-history fixtures, that one live pane's per-print, per-depth-update and per-frame work does not grow with session length; bound the indicator worker's forming-run fold without changing an output byte; and keep both as a regression check in the ordinary test step.

**Tier:** high. Campaign child of #367 (Q4); closes on integration: #380. Base: `campaign/lean-a-plus` (merge base `2452e577`).

Evidence page: [`docs/quality/session-length.md`](docs/quality/session-length.md); raw outputs under `docs/quality/session-length/`.

## What changed, by rate

| Change | Files | Rate |
| --- | --- | --- |
| The forming-run fold carries its state: a running bar and a checkpoint every 64 prints; a walk folds forward between rungs and jumps to a checkpoint, so it folds at most `min(len, 63 r)` prints (4,032 at the 64-rung ceiling) whatever the forming bar's length, never more than the old single pass. Lives in the engine beside `Bar::extend` (after the AI review) | `crates/engine/src/forming_run.rs` (new), `crates/engine/src/lib.rs` (one line), `indicator_worker.rs` | per drained batch (≈ per frame), indicator worker; per print: one fold, one checkpoint clone per 64 |
| The heatmap's live projection cut the retained prints at its seam by walking all of them every frame (+28 % at the envelope's edge, bounded only by the 100,000-print cap); it now bisects a running-newest-timestamp index kept beside the prints, so a late print can neither hide one nor turn the bisection off, and the projection is unchanged. The health count reads the deque's length | `crates/orderflow/src/history.rs`, `projection/tiers.rs`, `engine.rs` (one line) | per frame, book worker: O(retained) → O(log retained + kept) |
| The tape is reserved at load to twice what it holds, so the first live print after a backfill or a history prepend no longer copies the whole session (≈ 95 MB for a recovered 1.7 M-print day). A pane that never goes live keeps the unused half committed (never touched) | `state.rs` | rare (load); removes a per-frame outlier |
| Heap work counted per thread in the test binary only | `work_meter.rs` (new, `cfg(test)`), `main.rs` (registration) | test-only |
| The regression check | `app/tests/session_length_tests.rs` (new), one `mod` line in `app/tests/mod.rs`; the worker's test probe reports its folds and heap; sessions interleaved, CPU time the median lap | test-only |
[...]

Issue #418, opened 2026-09-12
test(app): fix the load-sensitive tests the contention step lists, and the Null capture_revision under load


milocaetano:
## Context

Campaign child of https://github.com/milocaetano/quantick/issues/367 (task key Q10). Base: `campaign/lean-a-plus`. Mission tier: high (one item may be a product race in evidence capture).

Q9 (PR #414, issue #407) added a CI step that runs the app test binary as 24 copies on 4 cores. It found eight pre-existing load-sensitive tests, filed one issue each and put them on a known-issue skip list that applies to the contention copies only: #408, #409, #410, #411, #413, #415, #416, #417. Decision D19 on #367 holds #414's merge until those are fixed and the step shows 10 consecutive green runs at the campaign tip.

Three of them (#413, #416, #417) share one symptom: an evidence bundle's screenshot or snapshot `capture_revision` is Null under load. That may be a product race in the evidence path (data honesty: evidence must say which revision it shows), not only test timing.

## Scope

- #413, #416, #417 first: find whether a bundle can be published with a Null `capture_revision` in production under load (the capture/verification path in `crates/app/src/control/evidence*`); if yes, fix the product path so the revision is always set or the bundle states it is unavailable (typed, not silently Null), with a regression test through the real gateway; if no, fix the tests.
- #408, #409, #410, #411, #415: force the interleaving each assertion depends on (the shape #382, #400, #404 used). Weakening an assertion is not a fix.
- Each fix removes that test's line from Q9's skip list (the list lives on #414's branch until #414 merges; coordinate: if #414 is not merged yet, record the removals in the PR body and the coordinator applies them to #414).

## Acceptance criteria

- [ ] A1: the Null `capture_revision` question answered with evidence (product race or test timing), and fixed at the right layer, with a real-gateway regression test if product.
- [ ] A2: each of the eight tests passes 24-copies-on-4-cores contention runs with 0 failures over at least 10 runs (the harness `tools/ci/contention.sh` from #414 with `CONTENTION_ROUNDS=10`, or its local equivalent), before/after table in the PR.
- [ ] A3: no assertion weakened; any production change reviewed under the full gates.
- [ ] A4: merged into `campaign/lean-a-plus` through a PR whose base is exactly that branch, with the merge read back; #408-#417 closed with integration evidence.

## Campaign assignment

[...]

PR #426, opened 2026-09-13
fix(control): keep a parked capture's image under load, and force what eight load-sensitive tests assert

### LINKED ISSUE #411: test(app): the duplicate-request-id parked-wait test loses its Success under load
## What fails

`app::tests::control_plane_tests::gateway_rejects_a_duplicate_request_id_while_a_wait_is_parked` fails under CPU contention:

```text
panicked at crates/app/src/app/tests/control_plane_tests.rs:2702:5:
assertion failed: matches!(client.read().unwrap().outcome,
    quantick_control::wire::ResponseOutcome::Success { .. })
```

The parked wait's own call does not come back `Success` on a loaded machine. Mechanism not yet read; likely the same frame-budget or request-deadline class as #364/#385.

## Reproduced

The harness is `tools/ci/contention.sh` from #407 (campaign #367, child Q9): N concurrent copies of the whole `quantick-app` test binary, each pinned with `taskset` to K cores and left at libtest's default `--test-threads` (= K), on the 4-vCPU `ubuntu-latest` runner. The test sources are identical at the campaign tip `52e35a1b` and at `f36cc022` (the only app diff between them is `layers_tests.rs`, from #404).

| Revision | Config | Copies failing on this test | Run |
| --- | --- | ---: | --- |
| `52e35a1b` | 24 copies on 4 cores | 1 of 24 (copy 3) | https://github.com/milocaetano/quantick/actions/runs/34708701953 |
| `52e35a1b` (local Windows, 12 cores) | 16 copies, affinity `0xF` | 1 of 16 | local, not CI |

Found by campaign #367 child Q9 (#407); a fix PR must remove the test's line from the contention skip list.

Out of scope for #407, which does not edit test sources: the contention step skips this test (in its copies only; the normal `Test` step still runs it) until the fix lands.

milocaetano:
## Summary

Fix the eight load-sensitive tests the CPU-contention step found, after first establishing that the three Null `capture_revision` failures come from a product race in the screenshot arc.

**Tier:** `high` (set by the coordinator: one item was a possible product race in evidence capture). Campaign child of #367 (Q10); closes on integration: #418, #408, #409, #410, #411, #413, #415, #416, #417.

## D2 finding: a product race, but never a Null revision on the wire

**No bundle can carry a screenshot without its revision.** `EvidenceScreenshot::capture_revision` is a non-optional `WireU64` (`control/evidence.rs`), set from the snapshot's own revision in `encode_screenshot` (`control/evidence/image.rs`). The manifest's `screenshot` is `Option<_>` with `skip_serializing_if = "Option::is_none"`. The `Null` in #413, #416 and #417 is `serde_json::Value` indexing into an **absent** `screenshot` object, not a descriptor missing its field. #413's own failure shows it: line 91 (`descriptor == manifest["screenshot"]`) passed because both were absent, and line 92 then compared `Null` with `"1"`.

**Why the image was absent: a delivered frame was dropped.** In `ControlAccess::begin_frame` (`control/gateway.rs`) the frame harvests the image, then `serve_awaiting_screenshot` (`control/gateway/screenshot.rs`) serves the parked captures, and at the end of the frame the unclaimed image is dropped (`self.screenshot = None`, "a rasterised frame is worth exactly one frame"). `serve_awaiting_screenshot` held a waiter back as soon as `elapsed_us_since(frame_started) > CONTROL_UI_BUDGET_US` (250 us), **before** the first waiter too. Under load the frame prelude alone (lifecycle and status polls, semantic events, the harvest) spends 250 us, so the waiter was held back and its image dropped. The capture then re-armed, waited out `deadline - CONTROL_SCREENSHOT_GRACE_MS`, and was answered with `coverage.not_captured: {subject: "screenshot", reason: "frame_not_delivered"}`. That label is false: the frame had been delivered. On a machine that spends the budget on *every* frame it is worse: the waiter is held back on every frame, never reaches its give-up, and the response worker answers its deadline with a bare `control.timeout` and no bundle at all.

Captured, not inferred: one instrumented copy under local contention printed the failing bundle's coverage:

```text
Q10DIAG coverage=[..., {"reason":"frame_not_delivered","subject":"screenshot"}, ...]
```

(the other 194 instrumented captures carried `screenshot.state_skew`, the gap an image-bearing bundle always lists).

**The fix is in the product, with no wire-shape change.** Two lines in `serve_awaiting_screenshot`:

1. A waiter is held back on time only while the frame holds no image: `out_of_time = elapsed > budget && self.screenshot.is_none()`. The count ceiling (`CONTROL_UI_MAX_REQUESTS_PER_FRAME`) still applies, and the first waiter is always under it.
2. A waiter that *is* held back re-arms the rasterise (`arm_screenshot`) instead of only requesting a repaint. Step 0's second round found this: once the first capture took the image, the captures parked behind it were held back with nothing asking the window for another frame, so they waited on an image nothing had requested.

The three tests pass without changing their assertions.

**Pre-existing, not changed:** a capture that reaches its give-up point (`deadline - CONTROL_SCREENSHOT_GRACE_MS`) with no image, on a machine that spends the budget on every frame, is still held back on time and ends in a bare `control.timeout` rather than `frame_not_delivered`. Exempting the give-up as well would be new behaviour with no deterministic test (it depends on the 250 ms grace window), so it is left as it was and named here.

**Real-gateway regression test.** `control_plane_tests::a_parked_capture_claims_its_image_in_a_frame_whose_budget_is_already_spent` parks a capture through a real socket, hands over the image, and runs one frame whose budget is already spent (`ControlAccess::begin_frame_over_budget_for_test`, a `#[cfg(test)]` seam over the production `begin_frame_since`). It asserts the capture is served in that frame, that its image is stamped with the capture's revision, and that the coverage does not say `frame_not_delivered`. A second test, `captures_parked_behind_the_first_ask_for_the_next_frame_when_the_budget_is_spent`, parks two captures, spends the budget on two frames, and asserts that the second capture re-arms after the first takes the image, then claims the next image (answers matched by request ID, since two response workers encode them). Without line 2 it fails at `the capture still parked asks the window for the next frame`. Before the fix, the first test fails deterministically:

```text
assertion `left == right` failed: the capture is served in the frame that holds its image, whatever that frame has spent
  left: 1
 right: 0
```

**A second, test-side exposure in the same three tests.** The park waits spun at most 600 back-to-back frames (`PARK_WAIT_FRAMES`) for the request to cross the socket and park. On a loaded machine they ran out first (`the capture did not park for an image within 600 frames`, `mod.rs:2442`, seen locally for #416). They are now one time-bounded helper, `run_frames_until_capture_parks`, with a 1 ms sleep between frames so the reader thread gets a core. The bound is `GATEWAY_TEST_WAIT`, as in #382 and #400.

## The other five (D3)

| Issue | Test | Root cause (captured) | What the test now forces |
| --- | --- | --- | --- |
| #408 | `gateway_client_reads_the_running_application_and_wrong_tokens_fail_closed` | The chart module's revision key is the whole chart projection. Between the two observer reads `timeline_revision` went `3 → 4` (instrumented diff, nothing else changed): the order-flow worker published the capture bucket late, and the next frame's `set_footprint_group` refolded the footprints. | `settle_flow_pane`: flush the order-flow worker, run the frame that adopts it, and repeat until a round changes neither the flow pane's timeline revision nor its viewport projection (at most 8 rounds, else it panics). The observer-read assertion is unchanged. |
| #409 | `attaching_a_script_and_detaching_it_leaves_the_pane_as_it_was` | 200 back-to-back frames passed before the indicator worker built the script. | The suite's existing `settle_indicators`: flush every pane's indicator worker, then apply its events (the frame loop's two steps). Used after the attach and after the detach. Both assertions unchanged. |
| #415 | `an_operator_cannot_detach_the_traders_own_indicator` | Same as #409. | `settle_indicators`. It also asserts the trader's indicator is on the pane **before** the refused detach, so "still on the pane" can no longer pass or fail on timing. |
| #410 | `a_keyed_call_that_expired_before_the_application_saw_it_leaves_its_key_free` | One 50 ms request window served both the call that must expire and the retry that must not. Under load the retry also expired before dispatch (`control.timeout`, "request expired before application-thread dispatch"). | The window is 750 ms (the precedent in `retry_readback_tests`). The first call still expires, since nothing serves it until its refusal is read. Each retry is served the moment it is queued, through `serve_queued_for_test` (the production `execute_on_ui`), paced by `pause_one_request_interval` (one interval of `CONTROL_CLIENT_RATE_PER_SECOND`, 50 ms today) so the connection's rate limit does not answer instead (a first cut at 10 ms met `control.backpressure`). The expired call's refusal is read with `read_with_extra_patience`. The success and "exactly one layout" assertions are unchanged. |
| #411 | `gateway_rejects_a_duplicate_request_id_while_a_wait_is_parked` | Instrumented: the final reuse came back `control.invalid_request` / "request_id is already in flight on this connection" (15 of 65 runs). Every terminal path in `gateway/server.rs` writes the answer **before** `slots.forget(request_id)`, so an immediate reuse can race the release. Filed as **#425**. | The reuse is retried on exactly `control.invalid_request`, paced by the same `pause_one_request_interval` (slower than the connection's rate limit, so the limiter's refusal cannot end the loop; step 0 caught a first cut at 5 ms, and ai-review asked for the pause to be derived from the limit), within `GATEWAY_TEST_WAIT`, then must succeed. An ID the gateway never released still fails. The product ordering is out of D3's scope (see *human decision* below). |

**A ninth test, by the same mechanism.** `an_operator_detaching_its_own_slot_leaves_the_traders_slot_of_the_same_number`, which sits beside #415 and uses the same 200-frame spin, failed this mission's own `cargo test --workspace` while contention runs shared the host, and one local contention round. It gets the same `settle_indicators`. That is a stated detour (mission S4): the verification loop could not be green without it.

No assertion was removed or loosened. Every waiting loop is bounded by time and fails with a message that names what it waited for.

## Contention proof (D4)

Local equivalent of `tools/ci/contention.sh` (Linux-only `taskset`): a PowerShell harness starts **24 concurrent copies of the whole `quantick-app` test binary**, each held to processor mask `0xF` (4 of this machine's 12 cores; children inherit the mask) with `--test-threads=4`, from `crates/app`. This is the shape Q9 used for its local reproductions. The host is shared (an Edge WebView and other sessions were using CPU), so these runs are at least as contended as CI, not less. *Before* is the base binary (`c55a4222`). *After* is the binary built from this branch. The branch was rebased onto `a3c962fa` (#419, #424) during the mission.

**The D4 proof: 24 copies of the whole binary on mask `0xF`, ten rounds at the final code (`01a367b3`, base `a3c962fa`).** 240 copies ran each test. The tree differs from the head `5a9c5c26` only by a 1 ms sleep inside the new `captures_parked_behind_the_first…` test's park wait.

| Issue | Test | Before: CI (issue evidence, 24 × 4) | Before: local 0xF, 3 rounds at `c55a4222` | After: local 0xF, 10 rounds at `01a367b3` |
| #408 | `gateway_client_reads_…_fail_closed` | 1–3 of 24 per run, in 4 runs | 0 / 72 | **0 / 240** |
| #409 | `attaching_a_script_and_detaching_it_…` | 1 of 16 | 0 / 72 | **0 / 240** |
| #410 | `a_keyed_call_that_expired_…_key_free` | 1 of 24, twice | **2 / 72** (1, 1, 0) | **0 / 240** |
| #411 | `gateway_rejects_a_duplicate_request_id_…` | 1 of 24 | **15 / 72** (1, 5, 9) | **0 / 240** |
| #413 | `synthetic_fractional_geometry_round_trips_…` | 1 of 24 | 0 / 72 | **0 / 240** |
| #415 | `an_operator_cannot_detach_the_traders_own_indicator` | 1 of 24 | 0 / 72 | **0 / 240** |
| #416 | `a_bundle_with_a_screenshot_maps_…` | 1 of 24 | 0 / 72 | **0 / 240** |
| #417 | `a_capture_that_wants_an_image_waits_…` | 1 of 24 | 0 / 72 | **0 / 240** |
| new | `a_parked_capture_claims_its_image_…` | n/a | n/a | **0 / 240** |
| new | `captures_parked_behind_the_first_…` | n/a | n/a | **0 / 240** |
| S4 | `an_operator_detaching_its_own_slot_…` | not listed | 0 / 72 (plus 1 in 10 rounds at `efbada89` before its fix, and once in this mission's own `cargo test --workspace`) | **0 / 240** |

An earlier ten rounds at `efbada89` (the first fix, base `c55a4222`, binary built with `-p`) also had 0 / 240 for each of the eight. Their failures elsewhere are counted below.

**A harsher local configuration, to reproduce what 0xF rarely does:** the eight tests only, 24 copies on mask `0x3` (2 cores).

| Build | Rounds | What failed |
| --- | ---: | --- |
| before (`c55a4222`) | 3 | #411 at its assertion 17 times; #416 at the frame-counted park wait (`the capture did not park for an image within 600 frames`) twice; #409 at `:3135` once; #415 at `:2910` once. The instrumented copy of the base (3 more rounds) also showed #408 at `:941` (`timeline_revision` 3 → 4) twice, #413 at `:93` (`Null` vs `"1"`, coverage `frame_not_delivered`) once, and #411's reuse refused as `already in flight` 15 times. |
| after (`01a367b3`) | 6 | 3 copies in one round, all at the connect handshake (`discover_in(...).select(None).unwrap()` → `control.instance_gone`, the client's 2 s `CONTROL_HANDSHAKE_TIMEOUT_MS`), none at any of the issues' assertions. That exposure is shared by every gateway test on 24 copies over 2 cores. It did not appear in any 0xF round. |

**Not caused by this branch:** in the same 0xF runs, other tests failed in both builds. None is touched by this branch, and each is filed:

| Test | Before, 3 rounds | After, 10 rounds at `01a367b3` | Issue |
| --- | --- | --- | --- |
| `gateway_a_client_that_never_reads_does_not_stall_another` | 0, 9, 1 | 1, 11, 6, 8, 10, 6, 5, 6, 6, 8 | #427 |
| `a_keyed_action_held_past_its_deadline_…_by_readback` | 0, 2, 3 | 0, 8, 0, 2, 0, 0, 0, 0, 9, 1 | #428 |
| `every_reachable_optional_row_replays_a_dropped_answer_and_begins_once` | 0, 0, 0 | 0, 0, 0, 0, 0, 0, 0, 5, 8, 1 | #432 |
| `gateway_request_timeout_is_structured_and_late_ui_work_is_discarded` | 1, 0, 0 | 0 (1 in the `efbada89` rounds) | #430 |
| `live_envelope_tests::burst::inside_the_envelope_every_print_arrives_and_no_queue_fills` | not measured | 13 / 24 in one probe round at `efbada89`; skipped afterwards so the rounds could measure the eight | #429 |

So **the contention step's 10-green landing condition (D19) is not met by this PR alone**: a whole-binary round was red in most rounds locally, on these tests. On a shared 12-core host they fail more often than on the CI runner, where Q9's calibration did not list them.

## Skip-list lines this allows removing

On #414's branch, `tools/ci/contention-known-issues.txt`: **all eight entries, lines 11 to 18** (#408, #409, #410, #411, #413, #415, #416, #417). Once they are gone the file holds only its header. The coordinator applies this to #414.

## Findings filed on the way

- **#425**: the gateway writes an answer before releasing its request ID (the #411 product race; see *human decision*).
- **#427, #428, #429, #430, #432**: load-sensitive tests outside the eight (table above). Each needs a fix, or a skip-list line citing its issue, before the contention step can hold ten consecutive green runs.

## Performance impact

- `serve_awaiting_screenshot` (product): **rare**. It returns before the changed line on every frame with no parked capture. With one parked, the change costs one `Option::is_none` per waiter. The only behavioural difference: a frame that holds an image can run one capture past the 250 us budget, a capture that would otherwise have run on a later frame, or never.
- `begin_frame` → `begin_frame_since(Instant::now())`: per frame, an extra call with identical work.
- Everything else is `#[cfg(test)]`.

## Human decision

- **#425: release a request ID before writing its answer.** The gateway's terminal paths write, then forget. A client that reuses an ID right after reading its answer can be refused with a **non-retryable** `control.invalid_request`. That contradicts contract §5.2 ("while the first request is in flight"). The fix is a reorder on each terminal path in `gateway/server.rs`. It is outside D2's evidence-path authorization, so this PR fixes #411 at the test layer and the product decision is the coordinator's or trader's. Once #425 lands, #411's wait can go.

## Verification loop

Run locally at ``5a9c5c26` (the head)`, each command on its own:

- [x] `env -u QUANTICK_BUBBLES cargo test --workspace` (`quantick-app` binary: 2060 passed, 0 failed, 9 ignored. The first run at this head hit #428's test once while the contention rounds shared the host; the rerun was green.)
- [x] `cargo test -p quantick-guards`

Plus the contention runs above (local), and required CI at the final head.

## Reviews

- `arch-review`: https://github.com/milocaetano/quantick/pull/426#issuecomment-5651453581. Step 0 was `code-review` at medium, three rounds, 3 findings (one Blocker in the product path), all fixed. The shape pass is clean; two Consider items are recorded there.
- `ai-review`: https://github.com/milocaetano/quantick/pull/426#issuecomment-5651287033 (full), then https://github.com/milocaetano/quantick/pull/426#issuecomment-5651457726 (follow-up at `5a9c5c26`). One WEAK (retry pacing) was fixed and its thread resolved; 0 open threads.
- `delivery-review`: FAIL at `5a9c5c26` on one traceability gap (DR-Q10-1: the write scope was not ledgered; https://github.com/milocaetano/quantick/pull/426#issuecomment-5651844726). `60ef4053` fixed it by adding G5 and rewording S3. The follow-up **PASS** is at https://github.com/milocaetano/quantick/pull/426#issuecomment-5651887292.
- Follow-ups at `60ef4053`: arch-review https://github.com/milocaetano/quantick/pull/426#issuecomment-5651849401, ai-review https://github.com/milocaetano/quantick/pull/426#issuecomment-5651852376.

CI at the head `60ef4053`: https://github.com/milocaetano/quantick/actions/runs/34744430953 (`ci` and `windows` pass). `60ef4053` changes only the mission archive's prose. The local checks were run at `5a9c5c26`.

## Notes for the reviewer

- `crates/app/src/app/tests/mod.rs` is the shared helper file. The only edit there is the park-wait helper that replaces `PARK_WAIT_FRAMES`, because `capture_with_screenshot` (used by #416) lives there. No open campaign PR touches that file.
- `control/gateway.rs` is outside this child's owned file list (mission G5). The `#[cfg(test)]` seam `begin_frame_over_budget_for_test` sits there, beside the other `*_for_test` gateway seams, because the frame it runs is `begin_frame`'s. The only production line is the split of `begin_frame` into a private `begin_frame_since(Instant::now())`, which does identical work. No other open campaign PR touches the file. The coordinator decides whether to accept it.

### DIFF
--- a/crates/app/src/app/tests/control_plane_tests.rs
+++ b/crates/app/src/app/tests/control_plane_tests.rs
@@ -2752,7 +2752,9 @@ fn gateway_rejects_a_duplicate_request_id_while_a_wait_is_parked() {
             std::time::Instant::now() < deadline,
             "the answered wait's request ID was never released"
         );
-        std::thread::sleep(std::time::Duration::from_millis(5));
+        // Slower than the connection's request rate limit, or the refusal
+        // that comes back is the limiter's rather than the duplicate check's.
+        std::thread::sleep(std::time::Duration::from_millis(50));
     };
     assert!(
         matches!(
@@ -2931,18 +2933,6 @@ fn an_operator_cannot_remove_an_object_the_trader_drew() {
     assert_eq!(result["removed"], true);
 }
 
-/// Let every pane's indicator worker finish what an attach or a detach sent
-/// it, then run the frame that applies the result: the application loop's own
-/// order, with the worker's half made deterministic. A fixed count of frames
-/// is not a wait; on a loaded machine they all pass before the worker thread
-/// has run at all (#409, #415, and the slot-number test below).
-fn flush_indicators_then_frame(app: &mut QuantickApp, ctx: &egui::Context) {
-    for pane in app.active_tab_mut().panes_mut() {
-        pane.indicator_worker.flush();
-    }
-    run_frame(app, ctx);
-}
-
 /// The tier's floor, again, on the surface the review found open: an
 /// operator detaches what an operator attached, and the trader's own
 /// indicator stays on the chart whatever slot id is named.
@@ -2963,7 +2953,8 @@ plot(close)
         .to_owned(),
         false,
     );
-    flush_indicators_then_frame(&mut app, &ctx);
+    // Settled, not waited out over a count of frames (#415).
+    settle_indicators(&mut app);
     assert_eq!(
         indicator_kinds(&app).len(),
         1,
@@ -3015,14 +3006,15 @@ plot(close)
     // The trader's own, on the first tab.
     let (traders_tab, _, traders_slot) =
         app.attach_script_indicator("the trader's".to_owned(), SCRIPT.to_owned(), false);
-    flush_indicators_then_frame(&mut app, &ctx);
+    // Settled, not waited out over a count of frames, as in #415.
+    settle_indicators(&mut app);
 
     // The second chart, whose slot numbering starts over from zero.
     app.cycle_tab(1);
     run_frame(&mut app, &ctx);
     let (operators_tab, _, operators_slot) =
         app.attach_script_indicator("an assistant's".to_owned(), SCRIPT.to_owned(), true);
-    flush_indicators_then_frame(&mut app, &ctx);
+    settle_indicators(&mut app);
     assert_ne!(traders_tab, operators_tab, "two charts, not one");
     assert_eq!(
         traders_slot.0, operators_slot.0,
@@ -3185,9 +3177,10 @@ fn attaching_a_script_and_detaching_it_leaves_the_pane_as_it_was() {
         )
         .expect("a script that compiles is attached");
     let slot_id = attached["slot_id"].as_str().unwrap().to_owned();
-    // The worker builds off the application thread; frames apply what it
-    // produced, exactly as the library's own click path is served.
-    flush_indicators_then_frame(&mut app, &ctx);
+    // The worker builds off the application thread. Settled rather than
+    // waited out over a count of frames: on a loaded machine 200 frames pass
+    // before the worker thread has run at all (#409).
+    settle_indicators(&mut app);
     assert_eq!(
         indicator_kinds(&app).len(),
         before.len() + 1,
@@ -3201,7 +3194,7 @@ fn attaching_a_script_and_detaching_it_leaves_the_pane_as_it_was() {
         )
         .expect("what an operator attached, an operator detaches");
     assert_eq!(detached["detached"], true);
-    flush_indicators_then_frame(&mut app, &ctx);
+    settle_indicators(&mut app);
     assert_eq!(
         indicator_kinds(&app),
         before,
--- a/crates/app/src/app/tests/control_plane_tests.rs
+++ b/crates/app/src/app/tests/control_plane_tests.rs
@@ -2724,9 +2724,10 @@ fn gateway_rejects_a_duplicate_request_id_while_a_wait_is_parked() {
     // Free again. The gateway writes the wait's answer before it releases
     // the ID, so a client that reuses the ID the instant it reads that answer
     // can still meet the duplicate refusal: on a loaded machine the wait
-    // thread is preempted between the two (#411). The reuse is retried on
-    // exactly that refusal, inside the shared test budget, and must then
-    // succeed; an ID the gateway never released fails here.
+    // thread is preempted between the two (#411; the ordering itself is
+    // #425). The reuse is retried on exactly that refusal, inside the shared
+    // test budget, and must then succeed; an ID the gateway never released
+    // fails here.
     let deadline = std::time::Instant::now() + GATEWAY_TEST_WAIT;
     let reused = loop {
         client
@@ -2934,7 +2935,7 @@ fn an_operator_cannot_remove_an_object_the_trader_drew() {
 /// it, then run the frame that applies the result: the application loop's own
 /// order, with the worker's half made deterministic. A fixed count of frames
 /// is not a wait; on a loaded machine they all pass before the worker thread
-/// has run at all (#409, #415).
+/// has run at all (#409, #415, and the slot-number test below).
 fn flush_indicators_then_frame(app: &mut QuantickApp, ctx: &egui::Context) {
     for pane in app.active_tab_mut().panes_mut() {
         pane.indicator_worker.flush();
@@ -3014,24 +3015,14 @@ plot(close)
     // The trader's own, on the first tab.
     let (traders_tab, _, traders_slot) =
         app.attach_script_indicator("the trader's".to_owned(), SCRIPT.to_owned(), false);
-    for _ in 0..200 {
-        run_frame(&mut app, &ctx);
-        if !indicator_kinds(&app).is_empty() {
-            break;
-        }
-    }
+    flush_indicators_then_frame(&mut app, &ctx);
 
     // The second chart, whose slot numbering starts over from zero.
     app.cycle_tab(1);
     run_frame(&mut app, &ctx);
     let (operators_tab, _, operators_slot) =
         app.attach_script_indicator("an assistant's".to_owned(), SCRIPT.to_owned(), true);
-    for _ in 0..200 {
-        run_frame(&mut app, &ctx);
-        if !indicator_kinds(&app).is_empty() {
-            break;
-        }
-    }
+    flush_indicators_then_frame(&mut app, &ctx);
     assert_ne!(traders_tab, operators_tab, "two charts, not one");
     assert_eq!(
         traders_slot.0, operators_slot.0,
--- a/crates/app/src/app/tests/control_plane_tests.rs
+++ b/crates/app/src/app/tests/control_plane_tests.rs
@@ -818,6 +818,30 @@ fn refused_handshake_code(
     );
 }
 
+/// Let the flow pane's order-flow worker publish everything it was sent, and
+/// run the frame that adopts it, until one such round changes nothing the
+/// chart projection reads from the pane. The frame can hand the worker new
+/// input (a price grid, a projection request), so one round is not always
+/// enough; a bounded number is.
+fn settle_flow_pane(app: &mut QuantickApp, ctx: &egui::Context) {
+    let observed = |app: &QuantickApp| {
+        let pane = &app.active_tab().flow_pane;
+        (
+            pane.state.timeline_revision(),
+            crate::control::chart::viewport_snapshot(pane),
+        )
+    };
+    for _ in 0..8 {
+        let before = observed(app);
+        app.active_tab_mut().tape_mut().flush_for_test();
+        run_frame(app, ctx);
+        if observed(app) == before {
+            return;
+        }
+    }
+    panic!("the flow pane kept changing after eight worker flushes and frames");
+}
+
 #[test]
 fn gateway_client_reads_the_running_application_and_wrong_tokens_fail_closed() {
     use quantick_control::{
@@ -830,6 +854,12 @@ fn gateway_client_reads_the_running_application_and_wrong_tokens_fail_closed() {
     let (mut app, _commands) = app_with_history(12);
     let directory = gateway_test_directory("read");
     let _descriptor_path = enable_test_gateway(&mut app, &ctx, &directory, 8);
+    // The observer-read claim below compares two reads, so nothing but those
+    // reads may move the chart between them. The order-flow worker is still
+    // publishing what the backfill sent it, and the frame that adopts its
+    // capture bucket refolds the footprints, which moves the chart's timeline
+    // revision (#408).
+    settle_flow_pane(&mut app, &ctx);
     let discovery =
         quantick_control_local::client::discover_in(&directory, &gateway_test_options()).unwrap();
     assert!(discovery.issues.is_empty());
@@ -2691,18 +2721,46 @@ fn gateway_rejects_a_duplicate_request_id_while_a_wait_is_parked() {
     let page = reply.expect("the parked wait was answered");
     assert_eq!(page.request_id, request_id);
     assert_eq!(success_result(&page)["timed_out"], true);
-    client
-        .send_with_request_id(
-            request_id.clone(),
-            crate::control::DESCRIBE_CAPABILITY_ID,
-            1,
-            serde_json::json!({}),
-        )
-        .unwrap();
-    assert!(matches!(
-        client.read().unwrap().outcome,
-        quantick_control::wire::ResponseOutcome::Success { .. }
-    ));
+    // Free again. The gateway writes the wait's answer before it releases
+    // the ID, so a client that reuses the ID the instant it reads that answer
+    // can still meet the duplicate refusal: on a loaded machine the wait
+    // thread is preempted between the two (#411). The reuse is retried on
+    // exactly that refusal, inside the shared test budget, and must then
+    // succeed; an ID the gateway never released fails here.
+    let deadline = std::time::Instant::now() + GATEWAY_TEST_WAIT;
+    let reused = loop {
+        client
+            .send_with_request_id(
+                request_id.clone(),
+                crate::control::DESCRIBE_CAPABILITY_ID,
+                1,
+                serde_json::json!({}),
+            )
+            .unwrap();
+        let reply = client.read().unwrap();
+        assert_eq!(reply.request_id, request_id);
+        let still_held = matches!(
+            &reply.outcome,
+            quantick_control::wire::ResponseOutcome::Failure { error }
+                if error.code.as_str() == codes::INVALID_REQUEST
+        );
+        if !still_held {
+            break reply;
+        }
+        assert!(
+            std::time::Instant::now() < deadline,
+            "the answered wait's request ID was never released"
+        );
+        std::thread::sleep(std::time::Duration::from_millis(5));
+    };
+    assert!(
+        matches!(
+            reused.outcome,
+            quantick_control::wire::ResponseOutcome::Success { .. }
+        ),
+        "the ID is free once the wait is answered: {:?}",
+        reused.outcome
+    );
 
     disable_test_gateway(&mut app, &ctx);
     std::fs::remove_dir_all(directory).unwrap();
@@ -2872,6 +2930,18 @@ fn an_operator_cannot_remove_an_object_the_trader_drew() {
     assert_eq!(result["removed"], true);
 }
 
+/// Let every pane's indicator worker finish what an attach or a detach sent
+/// it, then run the frame that applies the result: the application loop's own
+/// order, with the worker's half made deterministic. A fixed count of frames
+/// is not a wait; on a loaded machine they all pass before the worker thread
+/// has run at all (#409, #415).
+fn flush_indicators_then_frame(app: &mut QuantickApp, ctx: &egui::Context) {
+    for pane in app.active_tab_mut().panes_mut() {
+        pane.indicator_worker.flush();
+    }
+    run_frame(app, ctx);
+}
+
 /// The tier's floor, again, on the surface the review found open: an
 /// operator detaches what an operator attached, and the trader's own
 /// indicator stays on the chart whatever slot id is named.
@@ -2892,12 +2962,12 @@ plot(close)
         .to_owned(),
         false,
     );
-    for _ in 0..200 {
-        run_frame(&mut app, &ctx);
-        if !indicator_kinds(&app).is_empty() {
-            break;
-        }
-    }
+    flush_indicators_then_frame(&mut app, &ctx);
+    assert_eq!(
+        indicator_kinds(&app).len(),
+        1,
+        "the trader's indicator is on the pane"
+    );
     let refused = app
         .control_action(
             "indicator.script.detach",
@@ -3126,12 +3196,7 @@ fn attaching_a_script_and_detaching_it_leaves_the_pane_as_it_was() {
     let slot_id = attached["slot_id"].as_str().unwrap().to_owned();
     // The worker builds off the application thread; frames apply what it
     // produced, exactly as the library's own click path is served.
-    for _ in 0..200 {
-        run_frame(&mut app, &ctx);
-        if indicator_kinds(&app).len() > before.len() {
-            break;
-        }
-    }
+    flush_indicators_then_frame(&mut app, &ctx);
     assert_eq!(
         indicator_kinds(&app).len(),
         before.len() + 1,
@@ -3145,12 +3210,7 @@ fn attaching_a_script_and_detaching_it_leaves_the_pane_as_it_was() {
         )
         .expect("what an operator attached, an operator detaches");
     assert_eq!(detached["detached"], true);
-    for _ in 0..200 {
-        run_frame(&mut app, &ctx);
-        if indicator_kinds(&app).len() == before.len() {
-            break;
-        }
-    }
+    flush_indicators_then_frame(&mut app, &ctx);
     assert_eq!(
         indicator_kinds(&app),
         before,
@@ -4594,19 +4654,7 @@ fn a_capture_that_wants_an_image_waits_for_the_frame_instead_of_answering_blind(
         .expect("the request is sent");
     // Frames pass and the capture does not answer: it is waiting for the
     // window, which a headless context never rasterises on its own.
-    let mut waited = 0;
-    for _ in 0..PARK_WAIT_FRAMES {
-        run_frame(&mut app, &ctx);
-        waited = app
-            .control
-            .control_access
-            .as_ref()
-            .expect("control access is installed")
-            .awaiting_screenshot_for_test();
-        if waited > 0 {
-            break;
-        }
-    }
+    let waited = run_frames_until_capture_parks(&mut app, &ctx);
     assert_eq!(waited, 1, "the capture parked instead of answering blind");
 
     let mut access = app
@@ -4636,6 +4684,70 @@ fn a_capture_that_wants_an_image_waits_for_the_frame_instead_of_answering_blind(
             .expect("control access is installed")
             .awaiting_screenshot_for_test(),
         0
+    );
+
+    disable_test_gateway(&mut app, &ctx);
+    std::fs::remove_dir_all(directory).unwrap();
+}
+
+/// An image can be claimed only in the frame that holds it: `begin_frame`
+/// drops an unclaimed one when the frame ends. A loaded machine can spend the
+/// frame's whole request budget before the parked captures are served, and a
+/// capture held back on that budget lost the picture it parked for, waited
+/// out its deadline, and reported `frame_not_delivered` for a frame that was
+/// delivered (#413, #416, #417). This spends the budget on purpose.
+#[test]
+fn a_parked_capture_claims_its_image_in_a_frame_whose_budget_is_already_spent() {
+    let ctx = egui::Context::default();
+    let (mut app, _commands) = app_with_history(6);
+    run_frame(&mut app, &ctx);
+    let directory = gateway_test_directory("evidence-image-over-budget");
+    grant_annotate_for_test(&mut app, "all-reads,observe.evidence,observe.screenshot");
+    enable_test_gateway(&mut app, &ctx, &directory, 4);
+    let mut client =
+        quantick_control_local::client::discover_in(&directory, &evidence_test_options())
+            .unwrap()
+            .select(None)
+            .unwrap();
+
+    let request_id = client
+        .send(
+            "evidence.capture",
+            serde_json::json!({ "scopes": ["scene.controls"], "screenshot": true }),
+        )
+        .expect("the request is sent");
+    assert_eq!(run_frames_until_capture_parks(&mut app, &ctx), 1);
+
+    let mut access = app
+        .control
+        .control_access
+        .take()
+        .expect("control access is installed");
+    access.publish_screenshot_for_test(&mut app, test_screenshot(320, 200));
+    access.begin_frame_over_budget_for_test(&mut app, &ctx);
+    let still_parked = access.awaiting_screenshot_for_test();
+    app.control.control_access = Some(access);
+    assert_eq!(
+        still_parked, 0,
+        "the capture is served in the frame that holds its image, whatever that frame has spent"
+    );
+
+    let response = client.read().expect("the gateway answered");
+    assert_eq!(response.request_id, request_id);
+    let manifest = success_result(&response);
+    assert_eq!(
+        manifest["screenshot"]["capture_revision"], manifest["capture_revision"],
+        "the image is stamped with the capture that parked for it"
+    );
+    assert!(
+        !manifest["coverage"]["not_captured"]
+            .as_array()
+            .expect("the coverage lists what is missing")
+            .contains(&serde_json::json!({
+                "subject": "screenshot",
+                "reason": "frame_not_delivered"
+            })),
+        "a delivered frame is never reported as undelivered"
     );
 
     disable_test_gateway(&mut app, &ctx);
@@ -5368,7 +5480,12 @@ fn a_keyed_call_that_expired_before_the_application_saw_it_leaves_its_key_free()
     run_frame(&mut app, &ctx);
     let directory = gateway_test_directory("idempotency-settle");
     grant_annotate_for_test(&mut app, "all-reads,cockpit,cockpit.layout");
-    enable_test_gateway_with_limits(&mut app, &ctx, &directory, 4, Duration::from_millis(50), 4);
+    // One window serves both calls: the first must outlive it, the retry must
+    // not. 50 ms was short enough that a loaded machine let the retry expire
+    // too (#410). The first call expires however long the window is, because
+    // nothing serves it until its refusal has been read.
+    let window = Duration::from_millis(750);
+    enable_test_gateway_with_limits(&mut app, &ctx, &directory, 4, window, 4);
     let mut client =
         quantick_control_local::client::discover_in(&directory, &cockpit_test_options())
             .unwrap()
@@ -5388,7 +5505,12 @@ fn a_keyed_call_that_expired_before_the_application_saw_it_leaves_its_key_free()
             key(),
         )
         .unwrap();
-    let expired = client.read().unwrap();
+    // The refusal is written when the response worker's own timer fires, and a
+    // loaded machine can wake that thread after the client's ordinary read
+    // timeout: waited for as long as the shared test budget allows.
+    let expired = client
+        .read_with_extra_patience(GATEWAY_TEST_WAIT.as_millis() as u64)
+        .unwrap();
     assert_eq!(expired.request_id, first);
     assert_eq!(response_error(&expired).code.as_str(), codes::TIMEOUT);
     assert!(
@@ -5407,30 +5529,56 @@ fn a_keyed_call_that_expired_before_the_application_saw_it_leaves_its_key_free()
 
     // The invited retry. `control.request_in_progress` while the settle window
     // is still open is the contract's own instruction to ask again, so this
-    // asks again rather than treating it as the answer.
-    let mut answered = None;
-    for attempt in 0..40 {
-        let response = remote_call_with_key(
-            &mut app,
-            &ctx,
-            &mut client,
-            &format!("retry-{attempt}"),
-            "layout.tab.create",
-            serde_json::json!({}),
-            "layout-key-1",
-        );
+    // asks again rather than treating it as the answer. Each retry is served
+    // the moment it is queued, through the same `execute_on_ui` a frame's
+    // drain calls, so its own deadline is never spent waiting for a frame
+    // whose budget the frame's other work used up.
+    let deadline = std::time::Instant::now() + GATEWAY_TEST_WAIT;
+    let mut attempt = 0;
+    let answered = loop {
+        attempt += 1;
+        let sent = client
+            .send_with_idempotency_key(
+                RequestId::new(format!("retry-{attempt}")).unwrap(),
+                "layout.tab.create",
+                1,
+                serde_json::json!({}),
+                key(),
+            )
+            .unwrap();
+        let response = loop {
+            let mut access = app
+                .control
+                .control_access
+                .take()
+                .expect("control access is installed");
+            access.serve_queued_for_test(&mut app);
+            app.control.control_access = Some(access);
+            if client.reply_pending(Duration::from_millis(5)) {
+                break client.read().unwrap();
+            }
+            assert!(
+                std::time::Instant::now() < deadline,
+                "no answer to retry-{attempt}"
+            );
+        };
+        assert_eq!(response.request_id, sent);
         let still_running = matches!(
             &response.outcome,
             quantick_control::wire::ResponseOutcome::Failure { error }
                 if error.code.as_str() == codes::REQUEST_IN_PROGRESS
         );
         if !still_running {
-            answered = Some(response);
-            break;
+            break response;
         }
-        std::thread::sleep(Duration::from_millis(10));
-    }
-    let answered = answered.expect("the settle window closes and the key comes back");
+        assert!(
+            std::time::Instant::now() < deadline,
+            "the settle window closes and the key comes back"
+        );
+        // Slower than the connection's request rate limit, or the refusal
+        // that comes back is the limiter's rather than the store's.
+        std::thread::sleep(Duration::from_millis(50));
+    };
 
     assert!(
         matches!(
--- a/.claude/GOAL-archive-contention-listed-tests.md
+++ b/.claude/GOAL-archive-contention-listed-tests.md
@@ -68,10 +68,11 @@ Campaign child of #367 (Q10). Base: `campaign/lean-a-plus`. Issue: #418.
   `an_operator_detaching_its_own_slot_leaves_the_traders_slot_of_the_same_number`
   (same module, 200 back-to-back frames after an attach), failed this
   mission's own `cargo test --workspace` while contention runs shared the
-  host, and one local contention round. It gets the same flush helper: a
-  detour stated here and in the PR, because the verification loop cannot be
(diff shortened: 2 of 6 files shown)


crates/app/src/app/tests/control_plane_tests.rs
```diff
@@ -2691,18 +2721,49 @@ fn gateway_rejects_a_duplicate_request_id_while_a_wait_is_parked() {
...
+        if !still_held {
+            break reply;
+        }
+        assert!(
+            std::time::Instant::now() < deadline,
+            "the answered wait's request ID was never released"
+        );
+        // Slower than the connection's request rate limit, or the refusal
+        // that comes back is the limiter's rather than the duplicate check's.
+        std::thread::sleep(std::time::Duration::from_millis(50));
```
milocaetano: **WEAK (dimension 4, agent-tested) - the retry pacing restates the client rate limit as a literal 50 ms.**

Both new retry loops (here, and the idempotency retry at `control_plane_tests.rs:5564`) sleep `from_millis(50)` and justify it as "slower than the connection's request rate limit". That holds only because `quantick_control::limits::CONTROL_CLIENT_RATE_PER_SECOND` is 20 today (one token per 50 ms, burst `CONTROL_CLIENT_BURST` = 40).

Breaking variant: lower `CONTROL_CLIENT_RATE_PER_SECOND` to 10. Under the contention these loops exist for, retries run faster than the bucket refills, so it empties after about 4 s of the 30 s `GATEWAY_TEST_WAIT`. The limiter's `BACKPRESSURE` then ends the duplicate-ID loop, which only retries `INVALID_REQUEST`, and the test fails on "the ID is free once the wait is answered". In the idempotency loop the same refusal is taken as `answered`. Either way it looks like a gateway defect, and it is only test pacing.

Fix: derive the pause from the constant, e.g. `Duration::from_millis(1_000 / u64::from(CONTROL_CLIENT_RATE_PER_SECOND) + 1)`, as one shared helper for both loops. The comment then stays true when the limit changes. This is test-only and has no product impact.

milocaetano:
## AI review - PR #426 (`fix/contention-listed-tests`) @ e9d3d1d8911509c4c130291e2cbba2af1290be37
Identity: HEAD `e9d3d1d8911509c4c130291e2cbba2af1290be37`; branch `fix/contention-listed-tests`; worktree `C:\src\quantick-worktrees\fix-contention-listed-tests` (clean); base ref `origin/campaign/lean-a-plus`; base tip `a3c962fa2d209fc8473927902cf3dfffcc68c5a2` (matches `git ls-remote`); review key `a39e27f57f1d27e1cdf1d479578bde42f9164919`
Scope: full diff (round one, `git diff origin/campaign/lean-a-plus...HEAD`, 6 files, +461/-107). 1 finding: `PRRT_kwDOTfuoRs6h2lif` (WEAK, dimension 4). 0 FAIL. Unresolved 1, accepted 0, fixed 0.

1. Modular:      PASS - The product fix is a single condition inside the one function that owns parked captures (`control/gateway/screenshot.rs:205-207`, `serve_awaiting_screenshot`). `begin_frame` now forwards to a private `begin_frame_since` (`control/gateway.rs:694-707`) and changes nothing else. The diff edits 5 pre-existing files. Blast radius: `app/tests/control_plane_tests.rs` (284 changed lines, all tests). Production lines are 14 in `screenshot.rs` and 30 in `gateway.rs`. Duplicated park-wait loops in three files are merged into one `run_frames_until_capture_parks` (`app/tests/mod.rs:2460`), so the test-side units get smaller. Minor note: the seam `begin_frame_over_budget_for_test` (`gateway.rs:1185-1199`) could sit in the `retry_seams` test-seam module, but it matches the 16 existing `#[cfg(test)]` items in `gateway.rs`, and the file is under the size threshold. Not a finding.
2. Decoupled:    PASS - `begin_frame_since` is a private `fn` (`gateway.rs:702`). The only new crate-visible item is `pub(crate)` under `#[cfg(test)]` (`gateway.rs:1188-1189`). The budget exception reads state the unit already owns (`self.screenshot`, `screenshot.rs:207`), not state held by a caller. The test helpers (`settle_flow_pane`, `control_plane_tests.rs:826`) read through existing projections (`control::chart::viewport_snapshot`).
3. AI-ready:     PASS - The diff adds no new behaviour. It changes when the UI thread serves the existing `evidence.capture` call, which already goes through the capability registry and returns a typed `coverage.not_captured` gap (`control/evidence.rs:526-529`). The fix removes a false `frame_not_delivered` gap for a frame that was actually delivered, so a typed result becomes more honest (Data honesty). The test seams are not capabilities.
4. Agent-tested: WEAK - PASS on the fix itself. `a_parked_capture_claims_its_image_in_a_frame_whose_budget_is_already_spent` (`control_plane_tests.rs:4684`) fails without the diff: the over-budget frame pushes the waiter back, so `still_parked` is 1, not 0. It runs alone with `cargo test -p quantick-app a_parked_capture_claims`. Its expectations come from the contract, not from the code under test: capture revision equality, and no `frame_not_delivered`. It is deterministic because the spent budget is built into the seam (`gateway.rs:1194-1198`), not timed. It cannot sit below `app`, because the gateway lives in `app`. Deflake hunt: `Instant` and `thread::sleep` appear only as bounded deadlines against `GATEWAY_TEST_WAIT`, and never feed an expected value. `settle_flow_pane` and `settle_indicators` replace frame counting with an explicit flush. WEAK: both new retry loops pause for a literal 50 ms (`control_plane_tests.rs:2757`, `:5564`) that silently restates `CONTROL_CLIENT_RATE_PER_SECOND = 20`. If the rate drops to 10, the burst runs out under contention and the limiter's `BACKPRESSURE` reads as a gateway defect. Thread `PRRT_kwDOTfuoRs6h2lif`.
5. Extensible:   PASS - The exception depends on "this frame holds an image" (`self.screenshot.is_none()`, `screenshot.rs:207`), not on a capability name. Any future waiter whose dispatch `needs_screenshot()` gets it without an edit here. The count ceiling (`CONTROL_UI_MAX_REQUESTS_PER_FRAME`) is unchanged and still applies.
6. Scalable:     PASS - Rate: `serve_awaiting_screenshot` runs once per frame but returns at once when nothing is parked (`screenshot.rs:167`). That covers every frame of a session with no capture client. The time-budget bypass fires only in a frame that holds a rasterised image. The first served capture `take()`s that image (`evidence.rs:526`), so later waiters fall back to the time check. The overrun is bounded to one capture per rasterised frame, under the unchanged count cap of 4. The parked queue keeps its cap (`CONTROL_MAX_SCREENSHOT_WAITERS`). No new buffer, and no silent drop. The test loops are bounded by `GATEWAY_TEST_WAIT` (30 s) or by 8 rounds (`settle_flow_pane`).
Top fix: `crates/app/src/app/tests/control_plane_tests.rs` - derive the retry pause from `quantick_control::limits::CONTROL_CLIENT_RATE_PER_SECOND` in one helper shared by both loops - flips: 4 WEAK -> PASS

milocaetano:
## arch-review — PR #426 @ `5a9c5c26`

Base `origin/campaign/lean-a-plus` @ `a3c962fa` (campaign child Q10 of #367). Tier `high` (read from the worktree's `mission-tier`; it agrees with the archived goal's `**Tier:**` line). Full shape pass, all nine dimensions.

**step 0: `code-review` at medium (effort-first, no reuse notice), three rounds, 3 findings, 3 confirmed and fixed.**

| ID | Round | Where | Severity here | Finding | Disposition |
| --- | --- | --- | --- | --- | --- |
| S0-1 | 1 @ `c32de2b0` | `control_plane_tests.rs` #411 reuse loop | Should fix | It retried every 5 ms, faster than the connection's rate limit (burst 40, then 20/s), so under load the limiter's `control.backpressure` would end the loop and be reported as a leaked ID. | Fixed in `94954c96`, and derived from the limit in `01a367b3` (`pause_one_request_interval`) |
| S0-2 | 2 @ `e9d3d1d8` | `gateway/screenshot.rs` `serve_awaiting_screenshot` | **Blocker** (correctness) | The time-budget exemption covered only the first parked capture. Once it took the image, the rest were held back with only a repaint request, so nothing asked the window for another frame. On a machine that spends the budget on every frame they ended in a bare `control.timeout`. | Fixed in `01a367b3`: a held-back waiter calls `arm_screenshot`. New test `captures_parked_behind_the_first_ask_for_the_next_frame_when_the_budget_is_spent` fails without it (`the capture still parked asks the window for the next frame`) and passes with it. |
| S0-3 | 3 @ `01a367b3` | `control_plane_tests.rs` new test's park wait | Consider | The outer wait for the second capture ran frames back to back once the first had parked. | Fixed in `5a9c5c26` (1 ms sleep per iteration) |

Round 3 reported no correctness defect in the product change. The delta after it (`5a9c5c26`) is one test-only sleep, read in full here.

### Shape findings

- **Standardisation, Should fix, fixed.** A second indicator-settling helper duplicated the suite's `settle_indicators` (35 existing callers). Replaced in `94954c96`.
- **Hardcoded values, Consider, not changed.** `settle_flow_pane`'s `for _ in 0..8` is a test-only bound whose panic message states it. The `750` ms window in #410 is a named local with a comment citing the `retry_readback_tests` precedent. Neither configures product behaviour.
- **Tests, Consider, not changed.** Round 1's bug pass doubted, without a reproduction, that `settle_flow_pane` stops after the first round that changes nothing. The frame could hand the worker input whose result lands later. Not reproduced: #408 passed in every contention copy.

### Verdict

- **Correctness**: step 0 ran three rounds at medium. All three findings are fixed, one of them a Blocker in the product path, now covered by a test. Nothing is open.
- **Docking**: no capability added. The change is one condition and one call inside the screenshot arc's own loop. The test seam `begin_frame_over_budget_for_test` sits with the other `*_for_test` gateway seams and is `#[cfg(test)]`. Production `begin_frame` forwards to a private `begin_frame_since(Instant::now())`, with identical work.
- **Performance**: the product path is rare. `serve_awaiting_screenshot` returns before the changed lines unless a capture is parked. With one parked, it costs one `Option::is_none`. It adds one viewport command per held-back frame, only while captures wait (the same cost as the existing re-arm path). One capture may run past the 250 µs budget in a frame that holds an image, still under the four-per-frame count ceiling. Per frame, `begin_frame` makes one extra call. Everything else is `#[cfg(test)]`.
- **Operability**: no new surface. `evidence.capture` stays the named call, and its answer is now honest under load: the image a delivered frame carries, instead of a false `frame_not_delivered` or a bare timeout.
- **Proof**: integration tests through the real local gateway, `a_parked_capture_claims_its_image_in_a_frame_whose_budget_is_already_spent` and `captures_parked_behind_the_first_ask_for_the_next_frame_when_the_budget_is_spent`. Each fails deterministically without its half of the fix. The eight load-sensitive tests are proven under contention (PR body).
- **Accumulation**: trunk flat for tracked files. `control/gateway.rs` +28 lines (1,255 lines, under the 1,500 threshold, no ceiling). `gateway/screenshot.rs` +15. No baseline ceiling raised. The `guards --report` `file.largest` list is unchanged.
- **Language**: the guard passed (`cargo test -p quantick-guards` and `cargo test --workspace`). I read the branch name, all seven commit messages, the PR title and body, and the archive myself: all English. The one Portuguese string seen in this work was a Windows tool error on the local console, which never reached the repository.

milocaetano:
## AI review - PR #426 (`fix/contention-listed-tests`) @ 5a9c5c26b9c93ff8b8d005e882b16756b7cee444 (follow-up)
Identity: HEAD `5a9c5c26b9c93ff8b8d005e882b16756b7cee444`; branch `fix/contention-listed-tests`; worktree `C:\src\quantick-worktrees\fix-contention-listed-tests` (clean); base ref `origin/campaign/lean-a-plus`; base tip `a3c962fa2d209fc8473927902cf3dfffcc68c5a2` (unchanged, matches `git ls-remote`); review key `de2ec018e47bc6c68af321e2678bd4643862b0d5`
Scope: follow-up of the round-one review at `e9d3d1d8` (https://github.com/milocaetano/quantick/pull/426#issuecomment-5651287033). Delta `e9d3d1d8..5a9c5c26` covers commits `01a367b3` and `5a9c5c26`: 2 files, +120/-12. The base is unchanged. Findings: 1 (`PRRT_kwDOTfuoRs6h2lif`), now fixed and resolved. Unresolved 0, accepted 0, fixed 1. No new FAIL.

Open thread verified:
- `PRRT_kwDOTfuoRs6h2lif` (WEAK, dimension 4): **fixed**. Both retry loops (`control_plane_tests.rs:2765`, `:5659`) now call `pause_one_request_interval()` (`control_plane_tests.rs:2680-2683`). It sleeps `Duration::from_secs(1) / CONTROL_CLIENT_RATE_PER_SECOND`, so a lower limit slows the loops instead of draining the burst. Resolved.

Narrow check of the delta (no new FAIL):
- `screenshot.rs:223`: a waiter held back on time or count now calls `arm_screenshot(ctx)` rather than only `request_repaint()`. This is correct. The first waiter's `accept_screenshot` cleared `screenshot_armed`, and the frame's image is dropped at frame end, so the waiters behind it would otherwise wait on a picture nobody had asked for, until their deadline answered them with a bare timeout. Rate: at most one `ViewportCommand::Screenshot` per frame while a waiter is held back. That is the same cadence as the existing in-budget path, which already re-arms each frame for a still-parked waiter (`screenshot.rs:238`). The waiter cap (`CONTROL_MAX_SCREENSHOT_WAITERS`) is unchanged. Each extra image still enters through `accept_screenshot`, so the trader's capture notice is never bypassed.
- `screenshot.rs:49-54`: `screenshot_armed_for_test` is a read-only `#[cfg(test)] pub(crate)` accessor. It adds no production surface.
- The new test `captures_parked_behind_the_first_ask_for_the_next_frame_when_the_budget_is_spent` (`control_plane_tests.rs:4756`) fails without `01a367b3`: `armed` is false after the first claim. Its expectations come from the contract: exactly one waiter still parked, armed, then served, and each answer matched to its own request ID with its own capture revision. Answer order is not assumed. The over-budget frames come from the seam, not from timing. The 1 ms sleep and the `Instant` deadline only bound the park wait against `GATEWAY_TEST_WAIT`. They never feed an expected value.

1. Modular:      PASS - the delta stays inside `serve_awaiting_screenshot` and the test file (`screenshot.rs:223`).
2. Decoupled:    PASS - the only new item is a `#[cfg(test)]` accessor over state the unit owns (`screenshot.rs:52`).
3. AI-ready:     PASS - no new behaviour or capability. `evidence.capture` for a second parked caller now returns its image instead of a bare `control.timeout` (`screenshot.rs:223`).
4. Agent-tested: PASS - the WEAK is fixed (`control_plane_tests.rs:2680-2683`), and the new test fails without the delta (`control_plane_tests.rs:4803-4806`).
5. Extensible:   PASS - unchanged. Re-arming depends on held-back waiters, not on a capability name.
6. Scalable:     PASS - one screenshot command per frame while a waiter is held back, under the unchanged waiter cap. This is a rare path: a capture client is present and the budget is spent.
Top fix: none - no open findings.

milocaetano:
## delivery-review — PR #426 @ `5a9c5c26` — **FAIL** (one traceability finding, no product gap)

Base `origin/campaign/lean-a-plus` @ `a3c962fa` (campaign child Q10 of #367). Review key `de2ec018e47bc6c68af321e2678bd4643862b0d5`. Tier `high` (git dir `mission-tier`), so this was a **full** review: dossier, fresh criteria subagent on `sonnet`, all three passes. Independent of the implementing session.

**Checklist source:** 1 — the committed archive `.claude/GOAL-archive-contention-listed-tests.md` (R1–R7, D1–D6, S1–S4, A1–A5, G1–G4, C1–C3, verbatim request). Issue #418 was read as the second retained source. The completeness pass ran.

**Bracket:** HEAD `5a9c5c26`, clean tree, `arch-review-ok` / `ai-review-complete` unchanged (02:43 / 02:44), and no `delivery-review-ok`, both before the first dispatch and after the last verdict. Dossier: `git diff origin/campaign/lean-a-plus...HEAD` (6 files, +570/−108, 7 commits).

### Failing line first

**DR-Q10-1 — UNLEDGERED (traceability only): the child's write scope has no line in the map.** Attempt 0.

- *Source span* (the coordinator's prompt, "Worktree"): "You own the control-plane test modules under `crates/app/src/app/tests/`, `screenshot_evidence_tests.rs`, and (only if D2 finds a product race) the evidence capture path under `crates/app/src/control/evidence*` and `gateway/screenshot.rs`. A sibling (Q11) owns `crates/mcp/*`; Q4's PR (in integration) touches `crates/app/src/app/tests/mod.rs`: avoid editing that shared helper file unless indispensable, and say so." Also (Read first, 5): "do NOT commit it [`contention.sh`] to your branch".
- *What is missing:* no `R`, `G` or `C` line records this constraint. It is an operational obligation (delivery contract), so it needs a `G` line with its evidence. This campaign's own precedent is the same: `0c88cd98` added G5 ("the branch writes only … the files this child owns") after a completeness pass found this same gap.
- *Why it matters here:* the branch writes `crates/app/src/control/gateway.rs` (+30/−1). That file is **outside** the owned set above. The only map line that touches it is S3, and S3 calls the seam "not a production change". But the diff also edits production code there: `begin_frame` now forwards to a new private `begin_frame_since(…, Instant::now())`. The work is the same, and the PR body's *Performance impact* discloses it. The map still says otherwise.
- *What is already true (so this is not a product gap):* the `mod.rs` edit is stated as indispensable in the PR body (*Notes for the reviewer*: `capture_with_screenshot`, used by #416, lives there). `contention.sh` is not committed. `crates/mcp/*` is untouched. No other open PR against `campaign/lean-a-plus` touches `mod.rs` or `gateway.rs` (#414 and #431 checked).
- **Smallest fix:** add a `G5` to the archive. It states the write scope with the source span above. For evidence it cites the 6-file stat, the PR-body sentence about `mod.rs`, and the absent `contention.sh` and `crates/mcp` edits. It names `gateway.rs` as an **extension beyond the owned list**: a behaviour-identical `begin_frame` → `begin_frame_since` split plus a `#[cfg(test)]` seam, taken under D2 because it is the frame that runs the screenshot arc. Correct S3's "not a production change" in the same edit. Whether that extension is within the grant is the **coordinator's call**, so name it in the handoff. If the coordinator refuses it, the seam has to move, and that is a scope decision, not a repair round. This is a record-only delta: per the contract's third validation row, run guards, diff hygiene and link checks. Nothing at runtime needs to rerun. The new key then needs the arch/ai follow-ups, a delivery-review follow-up and CI at the new head.

### Completeness pass (source first, strong model)

I read the verbatim request and #418 first and derived the asks: the D2 question and its answer; a product fix or typed unavailability, with a real-gateway test; the wire-shape stop clause; D3's forcing for five tests; no weakened assertion; ≥10 contention rounds, before/after, with issue evidence; skip-list lines; D5's reviews and step-0 cap; D6's single `gh pr ready` with no merge; #418 A4's merge. Reconciliation: R1–R7, A1–A5, G1–G4 and C1–C3 cover all of them. The D2 stop clause is in R2/A2, #418 A4 is in C3, D6 is in C2, and #425 is recorded under S2 and in the PR's *Human decision*. The one real gap is **DR-Q10-1**. The suggested PR title ("adjust if no product race") was changed even though D2 found a race. That is wording, not an outcome.

### Ledger pass

| R | Grade | By |
| --- | --- | --- |
| R1 | COVERED | A1 |
| R2 | COVERED | A2 |
| R3 | COVERED | A3 |
| R4 | COVERED | A2, A3 |
| R5 | COVERED | A4 |
| R6 | COVERED | A5 |
| R7 | COVERED | A4, A5 (the eight are no longer excused; the other load-sensitive tests are filed as #427–#430 and #432, and the PR says plainly that D19's 10-green condition is not met by this PR alone) |

### Criteria pass (sonnet; every line DELIVERED on the first round, so nothing escalated)

| ID | Grade | Evidence |
| --- | --- | --- |
| A1 | DELIVERED | PR body *D2 finding*. Checked against the code: `capture_revision` is a non-optional `WireU64` (`control/evidence.rs:198,301,350`), and the manifest's `screenshot` is an `Option` (`evidence.rs:224`). So the Null was an absent object. The captured coverage shows `frame_not_delivered`. |
| A2 | DELIVERED | `gateway/screenshot.rs`: `out_of_time = … > CONTROL_UI_BUDGET_US && self.screenshot.is_none()`, and a held-back waiter now calls `arm_screenshot`. The seam `begin_frame_over_budget_for_test` sets `started = now − (budget + 1 µs)`, so the spent budget is deterministic, not timed. Both new tests' assertions were read: `still_parked == 0`, the image stamped with the capture's revision, no `frame_not_delivered`; the second waiter stays parked and armed, then is served, with answers matched by request ID. Nothing under `crates/control/` changed and no serde type changed. The "fails before" text is a PR-body quote, and removing the guard makes it follow from the code. |
| A3 | DELIVERED | Every `-` assertion in the five D3 tests has an equal or stronger `+`. #415 adds a pre-condition assertion. #411's final `Success` assertion stays, and the loop retries only on `control.invalid_request`. The #413, #416 and #417 assertions are untouched; only the shared park helper changed. |
| A4 | DELIVERED | Recounted from raw logs with `summarize_q10.py`. `logs-final2-r01..r10` (binary built 28 s after `01a367b3`, mask 0xF, 24 copies): 0/240 for each of the eight, both new tests and S4. `logs-whole-before-r01..r03`: #410 1,1,0 and #411 1,5,9, matching the table cell for cell. The other-tests table matches too. The 0x3 table's three after-failures are all at `.select(None).unwrap()` → `control.instance_gone` (lines 4266, 2697 and 4639 at `01a367b3`), as the body says. Reusing the `01a367b3` runs is sound: the delta to `5a9c5c26` is a 1 ms sleep in a new test only. |
| A5 | DELIVERED | PR body: "all eight entries, lines 11 to 18". Checked against `origin/ci/app-tests-under-contention:tools/ci/contention-known-issues.txt`: lines 11–18 are exactly #408–#417's eight names. |
| G1 | DELIVERED | Added lines, all 7 commit messages, and the PR title and body are English. The arch-review language check agrees. |
| G2 | DELIVERED | CI run 34740956179 at `5a9c5c26`: `Format`, `Lint`, `Build` and `Test` are separate green steps (plus the `windows` Build/Test). The performance impact is declared in the PR body. |
| G3 | DELIVERED | The arch-review at `5a9c5c26` ran code-review at medium for 3 rounds (≤ D5's cap). The Blocker and the Should-fix are fixed. The ai-review follow-up at `5a9c5c26` fixed and resolved the WEAK. `ai_review_threads.sh count 426` → 0. |
| G4 | DELIVERED | CI conclusion `success`, headSha `5a9c5c26…`, the exact reviewed head. |

Closing steps: C1 is this review, and not yet PASS. C2: the PR is open against `campaign/lean-a-plus` and still draft, which is correct until PASS. C3 is pending with the coordinator.

**Assumptions (rule 7):** S2 drove the design, because #411 is fixed at the test layer and the product race is left in place. It is correctly surfaced as a human decision (#425), not guessed. S3 is inaccurate as worded (see DR-Q10-1). S1 and S4 are stated and disclosed.
**Exclusions (rule 8):** the three *Not applicable* rows hold against the diff. Nothing touches the UI, the engine or the hot path.
**Rule 9, what could have failed and did not:** a wire-type change (none), a removed or loosened assertion (none), recount mismatches (none, including an S4 1/24 that corroborated the body instead of contradicting it), a stale CI head (it matches), open threads (0), a moved branch or marker during the review (none).

**Counts:** UNLEDGERED 1 (traceability) · R 7/7 COVERED · A 5/5 and G 4/4 DELIVERED · UNPROVEN, MISSING and PARTIAL 0. **Marker: not recorded.**

milocaetano:
## arch-review follow-up: PR #426 @ `60ef4053` (delta from `5a9c5c26`)

Base unchanged: `origin/campaign/lean-a-plus` @ `a3c962fa`. New key `a2aa4d09…`.

**Delta, read in full:** `5a9c5c26..60ef4053` touches one file, `.claude/GOAL-archive-contention-listed-tests.md`: a new gate G5 (the child's write scope and its evidence) and a corrected S3. It is prose only. No executable, test, fixture or config input changed, so the delivery contract's evidence-only row applies: `cargo test -p quantick-guards` passed and `git diff --check` is clean. Runtime evidence from `5a9c5c26` carries over, and CI runs again at this head.

**step 0:** not re-run. The three allowed rounds were used at `c32de2b0`, `e9d3d1d8` and `01a367b3`, and the delta contains no code for a bug pass to read.

**Dimension 8 (never waived):** the new prose is English. The guard's language scan passed.

**Carried forward unchanged:** every finding and verdict line of https://github.com/milocaetano/quantick/pull/426#issuecomment-5651453581. The corrected S3 now states the same thing the review found: `begin_frame_since` is a production refactor that changes no behaviour. It is not a test-only change.

Verdict: unchanged, and nothing open.

milocaetano:
## AI review - PR #426 (`fix/contention-listed-tests`) @ 60ef40535927342dcf8fe9f965bbb8d743c022be (follow-up 2)
Identity: HEAD `60ef40535927342dcf8fe9f965bbb8d743c022be`; branch `fix/contention-listed-tests`; worktree `C:\src\quantick-worktrees\fix-contention-listed-tests` (clean); base ref `origin/campaign/lean-a-plus`; base tip `a3c962fa2d209fc8473927902cf3dfffcc68c5a2` (unchanged, matches `git ls-remote`); review key `a2aa4d094073221bbf8e501a22891cca7e14952c`
Scope: follow-up of the review at `5a9c5c26` (https://github.com/milocaetano/quantick/pull/426#issuecomment-5651457726). Delta `5a9c5c26..60ef4053` is one commit (`60ef4053`), 1 file, +22/-3, prose only, in `.claude/GOAL-archive-contention-listed-tests.md`. The base is unchanged. Findings: 0 new. `PRRT_kwDOTfuoRs6h2lif` stays fixed and resolved. Unresolved 0, accepted 0, fixed 1.

Narrow check: no code changed, so no dimension can move and all six stay PASS from the previous round. The new ledger text is English. Its statements about the code match the diff I reviewed. S3 describes the `begin_frame` -> private `begin_frame_since(frame_started)` split, called with `Instant::now()`, and the `#[cfg(test)]` `begin_frame_over_budget_for_test` seam (`crates/app/src/control/gateway.rs:694-707`, `:1185-1199`). G5 says `tests/mod.rs` holds only the park-wait helper (`run_frames_until_capture_parks`) and that neither `crates/mcp` nor `tools/ci` is touched, which the diff's file list confirms. G5's claim that no open campaign PR touches `control/gateway.rs` is a coordination fact and is not graded here; that belongs to the coordinator or `delivery-review`.

1. Modular:      PASS - no code in the delta.
2. Decoupled:    PASS - no code in the delta.
3. AI-ready:     PASS - no code in the delta.
4. Agent-tested: PASS - no code in the delta.
5. Extensible:   PASS - no code in the delta.
6. Scalable:     PASS - no code in the delta; no touched path.
Top fix: none - no open findings.

milocaetano:
## delivery-review follow-up — PR #426 @ `60ef4053` — **PASS**

This follow-up applies the delivery contract's delta rules to the FAIL at `5a9c5c26` (https://github.com/milocaetano/quantick/pull/426#issuecomment-5651844726). The base is unchanged: `origin/campaign/lean-a-plus` @ `a3c962fa`. The new review key is `a2aa4d094073221bbf8e501a22891cca7e14952c`. The scope is unchanged, and the mode is unchanged: full, tier `high`.

**Delta, read in full:** `5a9c5c26..60ef4053` is one commit touching one file, `.claude/GOAL-archive-contention-listed-tests.md` (+22/−3): a new **G5** and a reworded **S3**. It changes no executable, test, fixture or config input.

### DR-Q10-1 (UNLEDGERED, traceability): **closed** (attempt 1)

- **G5** records the write scope from the source span (the owned files, Q11's `crates/mcp/*`, the uncommitted `contention.sh`, and `tests/mod.rs` only if indispensable and said so). It names `control/gateway.rs` as outside the owned list, justified under D2 and raised for the coordinator. I checked its evidence claims against the branch:
  - `git diff origin/campaign/lean-a-plus...HEAD --name-only` lists 6 files, with no `crates/mcp` and no `tools/ci`.
  - The `mod.rs` diff is only the park-wait helper (`run_frames_until_capture_parks` replacing `PARK_WAIT_FRAMES`).
  - The open campaign PRs #414 and #431 touch neither `mod.rs`, `control/gateway.rs` nor `crates/mcp`.
- **S3** now reads, among other things: "a production refactor that splits `begin_frame` into a private `begin_frame_since(frame_started)`. `begin_frame` calls it with `Instant::now()`, so the frame does identical work. That refactor is production code outside the owned file list (see G5)." This matches the `gateway.rs` diff.
- **G5 grade: DELIVERED.** Whether the coordinator accepts the `gateway.rs` extension is a coordinator decision, which G5 routes to the handoff. It is not a gap in this branch.

### Carried forward, unaffected by the delta

The completeness pass is otherwise unchanged, and nothing is now UNLEDGERED. R1–R7 are COVERED. A1–A5 and G1–G3 are DELIVERED, as in the prior report. No code, test or evidence input changed, so their conclusions stand.

G1 was re-checked on the new prose: it is English. For G3, the arch-review follow-up at `60ef4053` (https://github.com/milocaetano/quantick/pull/426#issuecomment-5651849401) and the ai-review follow-up at `60ef4053` (https://github.com/milocaetano/quantick/pull/426#issuecomment-5651852376) are both at the current key, and `ai_review_threads.sh count 426` returns 0.

**G4 re-graded at the new head: DELIVERED.** CI run 34744430953 has headSha `60ef4053…`, conclusion `success` (`ci` and `windows`).

**Bracket:** HEAD `60ef4053`, a clean tree, and unchanged markers (arch 04:07, ai 04:08) from the start of this follow-up to its verdict.

**Counts:** UNLEDGERED 0 · R 7/7 COVERED · A 5/5 and G 5/5 DELIVERED · UNPROVEN, MISSING and PARTIAL 0. Finding DR-Q10-1: attempts 1, closed. No deferrals.

Issue #434, opened 2026-09-13
fix(control): release the request ID before writing the answer, and fix five more load-sensitive tests


milocaetano:
## Context

Campaign child of https://github.com/milocaetano/quantick/issues/367 (task key Q13), decision D23. Base: `campaign/lean-a-plus`. Mission tier: high (gateway ordering).

Q10 (PR #426) found a product race and five load-sensitive tests outside its scope:
- #425: the gateway writes an answer before it releases the request ID, so a client that reuses the ID immediately is refused as a duplicate (#411's test now waits around it).
- #427, #428, #429, #430, #432: tests that fail under the CPU-contention harness in both builds.

## Scope

- #425: release the request ID before (or atomically with) writing the answer in `crates/app/src/control/gateway/server.rs`, preserving every other ordering guarantee; a real-gateway regression test that reuses the ID immediately after reading the answer; then remove #411's workaround wait in its test.
- #427, #428, #429, #430, #432: force the interleaving each assertion depends on (the shape #382, #400, #404, #426 used); never weaken an assertion; for #429 (a keep-up bound in the burst test) replace the timing assertion by a count-based one if the property is "no loss", keeping any timing claim in the measured artifacts instead.
- Each fixed test's line on the contention skip list (`tools/ci/contention-known-issues.txt`, from PR #414) is removed by whichever of #414 and this PR lands second.

## Acceptance criteria

- [ ] A1: #425 fixed with a real-gateway regression test; #411's workaround removed.
- [ ] A2: each of the five tests: 0 failures over at least 10 contention rounds (24 copies on 4 cores or the local equivalent), before/after table.
- [ ] A3: no assertion weakened; no wire-shape change.
- [ ] A4: merged into `campaign/lean-a-plus` through a PR whose base is exactly that branch, with the merge read back.

## Campaign assignment

Parent: https://github.com/milocaetano/quantick/issues/367. Stable key: Q13. Class: autonomous. Priority: 2. Risk: medium-high (gateway ordering). Executor: opus (D13). Dependencies: Q10 integrated (merge 989f2945). Branch/worktree/PR: null until claimed.

milocaetano:
executor: opus (D13); high tier; branch fix/gateway-request-id-ordering

milocaetano:
[...]

Issue #432, opened 2026-09-13
test(app): the optional-row replay matrix test times out a call under load


milocaetano:
Found by campaign #367 child Q10 (#418, PR #426) while proving its eight fixes; this test is not one of them and PR #426 does not touch it. Same class as #428 (the `retry_readback_tests` gateway window).

## What fails

`app::tests::retry_readback_tests::every_reachable_optional_row_replays_a_dropped_answer_and_begins_once` (#421) fails under CPU contention in `success_result` (`crates\app\src\app\tests\mod.rs:2300`):

```text
expected a success: ControlError { code: ErrorCode("control.timeout"), message: "request did not complete before its deadline", retryable: true, ... }
expected a success: ControlError { code: ErrorCode("control.timeout"), message: "request expired before application-thread dispatch", retryable: true, ... }
```

## Reproduced

A local equivalent of `tools/ci/contention.sh` (#414) on Windows: 24 concurrent copies of the whole `quantick-app` test binary, all held to processor mask `0xF` (4 of 12 cores), `--test-threads=4`, run from `crates/app` on a shared host. The runs used PR #426 at `01a367b3` on base `a3c962fa`.

| Rounds | Copies failing on this test |
| ---: | --- |
| 10 | 0, 0, 0, 0, 0, 0, 0, 5, 8, 1 of 24 |

milocaetano:
Fixed and integrated into `campaign/lean-a-plus` by PR #436 (head `de270b83`), merge commit `f74c0cbd8e91a3d67716160268dde933ee9bfcad` at 2026-09-13T13:15:07Z. CI green (run 34754351456); 0 failures in 240 contention copies per test; #425 has a real-gateway regression test. Closing as integrated_campaign (campaign #367, task Q13).

PR #431, opened 2026-09-13
perf(app): store the retained trade tape in chunks so growth never stalls the UI thread

milocaetano:
## Summary

Store each pane's retained trade tape in fixed-size chunks so that appending a print never copies the tape already held — removing the UI-thread stall at each capacity doubling (25 ms at print 2,097,152, 52 ms at 4,194,304) — with byte-identical output for every consumer, so that SE6 can move from 4 to an earned 5.

**Tier:** high (hot path: the tape is written on every live print on the UI thread and read per frame by the lane transport).

Campaign child of #367 (Q12); closes on integration: #423. Base: `campaign/lean-a-plus` at `a3c962fa`. Evidence page: [`docs/quality/chunked-tape.md`](docs/quality/chunked-tape.md); mission record: `.claude/GOAL-archive-chunked-trade-tape.md`.

## The chunk design

`quantick_engine::trade_tape::TradeTape` — one additive file in the headless engine, beside `Trade`:

- **65,536 prints per chunk** (3.5 MiB of 56-byte trades), each allocated once at full size and never grown. Power of two: a position is a shift and a mask. Under 2 % of the envelope's 222 MB tape (bounds the unused tail); one new chunk every 3.6 min at 300 prints/s; 61 at the envelope's edge.
- **Append**: writes into the newest chunk, kept beside the directory of full ones (a push reads nothing the directory points at); a full chunk moves into the directory and one new chunk is allocated. No print already held ever moves.
- **Read API**: `len`, `is_empty`, `get`/`Index`, `first`, `last`, `iter`/`range` (double-ended, exact-size), `since(start)`, `last_n(n)`, `slices(range)` (at most one contiguous slice per chunk, for bulk readers), `partition_point`; `TradeSeq` (a positional read trait: `len` + `get`, default binary search) for `[Trade]`, `Vec<Trade>` and the tape.
- `prepend` (history paging, rare) copies into a fresh tape, freeing each old chunk as it goes: peak = tape + one chunk (was two tapes).
- `push`, the small readers and the iterators are `#[inline]` — a non-generic function is not inlined across crates without it (no LTO), which cost a call per print until `d11d1b98`.

## Consumer changes

| Reader | Rate | Change |
| --- | --- | --- |
| `ChartState::ingest_live` | per print | `TradeTape::push` |
| `LaneTransport::command` | per frame | the same unsent suffix, copied slice by slice into one exact allocation — the only contiguous copy, of the requested range only |
[...]

Issue #430, opened 2026-09-13
test(app): the request-timeout gateway test asserts one frame drains the queue


milocaetano:
## What fails

`app::tests::control_plane_tests::gateway_request_timeout_is_structured_and_late_ui_work_is_discarded` fails under CPU contention:

```text
panicked at crates\app\src\app\tests\control_plane_tests.rs:1050:5:
assertion `left == right` failed
  left: 1
 right: 0
```

After the timed-out read, the test runs one `run_frame` and asserts the queue is empty. That is the #364 class: a frame drains only inside `CONTROL_UI_BUDGET_US` (250 us), and a loaded frame can spend the budget before the drain starts. `drain_gateway_requests` (#382) is the existing helper for this.

## Reproduced

Local equivalent of `tools/ci/contention.sh` (#414) on Windows: 24 concurrent copies of the whole `quantick-app` test binary, all held to processor mask `0xF` (4 of 12 cores), `--test-threads=4`, run from `crates/app`, on a shared host. Found by campaign #367 child Q10 (#418, PR #426) while proving its eight fixes; this test is not one of them and PR #426 does not touch it. *Before* = base `c55a4222`; *after* = PR #426 at `efbada89` (its diff does not reach this test).

| Build | Rounds | Copies failing on this test |
| --- | ---: | --- |
| before | 3 | 1, 0, 0 of 24 |
| after | 10 | one copy, in one round |

milocaetano:
Fixed and integrated into `campaign/lean-a-plus` by PR #436 (head `de270b83`), merge commit `f74c0cbd8e91a3d67716160268dde933ee9bfcad` at 2026-09-13T13:15:07Z. CI green (run 34754351456); 0 failures in 240 contention copies per test; #425 has a real-gateway regression test. Closing as integrated_campaign (campaign #367, task Q13).

PR #436, opened 2026-09-13
fix(control): release the request ID before writing the answer, and force five more load-sensitive interleavings

milocaetano:
## Summary

Release a request ID before writing its answer in the local control gateway (#425), and force what five more load-sensitive app tests assert (#427, #428, #429, #430, #432).

**Tier:** `high` (set by the coordinator: a production ordering change in the gateway). Campaign child of #367 (Q13); closes on integration: #434, #425, #427, #428, #429, #430, #432.

## #425: the ordering change and its guarantees

**Root cause.** Every terminal path in `gateway/server.rs` wrote the answer, then released the request ID (`send_response(...)`, then `slots.forget(...)`). A client that reads its answer and reuses the ID at once can reach the reader thread before the answering thread runs `forget`; on a loaded machine it did (15 of 65 instrumented runs in #425), and the reuse was refused with `control.invalid_request` / "request_id is already in flight on this connection", `retryable: false`. Contract §5.2 refuses a duplicate only *while the first request is in flight*.

**The change.** A tracked request ID is now a value, `InFlightId` in the new `gateway/server/answer.rs`: tracking returns it, and one function consumes it with the answer, `answer_and_release`, which encodes the frame, takes the connection's writer lock, releases the ID, then writes the frame. Dropped unanswered (a wait whose connection closed), an `InFlightId` releases the ID all the same, so no path can leave an ID held for the rest of the connection. A parked wait hands its `InFlightId` to the read it becomes; a response worker takes its own from a hand-over slot, so a worker that never started leaves it to be released with the refusal. `send_response` (refusals for requests never tracked) moved into the same file unchanged in behaviour. Paths now answering through `answer_and_release`:

| Path | Before | After |
| --- | --- | --- |
| `dispatch_prepared`, global buffered-response slots full | write, forget | release under the lock, write |
| `dispatch_prepared`, worker-side read | write, global slot back, forget | release under the lock, write, global slot back |
| `dispatch_prepared`, per-connection in-flight full | global slot back, write, forget | global slot back, release under the lock, write |
| response worker (UI requests, and a woken or timed-out wait's read) | write, forget, global slot back | release under the lock, write, global slot back |
| response worker could not be spawned | forget, counters back, write | counters back, release under the lock (the ID taken back from the hand-over slot), write |
| parked wait, `WakeReason::Shutdown` | write, forget | release under the lock, write |
| parked-wait thread could not be spawned | forget, write | the dropped closure releases the ID, then write (this runs on the reader thread, so no duplicate can be admitted in between) |

The two paths that write nothing (a wait whose connection closed or disconnected) still only release the ID, by dropping the `InFlightId`.

**Guarantees kept.**

- *No duplicate is accepted while the first is in flight.* The ID is released only when the answer is about to be written, under the lock every answer on the connection takes. A duplicate the reader admits between the release and the write is answered after the first answer, never before it. A client sees the same frames in the same order as a run where its duplicate reached the reader just after the old write-then-release, so the only new outcome on the wire is the one #425 asks for: a reuse sent after reading the answer is admitted.
- *No answer is lost.* Encoding still happens before the lock; the encode-refusal fallback is unchanged; a write that fails part-way still closes the connection.
- *Limits unchanged.* The global buffered-response slot is still returned after the write (it bounds answers not yet written), and the per-connection in-flight count is untouched (a keyed request still holds it through its settle window).
- *No wire change.* No type, field, error code, message or retryability in `quantick-control` changes.

**Regression test.** `control_plane_tests::a_client_that_reads_its_answer_can_reuse_the_request_id_at_once` connects to a real gateway with a `#[cfg(test)]` hook (`GatewayOptions::answer_written`, set through `ControlAccess::enable_for_test_observing_answers`). The test build samples, under the writer lock between the release and the write, whether the ID is still in flight, and reports it once the answer is written; the hook then holds the answering thread, which is where a preempted thread used to sit with the ID still in flight. The test asserts the ID was released before the frame was written, reads the snapshot answer, reuses its ID at once for `describe`, and must get a success while the writer is still held. It fails deterministically on both ways the old order can come back:

```text
# release after the write, at the call site (write, hook, then forget):
panicked at crates\app\src\app\tests\control_plane_tests.rs:2834:5:
a client that has read its answer may reuse the ID at once: Failure { error: ControlError { code: ErrorCode("control.invalid_request"), message: "request_id is already in flight on this connection", retryable: false, ... } }

# release moved to just below write_all inside write_answer:
panicked at crates\app\src\app\tests\control_plane_tests.rs:2824:5:
the ID was released before its answer's frame was written
```

**#411's workaround is gone.** `gateway_rejects_a_duplicate_request_id_while_a_wait_is_parked` sends the reuse once, with no retry loop, and asserts success.

## The five tests

| Issue | Test | Root cause | What the test now forces |
| --- | --- | --- | --- |
| #427 | `gateway_a_client_that_never_reads_does_not_stall_another` | The stalled client's eight answered replies hold the shared buffered-response slots until their threads write them, and the worker-side read before the UI read retried every 5 ms on `control.backpressure` / "global buffered response capacity is full". Captured: an instrumented copy of the base (24 copies of this test alone on one core, 72 runs) logged that refusal up to 16 times in one run. At 5 ms the retries run faster than the connection's limiter refills (20 per second, burst 40), so a run whose slots stay busy long enough spends the burst; the next request, the UI-side `snapshot`, is then refused by the limiter ("client request rate limit is exhausted") and never queued, which is exactly `gateway request queue did not reach 1 within 30s (queued 0)`. After the worker read has succeeded at least one global slot is free, so the limiter is the only refusal left on that path (reasoned from the code; the limiter refusal itself did not recur in the instrumented rounds). | The worker-side read's retries are paced by one interval of `CONTROL_CLIENT_RATE_PER_SECOND` (`pause_one_request_interval`, moved from `control_plane_tests.rs` to the suite's `mod.rs` so the retry-readback module can use it). The limiter's burst is then never spent, and the UI-side read reaches the queue. Every assertion unchanged. |
| #428 | `a_keyed_action_held_past_its_deadline_is_refused_as_unknown_and_reconciled_by_readback` | Its readbacks went through `remote_call`, which serves by running frames. A frame's drain admits requests only inside `CONTROL_UI_BUDGET_US` (250 us), and a loaded frame spends that before the drain starts (the #364 class), so the read waited frame after frame and expired inside the test's 750 ms window: `control.timeout`, "request did not complete before its deadline", at `success_result`. | `readback` now reads through `read_call`: one frame first (what `remote_call`'s frames gave the state), then the read served through the existing seam (`serve_queued_for_test`, the production `execute_on_ui`) the moment it is queued, as the module's `keyed_call` already serves its calls. Every assertion unchanged. |
| #432 | `every_reachable_optional_row_replays_a_dropped_answer_and_begins_once` | Same mechanism under the default 5 s window: "request did not complete before its deadline" and "request expired before application-thread dispatch" at `success_result`, both on readbacks. | The same `read_call`. Its wait for a moving readback is paced by `pause_one_request_interval` instead of 20 ms, so a slow rebuild cannot spend the connection's burst. Every assertion unchanged. |
| #430 | `gateway_request_timeout_is_structured_and_late_ui_work_is_discarded` | After the timed-out read, one `run_frame` was asserted to empty the queue; a loaded frame can spend its budget before the drain starts. | `drain_gateway_requests` (#382's helper): frames, bounded by `GATEWAY_TEST_WAIT`, must take the late request off the queue. Added: the late work is never answered a second time (`!reply_pending` over 100 ms), the "discarded" half of the test's name, which it did not assert before. |
| #429 | `live_envelope_tests::burst::inside_the_envelope_every_print_arrives_and_no_queue_fills` | The keep-up bound `late_frames * 2 < frames` is a wall-time claim: contention alone made the workers late on more than half the frames (`270 of 480 frames found the previous one untaken`) with nothing lost. | Per D2, the property the test is named after is asserted with counts, as before: every print in a bar, one CVD row per closed bar, the CVD values, one aggression per print, every depth update applied, every accepted command retired, `deferred`, `parked` and `coalesced_parked` 0. The `late_frames` bound is removed from the test; the keep-up claim is the measured one in `docs/quality/live-envelope/measure.txt` block 4, and `docs/quality/live-envelope.md` now says the test asserts counts only. The test still prints `late_frames`. |

**A sixth test, by #430's mechanism (mission S5).** `gateway_rejects_a_duplicate_request_id_while_the_first_is_in_flight` ran one `run_frame` and then a blocking read; it failed once in a local contention round of this branch (`control_plane_tests.rs:2411`, the served request not a success). It gets #430's `drain_gateway_requests`, assertions unchanged. It guards the duplicate rule #425 must keep, so it had to be reliable.

A side effect of `read_call`: every other test in `retry_readback_tests` reads back through the same function. Under this host's contention the base binary also failed `a_dropped_layout_answer_is_replayed_and_the_layout_is_made_once` and `an_interrupted_annotation_is_resolved_by_its_readback` on the same readback timeout; they are counted in the table below.

## Contention proof (D3)

Local equivalent of `tools/ci/contention.sh` (Linux-only `taskset`), the harness Q9 and Q10 used: a PowerShell script starts **24 concurrent copies of the whole `quantick-app` test binary**, each held to processor mask `0xF` (4 of this machine's 12 cores; children inherit the mask), `--test-threads=4`, from `crates/app`. The host is shared (other sessions and, in some rounds, a review agent's build), so these runs are at least as contended as CI.

- **Before**: the base binary at `989f2945` (the base this branch was cut from; the later base `6ba1a7f3` adds #431's chunked tape, which none of these tests' mechanisms touch), 3 rounds = 72 copies per test. A second set of 3 rounds ran while a workspace build shared the host ("busy"), and 2 rounds of an instrumented base binary (48 copies).
- **After**: the binary of `173d3018`, the last code commit (the head `de270b83` adds only mission prose), base `6ba1a7f3`, **10 rounds = 240 copies per test**.

| Issue | Test | Before: issue evidence | Before: local, `989f2945` | After: local, 10 rounds at `173d3018` |
| #425 | `a_client_that_reads_its_answer_can_reuse_the_request_id_at_once` (new) | n/a | fails deterministically on the old order (both mutations above) | **0 / 240** |
| #411 | `gateway_rejects_a_duplicate_request_id_while_a_wait_is_parked` (retry loop removed) | 15 of 65 instrumented runs refused the reuse (#425) | 0 / 72 (with the old retry loop) | **0 / 240** (no retry) |
| #427 | `gateway_a_client_that_never_reads_does_not_stall_another` | 0, 9, 1 of 24 (3 rounds); 6, 3, 0, 8, 7, 3, 10, 2, 5, 10 of 24 (10 rounds) | 0 / 72; instrumented: the worker read refused up to 16 times in one run | **0 / 240** |
| #428 | `a_keyed_action_held_past_its_deadline_…_by_readback` | 0, 2, 3 of 24; 3, 0, 0, 6, 0, 1, 9, 7, 0, 0 of 24 | **71 / 72** (busy: 69 / 72; instrumented: 43 / 48) | **0 / 240** |
| #429 | `live_envelope_tests::burst::inside_the_envelope_every_print_arrives_and_no_queue_fills` | 13 of 24 in one round | **5 / 72** (246 to 277 of 480 frames late; busy: 5 / 72; instrumented: 3 / 48) | **0 / 240** |
| #430 | `gateway_request_timeout_is_structured_and_late_ui_work_is_discarded` | 1, 0, 0 of 24; one copy in 10 rounds | 0 / 72 | **0 / 240** |
| #432 | `every_reachable_optional_row_replays_a_dropped_answer_and_begins_once` | 0, 0, 0, 0, 0, 0, 0, 5, 8, 1 of 24 | **72 / 72** (busy: 72 / 72; instrumented: 48 / 48) | **0 / 240** |
| S5 | `gateway_rejects_a_duplicate_request_id_while_the_first_is_in_flight` | not listed | 0 / 72; 1 of 96 copies of this branch's `0e464d3e` before its fix | **0 / 240** |

Every failure of the five in the base binary was at the assertion its issue names (`success_result` on a readback's `control.timeout` for #428 and #432, `N of 480 frames found the previous one untaken` for #429).

**Earlier after-runs, at intermediate heads**, for the record (each stopped when a review repair changed the code): `d7b93c3c` (before review), 6 full rounds: the six 0 except one #432 copy in round 1 that failed before its first readback, at the #438 discovery wait below; `0e464d3e` (after review round 1), 4 full rounds: the six 0, the S5 test 1 copy, which led to its fix.

**Not caused by this branch, and not among the six:** in the final 10 rounds, 16 of 240 copies failed, all in rounds 2, 3 and 10 and all at one place, `wait_for_test_gateway_descriptor` (`test gateway did not publish discovery within 30s`), on `synthetic_precision_…` (6), `synthetic_fractional_geometry_…` (4), `no_production_grant_…` (4), `the_feed_generation_…` (3) and `synthetic_fractional_image_…` (3). The base binary fails the same way (14 panics in 72 copies; 37 on the busy host). It happens before any of those tests asserts anything, in the gateway's start-up and discovery publish, which this branch does not touch. Filed as **#438**; each of those tests needs a fix or a skip-list line citing #438 before the step can hold ten green whole-binary rounds on this host.

## Skip-list lines this allows removing

Q9 is adding these to `tools/ci/contention-known-issues.txt` on #414 in parallel; whichever of #414 and this PR lands second removes them:

```text
app::tests::control_plane_tests::gateway_a_client_that_never_reads_does_not_stall_another #427
app::tests::retry_readback_tests::a_keyed_action_held_past_its_deadline_is_refused_as_unknown_and_reconciled_by_readback #428
live_envelope_tests::burst::inside_the_envelope_every_print_arrives_and_no_queue_fills #429
app::tests::control_plane_tests::gateway_request_timeout_is_structured_and_late_ui_work_is_discarded #430
app::tests::retry_readback_tests::every_reachable_optional_row_replays_a_dropped_answer_and_begins_once #432
```

The #411 line (`gateway_rejects_a_duplicate_request_id_while_a_wait_is_parked #411`) is already covered by #426's removal list. The #438 failures above are a new, separate exposure; this PR adds no skip-list line for them.

## Performance impact

- The answer path: **per control request**, not per trade, depth update or frame. The change moves one mutex-guarded `BTreeSet::remove` inside the writer lock the write already takes; encoding still happens outside it.
- Everything else is `#[cfg(test)]` or test code.

## Assumptions named for the reviewer

- **S2**: "atomically with writing" is read as "released under the writer lock, immediately before the frame is written".
- **S3**: only the request ID moves; the global buffered-response slot keeps its after-the-write release.
- **S4**: the regression test's hook is a `#[cfg(test)]` field on `GatewayOptions` and `ConnectionSlots`, with one `#[cfg(test)]` seam in `control/gateway.rs` (outside the owned file list, named in the mission). Nothing of it compiles into the binary.
- **S5**: the sixth test above, a stated detour.
- **S6**: `InFlightId` and the probe placement came from review rounds 1 and 2; both serve R1 and R2, no wire change.
- **AR-Q13-2 (Consider), not taken**: `ConnectionSlots::new` takes `&GatewayOptions` to copy one `#[cfg(test)]` field, with the allow scoped to that one parameter; one call site, so a second constructor would add more than it removes.

## Verification loop

Run locally at `173d3018` (the last code commit, base `6ba1a7f3`), each command on its own; the head `de270b83` changes only the mission archive's prose (evidence reused from `173d3018`, same base and toolchain):

- [x] `env -u QUANTICK_BUBBLES cargo test --workspace` (`quantick-app` binary: 2065 passed, 0 failed, 10 ignored)
- [x] `cargo test -p quantick-guards` (`server.rs` goes from 1,487 to 1,466 production lines)

Plus the contention runs above (local), and required CI at the final head.

## Reviews

- `arch-review` (step 0 `code-review` at `medium`, three rounds, full shape pass): round 1 at `05426152` https://github.com/milocaetano/quantick/pull/436#issuecomment-5652582184 (AR-Q13-1 Should-fix, AR-Q13-2 Consider); round 2 at `0e464d3e` https://github.com/milocaetano/quantick/pull/436#issuecomment-5652855378 (AR-Q13-1 partly fixed, AR-Q13-3 Should-fix new, AR-Q13-2 declined and accepted); round 3 at `de270b83` https://github.com/milocaetano/quantick/pull/436#issuecomment-5653030119: **pass**, 0 step-0 findings, every Blocker and Should-fix resolved.
- `ai-review`: round 1 https://github.com/milocaetano/quantick/pull/436#issuecomment-5652588337 (Agent-tested FAIL, Decoupled WEAK, two threads); follow-ups https://github.com/milocaetano/quantick/pull/436#issuecomment-5652842676 and https://github.com/milocaetano/quantick/pull/436#issuecomment-5653024672: all six PASS, **0 open threads**, completion recorded for key `11edbdf5`.
- Repair progress is on the PR as `quantick-review-progress:v1` records (2 batches of 3, no stalls).
- `delivery-review` (full, tier `high`) at `de270b83`: **PASS**, no findings, https://github.com/milocaetano/quantick/pull/436#issuecomment-5653492428.

CI at the head `de270b83`: https://github.com/milocaetano/quantick/actions/runs/34754351456 (`ci` and `windows` pass).


crates/app/src/control/gateway/server/answer.rs
```diff
@@ -0,0 +1,88 @@
...
+pub(super) fn answer_and_release(
+    writer: &Arc<Mutex<TcpStream>>,
+    codec: &BoundedCodec,
+    response: ResponseEnvelope,
+    slots: &ConnectionSlots,
+) {
+    let request_id = response.request_id.clone();
+    write_answer(writer, codec, response, || slots.forget(&request_id));
+    #[cfg(test)]
+    if let Some(written) = &slots.answer_written {
```
milocaetano: **FAIL - 4. Agent-tested: the new test does not pin the ordering it names (#425)**

`a_client_that_reads_its_answer_can_reuse_the_request_id_at_once` (`control_plane_tests.rs:2760`) sends the reuse only after `written.recv_timeout(...)` returns (`:2818-2820`), and this hook fires after `write_answer` has returned, so after the release in any order. The expectation comes from where the code under test puts its hook, not from the window the bug lives in.

Variants that keep the test green and reopen #425:
- the pre-fix order moved into the helper: `write_answer(writer, codec, response, || {}); slots.forget(&request_id);` then the hook;
- `release()` moved below `stream.write_all(&frame)` in `write_answer` (`answer.rs:74-87`). The reader's duplicate check (`server.rs:840`, `slots.is_in_flight`) takes only the `in_flight_ids` lock, not the writer lock, so a thread preempted between the write and the release makes the reader refuse the reuse.

The test fails without the diff only because `answer_written` does not exist there. Held at this line, the thread has already done everything the fix reorders.

Fix: observe inside `write_answer`, right after `write_all` and while the writer lock is still held, whether the ID is still in flight: pass `slots.is_in_flight(&request_id)` to the observer, or hold the thread there and have the test check a test-only in-flight readback before it releases the hold. The assertion "the ID is free at the moment its answer is on the wire" then fails under both variants above. Flips 4 to PASS.

crates/app/src/control/gateway/server.rs
```diff
@@ -67,6 +67,11 @@ use super::{
     WAITER_POLL_MS,
 };
 
+mod answer;
+#[cfg(test)]
+pub(super) use answer::AnswerWritten;
+use answer::{answer_and_release, send_response};
```
milocaetano: **WEAK - 2. Decoupled: the caller still decides which answers release an ID**

The diff moves the release into `answer_and_release`, but whether an ID is tracked is still state the call site carries in its head. Both writers are imported side by side here, and nothing ties `slots.track(&envelope.request_id)` (`server.rs:1048`, `:1273`) to the function that must later answer that ID. `answer_and_release` also forgets IDs that were never tracked (the reader-path worker read at `server.rs:1022`), so the two helpers differ only by a convention.

Variant that breaks it: the next refusal added to `dispatch_prepared` or `dispatch_parked_wait` after `slots.track`, such as a new capacity check, written with `send_response` like the pre-track refusals around it (`server.rs:1222`, `:1238`, `:1262`). It compiles, answers once, and leaves the ID in `in_flight_ids` for the rest of the connection's life. Every later reuse of that ID then gets the non-retryable `request_id is already in flight` refusal.

Fix: have `ConnectionSlots::track` return a non-`Clone` `TrackedId` that `answer_and_release` consumes, with `Drop` releasing it as the fallback. Then only the owner of a tracked ID can answer it, and a path that forgets to cannot leak it. Keep `send_response` for requests that were never tracked.

milocaetano:
## arch-review — PR #436, round 1

- **Head reviewed:** `0542615245e42bd79b75797742a8017259de1d33`
- **Base:** `origin/campaign/lean-a-plus` at `6ba1a7f3831f08ac5eafdc09f1d1735f00cb23d7` (from `campaign_context.sh base`; equals the PR's `baseRefOid`)
- **Review key:** `9f732346cbd2491c5d9bacd892f49319b0efae4b` (`campaign_context.sh key`)
- **Scope:** `git diff origin/campaign/lean-a-plus...HEAD`, 4 commits, 9 files, +535/−129: `control/gateway/server.rs`, new `control/gateway/server/answer.rs`, a `#[cfg(test)]` seam in `control/gateway.rs`, `app/tests/{mod,control_plane_tests,retry_readback_tests}.rs`, `live_envelope_tests/burst.rs`, `docs/quality/live-envelope.md`, the mission archive.
- **Tier:** `high` (git-dir `mission-tier`: `fix/gateway-request-id-ordering high`; agrees with the goal file's `**Tier:**`). Full shape pass, all nine dimensions.

**step 0: code-review at medium (effort-first, `Skill(code-review, "medium 436")`; no reuse notice in the returned report), 1 finding, 1 confirmed (narrowed, see AR-Q13-1).** Round 1 of at most 3.

Step 0 checked and cleared: every former `send_response` + `slots.forget` terminal pair now goes through `answer_and_release`; the only direct `forget` calls left are the parked wait's Disconnected / closed-socket paths, which write nothing. Lock order is writer lock → `in_flight_ids`, and no path takes them the other way (`is_in_flight`, `track` and `forget` drop the ID lock before any write), so no deadlock. A duplicate admitted between release and write is answered after the first answer. A total encode failure still releases the ID. I re-derived each of these from `server.rs:316-345, 790-1120, 1180-1350` and `answer.rs` and agree. One further effect: the old code forgot the ID *after* the write, so a reuse admitted in that gap could have its fresh tracking erased by the late `forget`. No `forget` now runs after an answer is written, so that latent hazard is gone too.

### Findings

**AR-Q13-1 — Should-fix — test-coverage** — `crates/app/src/control/gateway/server/answer.rs:75` (release) / `:41` (hook); test `crates/app/src/app/tests/control_plane_tests.rs:2760`
The `answer_written` hook fires after `write_answer` has returned, so the ID is already released at the hold point whatever order `write_answer` uses internally. If someone moves `release()` (line 75) below `stream.write_all(&frame)`, #425 comes back exactly: the answering thread can be preempted between write and release. `a_client_that_reads_its_answer_can_reuse_the_request_id_at_once` still passes in that case. The test does catch the call-site regression: a `forget` placed after `answer_and_release` runs after the hook, and a path moved back to `send_response` never fires the hook and fails at `:2820`. The in-function order is what the fix is about, and nothing pins it. Now that #411's retry loop is gone, `gateway_rejects_a_duplicate_request_id_while_a_wait_is_parked` would catch it only by luck, as a flake. Step 0 raised this; its "can never catch" wording is too broad, and it is confirmed at this narrower scope.
*Fix:* pin the order with a deterministic unit test in an `answer_tests.rs` sidecar. Split the lock-held tail of `write_answer` into a helper that is generic over `W: Write`, then drive it with a recording writer and assert that `release` ran before the first byte was written. Alternatively, in the test build, record `slots.is_in_flight(&request_id)` immediately before `write_all` and have the regression test assert it was `false`.

**AR-Q13-2 — Consider — readability** — `crates/app/src/control/gateway/server.rs:316`
`ConnectionSlots::new(#[cfg_attr(not(test), allow(unused_variables))] options: &GatewayOptions)`: the production constructor takes the whole options struct only to clone one test-only field, and it needs an `allow` on a parameter to compile, which is a local idiom used nowhere else in the gateway. It costs nothing at runtime.
*Fix:* keep `new()` parameterless and add a `#[cfg(test)]` constructor, or pass `#[cfg(test)] answer_written: Option<AnswerWritten>` explicitly, so the release build's signature does not mention options it ignores.

No other findings. Dimensions checked clean:
- **1 Docking.** The change consolidates seven hand-ordered terminal paths into one function, so a new answer path has one right call to make.
- **2 Performance.** Runs once per control request, not per trade, depth update or frame. One `BTreeSet::remove` moves inside a writer lock that was already taken, and encoding stays outside the lock.
- **3 Hardcoded values.** The only new literal is `100 ms` in a test, and test pacing is derived from `CONTROL_CLIENT_RATE_PER_SECOND`.
- **4 Test layout.** The hook is a `#[cfg(test)]` observation seam that is `None` by default. It follows the existing `descriptor_directory` / `enable_for_test` pattern and changes no production path.
- **5 Standardisation.** `read_call` reuses the module's `unkeyed_call` seam, and #430 reuses #382's `drain_gateway_requests`.
- **6 Readability.** Names carry intent, and the ordering objective is documented in `answer.rs`.
- **7 Second operator.** No surface.
- **8 Language.** See the verdict.
- **9 Trunk.** See the verdict.

The #429 assertion removal was decided by the coordinator (D2) and is documented in `live-envelope.md`, so it gets no severity here. After it, no automated test guards worker keep-up any more; that claim now rests only on the measured `measure.txt` block 4.

### Verification
- Final-head CI at `05426152`: `ci` pass (9m7s), `windows` pass (8m37s), run 34749514291.
- `cargo test -p quantick-guards` run locally in the worktree at `05426152`: all suites ok, 0 failed. This covers the language guard and the size/context ratchets.
- The four workspace checks were not re-run locally: the host is running a CPU-contention measurement, and the coordinator asked for no full workspace builds unless a finding needed one. The evidence for them is the green final-head CI above plus the PR's local run at `d7b93c3c` (the code head; `05426152` adds only the archive).

### Verdict (graded commit `05426152`)
- **Correctness:** step 0 found no production defect. The one test-quality finding (AR-Q13-1) is still open.
- **Docking:** yes. Every terminal answer path now docks through `answer_and_release`, and the next path needs one call, not an ordered pair.
- **Performance:** flat. The answer path runs once per control request, and a set removal moves inside a lock that was already held.
- **Operability:** no surface. There is no user-facing capability and no wire change.
- **Proof:** `a_client_that_reads_its_answer_can_reuse_the_request_id_at_once` is an integration test against a real gateway. It fails if a release moves after the call or if the hook is bypassed, but not if the order inside `write_answer` regresses (AR-Q13-1).
- **Accumulation:** trunk flat. No tracked file moved: `server.rs` went from 1,487 to 1,460 production lines, `gateway.rs` gained 25 test-only lines, and the new `answer.rs` is 88 lines, all under the 1,500 threshold. No ceiling was raised.
- **Language:** the guard passed (`quantick-guards`, CI). I read the diff's prose, the branch name `fix/gateway-request-id-ordering`, all four commit messages and the PR title and body myself: all English. The mission archive's quotation is an English, attributed quotation.

milocaetano:
## AI review - PR #436 @ 0542615245e42bd79b75797742a8017259de1d33
Identity: HEAD `0542615245e42bd79b75797742a8017259de1d33`; branch `fix/gateway-request-id-ordering`; worktree `C:\src\quantick-worktrees\fix-gateway-request-id-ordering`; base ref `origin/campaign/lean-a-plus`; base tip `6ba1a7f3831f08ac5eafdc09f1d1735f00cb23d7`; review key `9f732346cbd2491c5d9bacd892f49319b0efae4b`
Scope: full diff (round one), `git diff origin/campaign/lean-a-plus...HEAD`, 9 files, +535/-129. Findings: 2 (`PRRT_kwDOTfuoRs6h384E` FAIL, `PRRT_kwDOTfuoRs6h387T` WEAK); 2 unresolved, 0 accepted, 0 fixed.

1. Modular:      PASS - `gateway/server/answer.rs` (new, 88 lines) owns one thing, how an answer leaves the gateway: `answer_and_release` (`answer.rs:31`), `send_response` (`:47`), `write_answer` (`:55`). `server.rs` loses its old `send_response` body (+36/-61). Blast radius: 7 pre-existing files edited. The largest is `app/tests/control_plane_tests.rs` (+122/-54); in production it is `control/gateway/server.rs` (97 lines).
2. Decoupled:    WEAK - the release now happens in one place under the writer lock (`answer.rs:38`, `:71-74`). Only a convention ties `slots.track` (`server.rs:1048`, `:1273`) to the helper that must answer that ID, and both helpers are imported side by side (`server.rs:73`). Variant: a new post-track refusal written with `send_response` leaks the ID for the connection's lifetime. Thread `PRRT_kwDOTfuoRs6h387T`. The test hook runs through `GatewayOptions` into `ConnectionSlots::new(&GatewayOptions)` (`server.rs:316`), whose options parameter only a test build reads. That follows the existing `descriptor_directory` pattern and is not a finding.
3. AI-ready:     PASS - no new capability. The fix makes the existing named call safe to reuse by ID: once a client has read an answer, it can resend under the same ID (`answer.rs:22-30`). Every terminal path answers with a typed `known_error` code (`server.rs:994-1003`, `:1031-1041`, `:1107-1118`, `:1296-1307`, `:1335-1346`). Encode failure stays typed (`answer.rs:63-68`, `unencodable`). Every response carries the request's own `request_id` (`failure_response` `server.rs:1428`, `validated_success`), so releasing by `response.request_id` matches the old `envelope.request_id`.
4. Agent-tested: FAIL - `cargo test -p quantick-app a_client_that_reads_its_answer_can_reuse_the_request_id_at_once` runs alone and forces the interleaving with channels, not sleeps, but it sends the reuse only after a hook that fires once `write_answer` has returned (`answer.rs:39-42`). So it passes under the pre-fix order moved into the helper, and under `release()` after `write_all`. Thread `PRRT_kwDOTfuoRs6h384E`. The gateway lives in `app`, so the test cannot sit below it. Error path: `gateway_rejects_a_duplicate_request_id_while_a_wait_is_parked` still covers the refusal. Its single unretried reuse catches a regression only under preemption, so it samples the bug rather than pinning it. Hunt: `Instant`/`recv_timeout` appear only as the 30 s budget ceilings. `reply_pending(100ms)` in the timeout test (`control_plane_tests.rs:1057`) is a time-bounded negative, but after the timeout answer the response worker has no path that writes a second frame, so it cannot mask a defect. The retry pacing now derives from `CONTROL_CLIENT_RATE_PER_SECOND` (`mod.rs:1903`, `retry_readback_tests.rs:710`). `burst.rs` stops asserting `late_frames` (a wall-time claim) but keeps asserting the counts (`deferred == 0`, nothing lost), and still prints `late_frames` (`burst.rs:413-418`).
5. Extensible:   PASS - the next capability docks through the contract registry and never touches the dispatch paths edited here. The diff replaces seven call-site pairs (`send_response` then `slots.forget`) with one function (`answer.rs:31`). The remaining hazard in choosing between the two helpers is graded under 2.
6. Scalable:     PASS - rate: `answer_and_release` runs once per terminal answer, at most `CONTROL_CLIENT_RATE_PER_SECOND` (20/s) per connection. It adds one `BTreeSet` removal (O(log n), n capped by per-connection in-flight plus parked slots) and one `RequestId` clone inside the writer lock that was already held for the write. Nothing reads history and no new buffer is added. The burst rig's per-frame loop is unchanged. Test pacing is per retry, rare.
Top fix: `crates/app/src/control/gateway/server/answer.rs` - observe the ID's in-flight state inside `write_answer` right after `write_all`, still under the writer lock, and assert it is free - flips: 4. A `TrackedId` returned by `ConnectionSlots::track` and consumed by `answer_and_release` flips 2.

milocaetano:
{"authorization": null, "batches": 0, "evidence": ["Round-1 reviews at 05426152: arch-review https://github.com/milocaetano/quantick/pull/436#issuecomment-5652582184, ai-review https://github.com/milocaetano/quantick/pull/436#issuecomment-5652588337; no repair attempts before them."], "findings": {"AR-Q13-1": {"attempts": 0, "disposition": "open", "evidence": ["https://github.com/milocaetano/quantick/pull/436#issuecomment-5652582184"]}, "AR-Q13-2": {"attempts": 0, "disposition": "open", "evidence": ["https://github.com/milocaetano/quantick/pull/436#issuecomment-5652582184"]}, "PRRT_kwDOTfuoRs6h384E": {"attempts": 0, "disposition": "open", "evidence": ["https://github.com/milocaetano/quantick/pull/436#issuecomment-5652588337"]}, "PRRT_kwDOTfuoRs6h387T": {"attempts": 0, "disposition": "open", "evidence": ["https://github.com/milocaetano/quantick/pull/436#issuecomment-5652588337"]}}, "head": "0542615245e42bd79b75797742a8017259de1d33", "limits": {"attempts": 3, "batches": 3}, "mission": "gateway-request-id-ordering", "revision": 1, "schema": 1, "stage": "review", "stalled_batches": 0}

milocaetano:
{"authorization": null, "batches": 1, "evidence": ["Repair batch 1 at 0e464d3e targets AR-Q13-1, PRRT_kwDOTfuoRs6h384E, PRRT_kwDOTfuoRs6h387T; AR-Q13-2 (Consider) not taken, reason in the PR body."], "findings": {"AR-Q13-1": {"attempts": 1, "disposition": "open", "evidence": ["https://github.com/milocaetano/quantick/pull/436#issuecomment-5652582184"]}, "AR-Q13-2": {"attempts": 0, "disposition": "open", "evidence": ["https://github.com/milocaetano/quantick/pull/436#issuecomment-5652582184"]}, "PRRT_kwDOTfuoRs6h384E": {"attempts": 1, "disposition": "open", "evidence": ["https://github.com/milocaetano/quantick/pull/436#issuecomment-5652588337"]}, "PRRT_kwDOTfuoRs6h387T": {"attempts": 1, "disposition": "open", "evidence": ["https://github.com/milocaetano/quantick/pull/436#issuecomment-5652588337"]}}, "head": "0e464d3eb805fbf4a1be067db81c99be87f2d8b5", "limits": {"attempts": 3, "batches": 3}, "mission": "gateway-request-id-ordering", "revision": 2, "schema": 1, "stage": "repair", "stalled_batches": 1}

crates/app/src/control/gateway/server/answer.rs
milocaetano: **Follow-up @ 0e464d3e: still open.** Half of this is fixed. Pre-fix, the release ran after the answer was written. That order now fails deterministically: `drop(in_flight)` after `write_answer` returns leaves the ID in flight when `written()` probes it.

The probe still runs after the frame is written (`answer.rs:130-135`: `write_all`, then `written()`). So the variant this thread names is still green:

```rust
let mut stream = writer.lock()...;
if let Some(frame) = frame && stream.write_all(&frame).is_err() { ... }
release();   // moved below the write
written();
```

With this order `in_flight_when_written` reads `false` and the test passes. It is exactly #425: the frame is on the wire while the ID is still tracked, and the reader's `is_in_flight` (`server.rs:840`) does not take the writer lock. The assertion text, "released before its answer's frame was written", claims a moment the probe never observes.

Fix: probe between `release()` and `write_all`, for example a third closure that `write_answer` calls right before the write, and keep the hold where it is. Then any release that lands after the write fails the test.

crates/app/src/control/gateway/server.rs
milocaetano: **Follow-up @ 0e464d3e: fixed, resolving.** A tracked ID is now an `InFlightId` (`answer.rs:29-48`). `ConnectionSlots::track` has no caller left outside `InFlightId::track`, and both track sites create one (`server.rs:1050`, `:1277`). `answer_and_release` consumes it under the writer lock. Every other exit drops it: a disconnected or closed wait (`server.rs:1296-1297`), and a spawn failure, whose closure is dropped (`:1107`, `:1327`). A parked wait passes its token on to the read it becomes, not tracking the ID a second time (`:1315`). The named variant, a post-track refusal written with `send_response`, can no longer leave the ID held: the token drops at the end of scope.

milocaetano:
## AI review - PR #436 @ 0e464d3eb805fbf4a1be067db81c99be87f2d8b5
Identity: HEAD `0e464d3eb805fbf4a1be067db81c99be87f2d8b5`; branch `fix/gateway-request-id-ordering`; worktree `C:\src\quantick-worktrees\fix-gateway-request-id-ordering`; base ref `origin/campaign/lean-a-plus`; base tip `6ba1a7f3831f08ac5eafdc09f1d1735f00cb23d7` (unchanged since round one); review key `526e9b786ef35f14fce50747c7d016bdaa494d10`
Scope: follow-up. The subject is the round-one open threads, plus a no-new-FAIL check on the delta `0542615..0e464d3` (1 commit, 3 files, +108/-53). Base and scope are unchanged. Findings: 2. Unresolved: 1 (`PRRT_kwDOTfuoRs6h384E`, FAIL). Fixed: 1 (`PRRT_kwDOTfuoRs6h387T`, resolved). Accepted: 0.

1. Modular:      PASS - unchanged from round one. The delta keeps the in-flight token (`InFlightId`, `answer.rs:29-48`) in the file that owns how an answer releases its ID.
2. Decoupled:    PASS - `PRRT_kwDOTfuoRs6h387T` fixed. Every tracked ID is an `InFlightId`, and both track sites create one (`server.rs:1050`, `:1277`). `answer_and_release` consumes it under the writer lock (`answer.rs:58-89`). A disconnected or closed wait drops it (`server.rs:1296-1297`), and so does a failed spawn, whose closure is dropped (`:1107`, `:1327`). A parked wait passes its token on to the read it becomes (`:1315`). A post-track refusal written with `send_response` can no longer hold an ID for the connection's lifetime.
3. AI-ready:     PASS - unchanged. The delta adds no call and no error variant.
4. Agent-tested: FAIL - `PRRT_kwDOTfuoRs6h384E` stays open. The pre-fix order, release after the answer is written, now fails deterministically. The in-flight probe `written()` still runs after `write_all` (`answer.rs:130-135`), though. So `write_all; release(); written();` passes `a_client_that_reads_its_answer_can_reuse_the_request_id_at_once` and reopens #425. Details are in the thread reply.
5. Extensible:   PASS - unchanged.
6. Scalable:     PASS - rate: still once per terminal answer, at most 20/s per connection. The token adds one `Arc` clone and one `RequestId` clone per tracked request (UI-served and parked). The release still happens once, under the lock that was already held. Test-only probes are compiled out of production.
Top fix: `crates/app/src/control/gateway/server/answer.rs` - take the in-flight reading between `release()` and `write_all` (for example a pre-write closure) and keep the hold where it is - flips: 4.

milocaetano:
## arch-review — PR #436, round 2 (delta follow-up)

- **Head reviewed:** `0e464d3eb805fbf4a1be067db81c99be87f2d8b5`
- **Base:** `origin/campaign/lean-a-plus` at `6ba1a7f3831f08ac5eafdc09f1d1735f00cb23d7`. The base and its tip are unchanged since round 1.
- **Review key:** `526e9b786ef35f14fce50747c7d016bdaa494d10`, from `campaign_context.sh key`. It replaces round 1's `9f732346…`.
- **Scope:** this is a delta follow-up under `docs/workflow/delivery.md`, *Review once, follow up on the delta*.
  - Prior report: the round 1 comment at `05426152`.
  - Delta: one commit, `05426152..0e464d3e` ("fix(control): tie a tracked request ID to the answer that releases it"). It touches `gateway/server.rs`, `gateway/server/answer.rs` and `control_plane_tests.rs`, +108/−53.
  - Step 0 ran over the full current PR diff.
- **Tier:** `high`, unchanged.

**step 0: code-review at medium (effort-first, `Skill(code-review, "medium 436")`; no reuse notice in the returned report, and its findings name the delta's `InFlightId`, so it read the current head), 1 finding, 1 confirmed (AR-Q13-3).** Round 2 of at most 3. Step 0 did not re-raise AR-Q13-1; the residual below is the shape pass's own reading of the new probe.

### The delta, checked

- **Deadlock order.** Lock order is still writer lock → `in_flight_ids`. `InFlightId::drop` → `forget` and the test-only `is_in_flight` probe take only the ID lock, either under the writer lock or with no lock held. `InFlightId::track` → `track` takes only the ID lock. The reader's `is_in_flight` never holds the ID lock across a write. No reverse order exists.
- **Drop ordering on the answered paths.** `answer_and_release` moves the token into the `release` closure, which runs under the writer lock before `write_all` (`answer.rs:74`, `:123`). This is the same position as round 1.
- **Parked wait.** The token is created at park (`server.rs:1277`) and moved into the wait thread. On Disconnected or a closed socket, `drop(in_flight)` releases it, as the old `forget` did. On Shutdown, `answer_and_release(Some)` releases it. On Woken or TimedOut, the token is handed to `dispatch_prepared(Some)`, whose three early terminals release it with their answer. The UI path reuses it through `unwrap_or_else` without tracking the ID twice.
- **One token dropped unused.** A `Some` token reaching the `PreparedDispatch::Parked` early return (`server.rs:975`) would be dropped silently. That path is unreachable, because `to_read` always builds a `Ui` dispatch.
- **Spawn-failure paths.** Both now rely on the dropped closure to release the ID; see AR-Q13-3.
- **Latent hazard from round 1 stays closed.** No `forget` runs after an answer is written, so a late release can no longer erase a freshly admitted reuse.
- **ai-review WEAK (decoupling).** The token type now carries the "tracked ⇒ released exactly once" invariant, instead of every call site pairing `track` with `forget`. It is a real improvement.

### Findings

**AR-Q13-1 — Should-fix — test-coverage — still open (partially addressed)** — `crates/app/src/control/gateway/server/answer.rs:135` (the probe), `:123` (release); test `crates/app/src/app/tests/control_plane_tests.rs:2822`
The probe now runs in the `written` closure, which is called after `write_all` and therefore after the release as well. Take the mutation that moves `release();` to just below the `write_all` block, between lines 134 and 135: `lock; write_all; release(); written()`. The probe then reads the ID as already released and returns `in_flight_when_written = false`, so the assertion at `:2822` and the reuse half both pass while #425 is back. The mutation is caught only if the release lands after `written()`. So the probe measures "in flight after release and after the write", not "in flight when the frame went out" as its name and the test's message say.
*Fix:* sample `is_in_flight` immediately before `write_all`, after `release()` — for example a test-only `before_write` closure. Correct order gives `false`. A release anywhere after the write gives `true` and fails deterministically. Then re-run the mutation with `release()` directly below `write_all`.

**AR-Q13-2 — Consider — readability — declined by the owner, accepted.** `server.rs:316`. It has one call site, the `allow` is scoped to a single parameter, and it costs nothing at runtime. Closed with no change.

**AR-Q13-3 — Should-fix — correctness — new in this delta** — `crates/app/src/control/gateway/server.rs:1106-1122` (response worker could not be spawned). The parked-wait variant at `:1326-1349` is not reachable, as explained below.
Round 1 answered this refusal through `answer_and_release`, so the ID was released under the writer lock. Now the `InFlightId` inside the dropped closure releases the ID as soon as `Builder::spawn` fails, with no lock held, and the refusal then goes out through `send_response`.
The race: when this `dispatch_prepared` runs on a parked wait's thread (Woken or TimedOut), the connection's reader keeps running. A duplicate of the ID that the client sent while the wait was unanswered can then be admitted between the drop and the refusal's write. A worker-side duplicate is answered on the reader thread and can reach the wire before the refusal.
That breaks two things: R1's "no duplicate accepted while the first is in flight", and `answer.rs`'s own "answered after this answer, never before it". It also makes the PR body's #425 table wrong: the table's row "response worker could not be spawned" says *release under the lock, write*.
The parked-wait spawn failure runs on the reader thread itself, so no concurrent admission is possible there.
Why Should-fix and not Blocker: the base had the same order (`forget`, then `send_response`), so this only undoes round 1's improvement, and it needs a thread-creation failure to happen at all. The ID is still released; nothing leaks.
*Fix:* keep the token recoverable when spawn fails. Move it into the closure through an `Arc<Mutex<Option<InFlightId>>>` (or an equivalent hand-back). On `Err`, `take()` it and answer with `answer_and_release(writer, codec, refusal, token)`. The release then happens under the lock again, and the table row becomes true again. Otherwise, correct the PR body's row and `answer.rs`'s guarantee to name this path as the exception.

### Carried forward from round 1 (unaffected by the delta)
The `dispatch_prepared` / `dispatch_parked_wait` wiring apart from the token and the two spawn-failure refusals, the #427/#428/#430/#432 test changes, the #429 removal (D2), `docs/quality/live-envelope.md`, dimensions 2, 3, 5, 6, 7 and 9, and the language grading of the first four commits. I re-read the delta's own prose and its commit message: English.

### Verification
- CI at `0e464d3e`: `ci` pass (7m15s) and `windows` pass (6m9s), run 34752013372.
- Run locally: nothing new. No workspace build was done, because the contention run occupies cores 0-3 and the delta's tested inputs are covered by final-head CI.
- Reused from `05426152`: `cargo test -p quantick-guards` (0 failed). The delta adds no file and no tracked-size input; it grows `answer.rs` from 88 to 136 lines, and `server.rs` stays under 1,500.
- The mutation claims above are argued from the code at `answer.rs:120-136`; I did not run them, since the reviewer does not edit tracked files.

### Verdict (graded commit `0e464d3e`)
- **Correctness:** step 0 returned one finding and I confirmed it as AR-Q13-3; no other defect. AR-Q13-3 is open: a spawn-failure refusal on the wait thread now releases the ID outside the writer lock.
- **Docking:** improved. A tracked ID is a token that can only be released once, and a new answer path cannot forget to release it.
- **Performance:** flat. Per control request, one `Arc` clone and one `RequestId` clone move into the token at track time; everything else in the delta is test-only.
- **Operability:** no surface.
- **Proof:** the regression test now catches a release moved past the probe, but still not a release placed directly after `write_all` (AR-Q13-1).
- **Accumulation:** trunk flat. No tracked file moved and no ceiling was raised.
- **Language:** the guard passed in CI at `0e464d3e`. I read the delta's code comments, doc comments and commit message myself: English.

milocaetano:
{"authorization": null, "batches": 2, "evidence": ["Round-2 follow-ups at 0e464d3e: arch https://github.com/milocaetano/quantick/pull/436#issuecomment-5652855378, ai https://github.com/milocaetano/quantick/pull/436#issuecomment-5652842676 (PRRT_kwDOTfuoRs6h387T resolved). Repair batch 2 at 173d3018 targets AR-Q13-1, PRRT_kwDOTfuoRs6h384E, AR-Q13-3."], "findings": {"AR-Q13-1": {"attempts": 2, "disposition": "open", "evidence": ["https://github.com/milocaetano/quantick/pull/436#issuecomment-5652582184"]}, "AR-Q13-2": {"attempts": 0, "disposition": "accepted", "evidence": ["https://github.com/milocaetano/quantick/pull/436#issuecomment-5652855378"]}, "AR-Q13-3": {"attempts": 1, "disposition": "open", "evidence": ["https://github.com/milocaetano/quantick/pull/436#issuecomment-5652855378"]}, "PRRT_kwDOTfuoRs6h384E": {"attempts": 2, "disposition": "open", "evidence": ["https://github.com/milocaetano/quantick/pull/436#issuecomment-5652588337"]}, "PRRT_kwDOTfuoRs6h387T": {"attempts": 1, "disposition": "fixed", "evidence": ["https://github.com/milocaetano/quantick/pull/436#discussion_r3999418444"]}}, "head": "173d30189f2a2ddd9d3e4c85e25fc8d0f94bca2f", "limits": {"attempts": 3, "batches": 3}, "mission": "gateway-request-id-ordering", "revision": 3, "schema": 1, "stage": "repair", "stalled_batches": 0}

crates/app/src/control/gateway/server/answer.rs
milocaetano: **Follow-up @ de270b83: fixed, resolving.** The probe is now `before_write` (`answer.rs:82-87`). It runs under the writer lock between `release()` and `write_all` (`write_answer`, `answer.rs:130-131`). A release that lands anywhere after the write makes `slots.is_in_flight` read `true` at the probe, and `a_client_that_reads_its_answer_can_reuse_the_request_id_at_once` fails with "the ID was released before its answer's frame was written". That covers both variants this thread named: the pre-fix order after `write_answer` returns, and `release()` directly below `write_all`. The test can no longer pass with the frame on the wire while the ID is still tracked. To slip past it, the probe itself would have to move, which is an edit to the test's instrument, not to the fix.

milocaetano:
## AI review - PR #436 @ de270b8312f529581dee8aba4da57f5d084dfd29
Identity: HEAD `de270b8312f529581dee8aba4da57f5d084dfd29`; branch `fix/gateway-request-id-ordering`; worktree `C:\src\quantick-worktrees\fix-gateway-request-id-ordering`; base ref `origin/campaign/lean-a-plus`; base tip `6ba1a7f3831f08ac5eafdc09f1d1735f00cb23d7` (unchanged since round one); review key `11edbdf5cd62cca5bc75fa41890d7975f5a1a76e`
Scope: follow-up. The subject is the one open thread, plus a no-new-FAIL check on the delta `0e464d3..de270b8`: `173d3018` changes code in 3 files, and `de270b83` is prose in the goal archive only. Base and scope are unchanged. Findings: 2. Unresolved: 0. Fixed: 2 (`PRRT_kwDOTfuoRs6h384E` this round, `PRRT_kwDOTfuoRs6h387T` last round). Accepted: 0.

1. Modular:      PASS - unchanged. `InFlightId::take` (`answer.rs:43-48`) stays with the token it serves.
2. Decoupled:    PASS - unchanged from the last round. A response worker now takes its token from a hand-over slot (`server.rs:1051-1054`, `:1096`). A worker that never started leaves the token there, and the refusal it would have sent releases it under the writer lock (`server.rs:1114-1126`). The ID is no longer released outside the lock when the closure is dropped.
3. AI-ready:     PASS - unchanged. No new call and no new error. The spawn-failure refusal keeps its typed `control.backpressure`.
4. Agent-tested: PASS - `PRRT_kwDOTfuoRs6h384E` fixed. The `before_write` probe runs under the lock between `release()` and `write_all` (`answer.rs:130-131`, `:82-87`). So `a_client_that_reads_its_answer_can_reuse_the_request_id_at_once` fails whenever the release lands after the write. `gateway_rejects_a_duplicate_request_id_while_the_first_is_in_flight` now serves its first request with `drain_gateway_requests` rather than one `run_frame` (`control_plane_tests.rs:2408-2410`). Its assertions are unchanged, and it follows the #430 pattern from round one.
5. Extensible:   PASS - unchanged.
6. Scalable:     PASS - rate: once per UI-served request, at most 20/s per connection. The hand-over adds one `Arc<Mutex<Option<InFlightId>>>` allocation, taken once by the worker, or by the spawn-failure path, where there is no contention. No history reads, no new buffer.
Top fix: none - no open findings.

milocaetano:
## arch-review — PR #436, round 3 (delta follow-up, last step-0 round)

- **Head reviewed:** `de270b8312f529581dee8aba4da57f5d084dfd29`
- **Base:** `origin/campaign/lean-a-plus` at `6ba1a7f3831f08ac5eafdc09f1d1735f00cb23d7`, unchanged.
- **Review key:** `11edbdf5cd62cca5bc75fa41890d7975f5a1a76e`, from `campaign_context.sh key` at `de270b83`. It replaces round 2's `526e9b78…`.
- **Scope:** a delta follow-up under `docs/workflow/delivery.md`; the prior report is round 2's comment at `0e464d3e`. The delta is two commits:
  - `173d3018` — "fix(control): sample the ID before the write, and release it with the refusal a failed worker sends". It touches `answer.rs`, `server.rs` and `control_plane_tests.rs`, +32/−19.
  - `de270b83` — "docs: ledger the review repairs and the sixth load-sensitive test". It adds S5 and S6 to the mission archive only: prose, no code.
- **Tier:** `high`, unchanged.

**step 0: code-review at medium (effort-first, `Skill(code-review, "medium 436")`; no reuse notice in the returned report, and it names the `InFlightId::take(&handoff)` hand-back, so it read the round-3 code), 0 findings.** Round 3 of 3, the last allowed step-0 round. It raised one non-blocking note, not a finding: a failed assertion before `release.send(())` in the #425 test leaves the held thread blocked for up to `GATEWAY_TEST_WAIT`. That slows a failing run and cannot change a result.

### Targets verified

- **AR-Q13-1 — resolved.**
  - `write_answer` now runs `release()`, then `before_write()`, then `write_all`, all under the writer lock (`answer.rs:130-139`). The probe is the `before_write` closure in `answer_and_release`.
  - Correct order: the probe reads `false`.
  - A release moved anywhere after `write_all`: the probe reads `true`, and `control_plane_tests.rs` asserts `!in_flight_when_written`. The owner reports running exactly that mutation (release directly below `write_all`) and seeing the test fail with the expected message.
  - The hold hook still fires after `write_answer` returns and the lock is released, so the reuse half is unchanged.
  - The one order the probe cannot see is a release moved out from under the lock. That is not the #425 failure, and `answer_and_release`'s signature makes it hard to write by accident.
- **AR-Q13-2 — closed in round 2** (Consider, declined by the owner).
- **AR-Q13-3 — resolved.**
  - `dispatch_prepared` moves the token into `Arc<Mutex<Option<InFlightId>>>` (`server.rs:1053`). The worker takes it only when it answers (`:1096`).
  - On `spawn.is_err()` the dropped closure releases only its `Arc` clone, and the caller takes the token back with `InFlightId::take(&handoff)`. It then answers through `answer_and_release` (`:1111-1127`), so the ID is released under the writer lock before the refusal is written.
  - This restores the order the PR body's table gives for that row, and it closes the wait-thread window.
  - `InFlightId::take` drops its guard before returning, so the hand-over mutex is never held while the writer lock is taken. No new lock order appears.
  - If the worker panics before taking the token, dropping the `Arc` releases the ID. Nothing leaks.
  - The parked-wait spawn failure (`:1330`) still releases by drop and answers with `send_response`. That runs on the reader thread, so no duplicate can be admitted in between, as graded in round 2.

### The delta, checked for regressions

- `gateway_rejects_a_duplicate_request_id_while_the_first_is_in_flight` (`control_plane_tests.rs:2408`): one `run_frame` became `drain_gateway_requests`, #430's helper. Every assertion is kept: the duplicate is refused with `control.invalid_request` before the drain, and the first request is then served. The test still guards the rule R1 keeps. It is a stated detour (S5).
- The doc comments on `InFlightId`, `AnswerWritten`, `write_answer` and the #425 test now match the code.
- S5 and S6 in the mission archive are English and accurate to the diff.

No new findings.

### Carried forward (unaffected by the delta)
- Round 1's conclusions on the ordering change, the lock order, the #427/#428/#429/#430/#432 test changes, and dimensions 2, 3, 5, 6, 7 and 9.
- Round 2's on the `InFlightId` token and the parked-wait handoff.

### Verification
- CI at `de270b83`: `ci` pass (8m35s) and `windows` pass (8m40s), run 34754351456, `headSha` de270b83, conclusion success.
- Run locally: nothing new. There was no workspace build, because contention rounds occupy cores 0-3.
- The `quantick-guards` result is reused from `05426152`. The delta adds no file; `answer.rs` is at 143 lines and `server.rs` stays under the 1,500-line threshold.
- `de270b83` is a prose-only commit on top of `173d3018`, which step 0 graded. Its evidence is covered by the evidence-only row of the delivery contract.

### Verdict (graded commit `de270b83`)
- **Correctness:** step 0 returned 0 findings at medium in round 3. AR-Q13-1 and AR-Q13-3 are resolved; no Blocker or Should-fix is open.
- **Docking:** a tracked ID is a token that can only be released once, and every answering path releases it through `answer_and_release` under the lock. The next path needs one call.
- **Performance:** flat. Per control request, the delta adds one `Arc<Mutex<Option<_>>>` allocation on the UI-dispatch path, next to the thread spawn that path already pays for. Nothing per trade, depth update or frame.
- **Operability:** no surface.
- **Proof:** `a_client_that_reads_its_answer_can_reuse_the_request_id_at_once` is an integration test against a real gateway. It fails if the release moves after the write, and fails if the ID is held when the client reuses it.
- **Accumulation:** trunk flat. No tracked file moved and no ceiling was raised.
- **Language:** the guard's CI verdict at this head is given above. I read the delta's comments, the S5/S6 prose and both commit messages myself: English.

**Result: pass.** No open Blocker or Should-fix at `de270b83`. Final-head CI is green, so the mission owner may record `arch-review-ok` for key `11edbdf5…`.

milocaetano:
{"authorization": null, "batches": 2, "evidence": ["Round-3 follow-ups at de270b83: arch https://github.com/milocaetano/quantick/pull/436#issuecomment-5653030119 (pass, 0 step-0 findings), ai https://github.com/milocaetano/quantick/pull/436#issuecomment-5653024672 (six PASS, 0 open threads)."], "findings": {"AR-Q13-1": {"attempts": 2, "disposition": "fixed", "evidence": ["https://github.com/milocaetano/quantick/pull/436#issuecomment-5653030119"]}, "AR-Q13-2": {"attempts": 0, "disposition": "accepted", "evidence": ["https://github.com/milocaetano/quantick/pull/436#issuecomment-5652855378"]}, "AR-Q13-3": {"attempts": 1, "disposition": "fixed", "evidence": ["https://github.com/milocaetano/quantick/pull/436#issuecomment-5653030119"]}, "PRRT_kwDOTfuoRs6h384E": {"attempts": 2, "disposition": "fixed", "evidence": ["https://github.com/milocaetano/quantick/pull/436#discussion_r3999492056"]}, "PRRT_kwDOTfuoRs6h387T": {"attempts": 1, "disposition": "fixed", "evidence": ["https://github.com/milocaetano/quantick/pull/436#discussion_r3999418444"]}}, "head": "de270b8312f529581dee8aba4da57f5d084dfd29", "limits": {"attempts": 3, "batches": 3}, "mission": "gateway-request-id-ordering", "revision": 4, "schema": 1, "stage": "review", "stalled_batches": 0}

milocaetano:
## delivery-review (full, tier `high`): **PASS**

- **Head:** `de270b8312f529581dee8aba4da57f5d084dfd29`
- **Base:** `origin/campaign/lean-a-plus` @ `6ba1a7f3`
- **Review key:** `11edbdf5cd62cca5bc75fa41890d7975f5a1a76e`
- **Scope:** the full diff `BASE...HEAD`, 9 files, +625/-139.
- **Checklist source:** 1, `.claude/GOAL-archive-gateway-request-id-ordering.md` (R1..R11, D1..D5, S1..S6, A1..A9, G1..G6, C1..C3, verbatim request), plus issues #434, #425, #427, #428, #429, #430 and #432.
- **Mode:** full. The completeness pass ran on the strong model. The criteria pass ran on `sonnet`, and the one line it did not grade DELIVERED was escalated to the strong model.
- **Bracket:** HEAD, working tree and review markers were identical before the first dispatch and after the last verdict.

### Completeness pass (source first)
I derived the distinct asks from the verbatim coordinator prompt and #434, #425 and #427..#432, then compared them with the map:

- **Ordering fix and kept guarantees:** R1.
- **Real-gateway regression test:** R2.
- **Removal of #411's wait:** R3.
- **Wire-change stop trigger:** R4 / D1.
- **Forcing the five interleavings:** R5.
- **No weakened assertion:** R6.
- **#429 counts, measured timing claim, and saying so in the PR:** R7.
- **At least 10 contention rounds with a before/after table:** R8.
- **Skip-list lines**, including #427's "a fix must remove the test from any contention skip list": R9. Removal by whichever of #414 and this PR lands second is the #434 scope wording.
- **Purpose:** R10.
- **Integration with read-back:** R11 / C2. The merge itself is the coordinator's.
- **Operational instructions** (tier reviews, readiness once cd-prefixed, never merge or touch main, environment notes, trailers, handoff): D4, D5, G3, G6, C2 and C3.

**UNLEDGERED: none.** #438 (the discovery-publish timeout on five other tests) is a new exposure outside the request, filed separately. It is not an omission.

### Ledger pass
R1..R10: **COVERED**, each discharged by a DELIVERED criterion (A1..A8). R11: **COVERED** for this child's part (A9, the PR base). The campaign merge and its read-back belong to the coordinator (C2), not to this candidate.

### Criteria pass (merged grades)
| ID | Grade | Evidence |
| --- | --- | --- |
| A1 | DELIVERED | `control/gateway/server/answer.rs` `write_answer`: release, then probe, then `write_all`, all under the writer lock. Every tracked terminal path in `server.rs` goes through `answer_and_release` / `InFlightId`; the no-answer paths release by drop. The global slot is still returned after the write. |
| A2 | DELIVERED | `a_client_that_reads_its_answer_can_reuse_the_request_id_at_once` holds the writer after the write, asserts the release came before the frame, reuses the ID and asserts `Success`. `old-order-regression-q13.txt` records its failure on the old order. |
| A3 | DELIVERED | `gateway_rejects_a_duplicate_request_id_while_a_wait_is_parked`: retry loop removed, one send, `Success` asserted. |
| A4 | DELIVERED | No file under `crates/control/` is touched. CI at head is green; the local `workspace-test-4.log` has 0 failed. |
| A5 | DELIVERED | No `assert*` is removed from the #427, #428, #430 or #432 bodies. The forcing is `pause_one_request_interval`, `read_call`, `drain_gateway_requests`. #430 gains a `!reply_pending` assertion. |
| A6 | DELIVERED | `burst.rs` drops the `late_frames` bound and keeps the counts. `docs/quality/live-envelope.md` says the test asserts counts only and that the keep-up claim is measured. |
| A7 | DELIVERED | Reproduced with `summarize_q13.py`: before, #428 71/72, #432 72/72 and #429 5/72 fail. After, 10 rounds of 24 copies: #411, #425, #427, #428, #429, #430 and #432 each 240/240 ok. The S5 test, grepped directly (it is not in the script's map), is 240/240 ok. |
| A8 | DELIVERED | The PR body lists the five skip-list lines. |
| A9 | DELIVERED | `baseRefName` = `campaign/lean-a-plus`. |
| G1 | DELIVERED | Every authored line of the diff and every commit message is English. |
| G2 | DELIVERED | CI jobs `ci` (fmt, clippy, build, test) and `windows` succeed at `de270b83`. The performance impact is declared in the PR body. |
| G3 | DELIVERED | arch-review round 3 is **pass** (https://github.com/milocaetano/quantick/pull/436#issuecomment-5653030119). ai-review final shows 0 unresolved (https://github.com/milocaetano/quantick/pull/436#issuecomment-5653024672). `ai_review_threads.sh count 436` = 0. The progress record is revision 4: 2 of 3 batches, no stalls, every finding fixed or accepted. |
| G4 | DELIVERED | Run 34754351456: `headSha` is `de270b83`, conclusion `success`. |
| G5 | DELIVERED | The files are the owned set plus the named detours. Nothing touches `state*`, `tab/*`, `ci.yml` or `tools/ci/*`. |
| G6 | DELIVERED *(escalated; first pass PARTIAL)* | Both trailers are on all 7 commits. Every scratchpad path of this mission carries `-q13`. The four checks are itemised separately, and each of the four workspace-test logs is its own run. `cargo check` and `df` before the first edit are recorded in the mission record at `05426152` ("268 GB free"), which predates the reviews, and nothing contradicts it. That is an operational obligation whose only possible trace is the record. Readiness through the Bash tool is not due until C2. |

**Differing readings:** on G6, sonnet read the absence of raw `cargo check`/`df` output as PARTIAL. The strong pass graded it DELIVERED: the obligation is observable only through the record, the record is specific and contemporaneous, and no evidence contradicts it.

### Closing steps
- **C1:** this verdict (PASS).
- **C2** (readiness against `campaign/lean-a-plus`) and **C3** (handoff): pending owner steps, not yet due. C2 still has to show that `gh pr ready 436` ran once, cd-prefixed, from the Bash tool and never through PowerShell.

### Notes (not findings)
- **The old-order evidence has two captures:**
  - Mutation 2 (`control_plane_tests.rs:2824:5`) matches the `assert!` of "the ID was released before its answer's frame was written" in the test file at `173d3018`.
  - Mutation 1 (`:2834:5`) does not match any committed state of that file, so it was captured on an intermediate working tree. Its panic text matches the reuse assertion exactly.

  The claim "fails on the old order" therefore rests on the mutation-2 capture at the final test file.
- **`gateway/server/answer.rs`** is a new child module (`mod answer;` in the owned `server.rs`). The mission names it only in S6, not in G5's owned list. It stays inside the owned tree, so the record is adequate.
- **S1:** "at least as contended as CI" is argued, not measured against the Linux harness. CI's contention step (#414) is not yet on this base.

### What could have failed and did not
- The A7 counts were recomputed from the raw logs, and the S5 test was grepped outside the script's map. The before, after and #438 counts all match the PR table.
- The wire crate was checked by path rather than by the PR's prose.
- Every changed test body was diffed for removed assertions; none were removed.
- The CI job was confirmed to run the four checks by reading `ci.yml`, and CI `headSha` equals HEAD.
- The trailers were read from `git log`, not from the PR body.
- The head-only delta `173d3018..de270b83` was confirmed to be the goal archive alone, so the local evidence reuse at `173d3018` is sound.

**Counts:**
- UNLEDGERED 0
- R: 11 COVERED
- A: 9/9 DELIVERED
- G: 6/6 DELIVERED (1 escalated)
- UNPROVEN, MISSING, PARTIAL: 0
- Findings: none.

PR #547, opened 2026-09-19
feat(ci): fast affected-crates job on drafts, full CI gates readiness

milocaetano:
Give agents a fast CI signal on draft PRs while keeping full final-head CI mandatory for readiness.

Objective: Draft PRs get a fast affected-crates CI job; full `ci` + `windows` stay mandatory at the exact head for `gh pr ready`, merge and the final verifier.
Tier: medium
Source: retained user request, 2026-09-19 (verbatim in the goal below)
Criteria: A1-A9 delivered (A6 final proof is the verifier PASS on this PR); G1-G4 delivered; G-AI1-4 delivered (0 open threads)
Validation: local fmt/clippy/build/test green (7f50c1ca; later commits touch only tools/ci, ci.yml, hooks tests); guardrails_test.sh 292 passed 0 failed @ 78c2c37f; campaign_context_test.sh 69 passed; test_affected_crates.py 14 OK; cargo test -p quantick-guards green; full CI green at 5fb3d128 (run 35447339152) and at head 78c2c37f (run 35448369981)

## What changed

- `.github/workflows/ci.yml`: new `fast` job (draft pushes only) runs `cargo fmt --all -- --check`, then `cargo clippy --all-targets` and `cargo test` over `-p` flags from `tools/ci/affected_crates.py`. `ci` and `windows` run on push to `main`, `ready_for_review`, pushes to non-draft PRs, and drafts labelled `full-ci`. The flip always re-runs full CI: skipping it for a labelled draft was tried on this PR and made `gh pr checks` show only skipped `ci`/`windows`, hiding the green run (fixed in 78c2c37f).
- `tools/ci/affected_crates.py`: owner crate by manifest dir + crates naming the changed file by path (one `git grep -F -f` pass), closed over reverse normal/dev/build deps from `cargo metadata`; `quantick-guards` always; workspace inputs (`Cargo.toml`, `Cargo.lock`, `rust-toolchain.toml`, `.cargo/`, fmt/clippy config) select everything. Tests compare every crate's closure with `cargo tree --invert`.
- `.claude/hooks/full_ci.sh`: one definition of "full CI green at a commit" — newest non-skipped GitHub Actions `ci` and `windows` runs at that SHA concluded `success`; exit 2 (unknown) becomes `ask`, never a pass.
[...]

PR #548, opened 2026-09-19
demo: engine-only change to measure #547's fast CI job (do not merge)

milocaetano:
Throwaway draft stacked on #547. It changes one doc comment in `crates/engine/src/lib.rs` so #547's `fast` job can be measured on an engine-only diff and its selected crates compared with `cargo tree -i quantick-engine`. Closed unmerged once measured.


milocaetano:
Measured; evidence is in #547's body. Closed unmerged as planned.

PR #550, opened 2026-09-19
feat(read-cost): make per-PR read cost visible and bounded

milocaetano:
Make per-PR read cost visible and bounded: the measurement CI already takes becomes a sticky comment, a committed ledger, a ceiling drawn from history, and a `pr-gate` line that warns and never blocks.

Objective: Surface the read-cost measurement CI already produces — a sticky PR comment, a committed ledger of merged PRs, a ceiling derived from history, and a non-blocking pr-gate warning — so an author sees the number and the trend is on record.
Tier: medium
Source: Retained user request of 2026-09-19, quoted verbatim in the goal file below (no issue).
Criteria: A1 delivered, A2 delivered (under D3), A3 delivered, A4 delivered, A5 delivered, A6 delivered, G1 delivered, G2 delivered, G3 delivered, G-AI1..G-AI4 delivered. Nothing deferred, nothing open.
Validation: at head e89d4df3, rebased onto main bafba853 — cargo fmt --all -- --check, cargo clippy --workspace --all-targets, cargo build --workspace, env -u QUANTICK_BUBBLES cargo test --workspace: all green, no failures; sh .claude/hooks/guardrails_test.sh: 312 passed, 0 failed, plus campaign_context 8 passed, 0 failed; python -m unittest discover -s tools/read_cost -p 'test_*.py': 44 passed; python -m ruff check --select F tools/ bridge/mt5/: clean; every CI job green at this exact head (ci, windows, windows-tests, windows-app-tests, harness-app, harness-scenario, harness-combined, fast, read-cost), and the sticky comment on this PR was posted once and updated in place across every push.
Read cost: 0 production lines — this branch changes no Rust file, so the calculator reports `not_applicable`.
Reviews: arch-review PASS https://github.com/milocaetano/quantick/pull/550#issuecomment-5746986041 · ai-review COMPLETE https://github.com/milocaetano/quantick/pull/550#issuecomment-5746986450 · delivery-review PASS (completeness-only, the medium tier's mode) https://github.com/milocaetano/quantick/pull/550#issuecomment-5746986924

## What lands

- **A sticky comment per pull request.** `tools/read_cost/report.py comment` renders `production_lines`, the changed-file count and the top five files the change pulls in *by reference alone*; CI updates one comment in place, keyed by the marker the renderer itself prints. It also asks for the ledger row while that row is missing.
[...]

PR #562, opened 2026-09-20
fix(hooks): let final completion tell a refusal from an advisory

milocaetano:
Objective: Let the final verifier tell a readiness refusal from a readiness advisory, so a branch over the read-cost ceiling can still complete a mission.
Tier: small
Source: found by PR #561, whose mission completion the advisory blocked; the trader chose to fix it in its own PR.
Criteria: A1 delivered (only a `deny` decision refuses), A2 delivered (the advisory path is proven by a test that fails without the change), A3 delivered (no other gate behaviour moves).
Validation: `sh .claude/hooks/guardrails_test.sh` — 318 passed, 0 failed with the fix; the two new cases fail without it (output in the PR body).

## The defect

`mission_ship_gate.sh` reuses `pr-gate` rather than reimplementing the
readiness policy, which is right — but it read *any* output from it as a
refusal:

```sh
readiness=$(printf '%s' "$payload" | sh "$script_dir/guardrails.sh" pr-gate 2>&1)
[ -z "$readiness" ] || fail "The shared readiness gate refused this PR: $readiness"
```

The same gate also carries advisories. The read-cost ceiling added in #550
prints one on stdout, as a `systemMessage` plus `additionalContext` with **no**
`permissionDecision`, and its own text says *"Nothing is blocked"* —
`docs/quality/read-cost/ledger.md` states it outright: "It is a warning and
never a block."

So since that ceiling landed, every feature branch over 18,000 production lines
of read cost could pass every review, every marker and full CI, and still not
complete a mission. PR #561 is the first branch over it:

```
MISSION-COMPLETION:FAIL The shared readiness gate refused this PR:
{"systemMessage":"Read cost: this branch asks a reader to hold 27148 production
lines, over the 18000 ceiling for feature branches in
docs/quality/read-cost/ledger.md. Nothing is blocked. …"}
```

A real refusal looks different, and the difference is the fix's whole content:

```
{"hookSpecificOutput":{…,"permissionDecision":"deny","permissionDecisionReason":"…"}}
```

## The change

One `case` in place of the `-z` test: a refusal is a `deny` decision, and only
that. Nothing else about the gate moves — the marker checks, the thread list,
the CI verdict and the published reconciliation are untouched.

## Proof

Two cases in `guardrails_test.sh`, beside the existing completion cases, driving
the real `mission_ship_gate.sh` through the harness's stubs:

- a readiness advisory carrying no decision **does not** block completion;
- a readiness denial **still** blocks it.

Without the fix the first fails and the second passes; with it both pass.

```
# with the fix
318 passed, 0 failed

# without it
FAIL a readiness advisory carrying no decision does not block completion: expected pass containing "", status=1
  output: MISSION-COMPLETION:FAIL The shared readiness gate refused this PR: {"systemMessage":"Read cost: over the ceiling. Nothing is blocked."…}
317 passed, 1 failed
```

## Blast radius

`.claude/hooks/mission_ship_gate.sh` (one test replaced by a `case`, plus the
comment saying why) and `.claude/hooks/guardrails_test.sh` (two cases). No Rust,
no skill, no baseline.


milocaetano:
# Architecture review — PR #562, `fix/ship-gate-reads-advisories` @ `44cdc8e6`

step 0: code-review at low (effort-first, no reuse notice), 1 finding, 1
confirmed and fixed. Level `low` is what the `small` tier buys, read from
`.git/worktrees/fix-ship-gate-advisories/mission-tier`
(`fix/ship-gate-reads-advisories small`).

Scope: `git diff origin/main...HEAD` at `4b5e2b7e...44cdc8e6` — 3 files, 59
changed lines: `.claude/hooks/mission_ship_gate.sh`,
`.claude/hooks/guardrails_test.sh` and one row in
`docs/quality/read-cost/ledger.md`. The diff is a shell script and its tests,
so it takes the full shape pass rather than the prose waiver; every dimension
below was read, and most are reached only trivially.

## Step 0 findings

**Confirmed and fixed — `.claude/hooks/mission_ship_gate.sh:57`.** The first
commit replaced the empty/non-empty test with a `deny` match, and that test was
also the only thing catching a readiness gate that *broke*: a stack trace, a
missing file or a non-zero exit would have passed silently. Fixed in
`44cdc8e6`, which keeps three failing shapes — a `deny` decision, a non-zero
exit (`could not be evaluated`), and output that is neither a decision nor an
advisory (`neither a decision nor an advisory`) — with a test case for each.
The reviewer's suggestion was the exit status; the third shape is the same
argument applied to a gate that falls over while exiting zero.

## Shape findings

None at Blocker or Should fix.

**Consider — the advisory is recognised by `"hookEventName":"PreToolUse"`.**
That is the one field every hook payload in `guardrails.sh` prints, deny and
ask alike (`guardrails.sh:203,221`), so it is the honest marker of "this is a
hook payload, and it carries no decision". It is a substring match rather than
a parse, which is what the rest of the repo does with these payloads — 
`guardrails_test.sh:201,205,1115` and `campaign_context_test.sh:81` all match
`'"permissionDecision":"deny"'` the same way, and the emitter is a `printf`
with a fixed shape two files away. A JSON parser here would be the second way
to do a solved thing, which dimension 5 refuses.

## Evidence

**The behaviour is proved by tests that fail without the change.** The harness
drives the real `mission_ship_gate.sh` through its own stubs, so these are
behavioural, not textual:

```
with the change:     320 passed, 0 failed
without commit 1:    317 passed, 1 failed
                     FAIL a readiness advisory carrying no decision does not
                     block completion
```

The three cases added: an advisory carrying no decision does not block; a
denial still blocks; a gate that exits non-zero blocks; output that is neither
blocks.

`sh .claude/hooks/guardrails_test.sh` — 320 passed, 0 failed, at this head.
The four cargo checks are untouched territory (no Rust in the diff) and run in
CI.

## Verdict

**Correctness** — step 0's finding was real and is closed by the second commit,
which restores the property the first one dropped. Nothing open. The failure
mode that motivated the PR is itself proven: PR #561 passed every review,
marker and full-CI check and could not complete, because the read-cost
advisory — a warning `docs/quality/read-cost/ledger.md` calls "never a block" —
was read as a refusal.

**Docking** — unchanged. The gate still reuses `pr-gate` rather than
reimplementing the readiness policy; what changed is how it reads that gate's
answer. No new port, and none owed.

**Performance** — irrelevant: one `case` in a script that runs once per
completion attempt. Rate: rare.

**Operability** — no surface. No capability, action or registry entry is added;
the verifier is already a named script both workflow skills call.

**Proof** — `guardrails_test.sh`'s four new cases, each runnable as part of `sh
.claude/hooks/guardrails_test.sh`, which CI runs. The advisory case fails
without the first commit; the two breakage cases fail without the second (they
were written against it).

**Accumulation** — the trunk did not grow in any ratcheted sense: no Rust, so
`crates/app` and the size ratchet are untouched, and `.claude/hooks/*.sh` is
outside the context ratchet, which covers Markdown. The gate gained 18 lines,
12 of them the comment saying why a decision and an advisory are different
things.

**Language** — the guard covers Rust and is unaffected here; I read the prose
myself: branch name, three commit messages, PR title and body, both new
comments and all four test names are English.

ARCH-REVIEW: PASS

milocaetano:
## AI review - PR #562 @ 44cdc8e63f7f647595e4e881cd75e640d07c186a

Identity: HEAD `44cdc8e63f7f647595e4e881cd75e640d07c186a`; branch
`fix/ship-gate-reads-advisories`; worktree
`C:\src\quantick-worktrees\fix-ship-gate-advisories` (clean); base ref `main`;
base tip `4b5e2b7e4c7ed3ed2e94d38e808f5c835e1185f9`; review key as recorded by
the shared producer at publication.

Scope: full diff, round one. 3 files, 59 changed lines. Findings: 0 FAIL,
0 WEAK; no threads opened.

1. **Modular: PASS** — the change stays inside the one function-sized region it
   belongs to, `mission_ship_gate.sh:61-75`, and the file's name says what it
   is. Blast radius: `.claude/hooks/mission_ship_gate.sh`, 18 lines, of which
   12 are the comment explaining the distinction; the test file gains 4 cases
   and nothing else moves.

2. **Decoupled: PASS** — the gate still asks `pr-gate` for the readiness
   verdict rather than reimplementing markers, tiers or thread counting
   (`mission_ship_gate.sh:61`); the diff changes only how the answer is read.
   It depends on the payload shape `guardrails.sh:203` prints, which is the
   same coupling every other reader in the repo already has
   (`guardrails_test.sh:201`, `campaign_context_test.sh:81`) — one shape, one
   emitter, four readers, no new one invented here.

3. **AI-ready: PASS** — the behaviour is a named call with a typed result
   already: `sh .claude/hooks/mission_ship_gate.sh mission <pr>` exits non-zero
   with a reason on the failure paths and prints `MISSION-COMPLETION:PASS` on
   the success path, and the diff adds two distinct failure reasons rather than
   a single opaque one (`could not be evaluated`, `neither a decision nor an
   advisory`). Calling it twice is safe: it reads state and publishes an
   idempotent report.

4. **Agent-tested: PASS** — `sh .claude/hooks/guardrails_test.sh` runs alone,
   below any crate, with no clock, no randomness and no network: the `gh`,
   `review_report.sh` and now `guardrails.sh` stubs are files in a temporary
   root. The named cases are "a readiness advisory carrying no decision does
   not block completion", "a readiness denial still blocks completion", "a
   readiness gate that exits non-zero blocks completion" and "readiness output
   that is neither decision nor advisory blocks completion". The first fails
   without the first commit — measured, 317 passed / 1 failed — and the
   expectations come from the ledger's own prose, not from the code under test.

5. **Extensible: PASS** — the next decision kind is added, not patched in: the
   `case` has an arm per shape, and `ask` (which `guardrails.sh:221` can
   already emit) falls into the advisory arm today and would take its own arm
   the day completion must treat it differently. No enum, no positional list.

6. **Scalable: PASS** — rate: rare. One `case` on a string, once per completion
   attempt, in a script that already shells out to `gh` several times. No
   buffer, no loop, no growth with session length.

Top fix: none. The one finding of the round came from the bug pass, is
confirmed, and is fixed in `44cdc8e6` with its own two test cases.

AI-REVIEW: COMPLETE

Issue #567, opened 2026-09-21
docs(agentic): require every velocity change to prove its saving against a binding quality gate


milocaetano:
Campaign parent: https://github.com/milocaetano/quantick/issues/563
Task key: T4. Base: `origin/campaign/mission-velocity`. Depends on T2.

## Context

The campaign's own failure mode is reaching the speed target by looking less:
fewer review rounds, a thinner diff read, a gate skipped. That buys hours now
and costs them back in defects. The protocol that decides whether a change
stays has to exist before the first change lands, and quality has to be a gate
rather than a note.

## Scope

In: the written protocol and the mechanism that enforces it.

Out: the individual experiments, which belong to the fix children.

## Acceptance criteria

- [ ] The protocol states how a change proves its saving: on a real mission,
      against the T2 baseline, with the T1 harness, and what sample makes the
      result reportable.
- [ ] The revert rule is explicit and operative: a change that does not prove a
      saving is reverted from the campaign branch, not kept on faith.
- [ ] The quality counter-metric is binding and named: findings per review,
      `cargo run -p quantick-guards -- --report` and green CI must hold or
      improve, and a speed gain bought by reviewing less is recorded as a
      failure.
- [ ] Experiments are serialized, or the protocol says how concurrent changes
      are disentangled, so the campaign does not confound its own results.
- [ ] The protocol is enforced by something a later session reads — a gate, a
      checklist in the campaign record, or a hook — not by this issue alone.

milocaetano:
**D8 amends this issue's third acceptance criterion.** See the decision on #563 — the charter's C5 changed after this issue was written.

The criterion currently reads that the quality counter-metric is binding and that a speed gain bought by reviewing less is recorded as a failure. That is no longer the campaign's position. Replace it when you pick this up:

- **Non-negotiable:** green CI and `cargo run -p quantick-guards -- --report` hold or improve.
- **Priced, not protected:** review depth and process ceremony. Reducing them is an authorized lever.
- **Still a failure:** a reduction chosen on feel instead of on measurement, a gate weakened silently, or a metric redefined after the fact to manufacture a pass.
- **What the protocol must therefore require of every reduction:** name what it removes, state what it raises in escaped-defect risk, and cite the measurement that justified it. Findings per review stays a reported number so the trade stays visible.

The revert rule is unchanged and now matters more, not less: a change that does not prove its saving comes out.

Write the protocol for a one-developer project on a personal token budget. Its own overhead counts against the number it protects.

[...]

Issue #566, opened 2026-09-21
test(agentic): rank where the development loop tokens and wall clock actually go


milocaetano:
Campaign parent: https://github.com/milocaetano/quantick/issues/563
Task key: T3. Base: `origin/campaign/mission-velocity`. Depends on T2.

## Context

A first reading of the transcripts shows every request carrying roughly 180k to
200k tokens of cached context, with thousands of requests on a heavy day. The
cost is context size multiplied by request count. Which part of the loop owns
that product is unknown, and every fix scheduled before the answer is a guess,
which is the mistake #556 made.

## Scope

In: a ranked attribution of the development loop's cost.

Out: the fixes themselves. They become their own campaign children, created
from this ranking.

## Acceptance criteria

- [ ] A report ranks the cost sinks by measured share of tokens and of wall
      clock, separately.
- [ ] Each sink names the phase it belongs to — mission setup, implementation,
      each review in the chain, CI watch, repair rounds, campaign coordination
      — and whether it is main-thread or subagent cost.
- [ ] Wall clock is separated into agent time and waiting time, because waiting
      on CI costs no tokens and a token fix does not shorten it.
- [ ] The ranking says, for each of the top sinks, what would have to change
      for it to shrink, without implementing it.
- [ ] Sinks the measurement cannot see are listed as unmeasured rather than
      omitted.
- [ ] Fix children are created on the campaign from the top of this ranking,
      each with its own acceptance evidence, and linked in the charter index.

milocaetano:
Scheduling input for this ranking, raised by the trader on 2026-09-21 and filed as candidate #570.

#557 moved phase two of a mission — reviews, CI watch and repair — into fresh PR-only agent contexts, and that is merged and reachable from `main` at `4b5e2b7e`. The `mission` skill now draws the boundary at step 8: *From here the main thread only dispatches.*

Steps 1 through 7 are not covered. The request ledger, the injected gates, the grounding and the whole implementation still run in the main thread with the conversation accumulating behind them. The campaign coordinator has no equivalent rule at all.

This ranking must price that phase explicitly — main-thread phase-one cost as a share of mission tokens and of wall clock — so #570 is scheduled on a number rather than on the argument. If it prices low, say so and #570 closes unimplemented.

milocaetano:
**D7 applies to this ranking.** The trader has named token spend as the axis that hurts, so rank by tokens first and report wall clock as the secondary axis.

Two consequences for how you rank:

[...]

Issue #565, opened 2026-09-21
test(agentic): establish the mission cost baseline and grade the seven merged optimizations


milocaetano:
Campaign parent: https://github.com/milocaetano/quantick/issues/563
Task key: T2. Base: `origin/campaign/mission-velocity`. Depends on T1.

## Context

Six optimizations merged between 2026-09-19 and 2026-09-20 — #546, #547, #550,
#552, #553, #556 and #557 — and nobody measured whether they worked. #556's own
pull request body records that it optimized a frame the model never reads. The
campaign needs both the baseline it will be graded against and an honest
verdict on what is already in the trunk.

## Scope

In: the baseline reading, and a per-pull-request verdict for those seven.

Out: ranking the remaining cost sinks, which is T3. Out: proposing fixes.

## Acceptance criteria

- [ ] A report under `docs/quality/velocity/` states the tokens and the wall
      clock a mission costs at baseline, with dispersion and the sample size.
- [ ] For each of #546, #547, #550, #552, #553, #556 and #557 the report states
      `reduced`, `did not reduce` or `cannot attribute`, with the number behind
      it. "Cannot attribute" is a valid and expected verdict given the short
      window and the overlapping merges.
- [ ] The report names what else moved in each window, so a reader sees the
      confounding rather than having it hidden.
- [ ] The report does not claim statistical significance the sample cannot
      carry, and says explicitly where it stops.
- [ ] The baseline figures are copied into the campaign charter's baseline
      section and into the checkpoint `metrics.baseline`.

milocaetano:
executor: opus — chooses the attribution registry and grades seven merged pull requests against a target it must be willing to fail; open judgement. Tier: medium. Base: `origin/campaign/mission-velocity` at `4f998b7f`. Campaign: #563. Decisions D5 and D6 in checkpoint 3 apply.

milocaetano:
Integrated into `campaign/mission-velocity` by PR #571, merge commit `958e8138`, green CI at head `62c35a2a`, all three reviews current against review key `80116cae980d782d94e0ff8d94c96d11728f1efa`.

[...]

Issue #564, opened 2026-09-21
feat(agentic): measure mission token and wall-clock cost from the sources that already exist


milocaetano:
Campaign parent: https://github.com/milocaetano/quantick/issues/563
Task key: T1. Base: `origin/campaign/mission-velocity`.

## Context

The campaign cannot claim a saving without a number to compare against, and it
cannot trust a number whose method was chosen after seeing the data. Three
sources already exist and none of them is read by any committed tool:

- per-request `usage` and `timestamp` in the local session transcripts under
  the Claude projects directory (650 `.jsonl` files, main sessions and subagent
  transcripts alike),
- pull request and CI timings from `gh`,
- `docs/quality/read-cost/ledger.md` and `tools/read_cost/measure.py`.

## Scope

In: a committed harness under `tools/` that reads those three sources and
reports tokens and wall clock per mission, plus the registered method it
implements.

Out: reading the data to draw conclusions, which is T2. Out: any change to the
development loop itself.

## Acceptance criteria

- [ ] The method is registered in a commit that precedes the first commit that
      reports a reading: what a "mission" is for attribution purposes, how a
      transcript is assigned to a mission or pull request, which `usage` fields
      count, how wall clock is bounded, and what counts as a reduction.
- [ ] The harness reads only `usage` fields and timestamps from transcripts. No
      transcript content is read into a report, committed, or published.
- [ ] It attributes cost to main thread versus subagents separately, using the
      `subagents/` transcript layout.
- [ ] It reports dispersion, not only a mean, and marks a result inconclusive
      when n is too small to carry a claim.
- [ ] It runs from the repository with one documented command and its output is
      deterministic for a fixed input set.
- [ ] Transcripts are not in the repository and not reproducible in CI: the
      harness runs locally and commits only derived aggregates. Its README says
      so.
- [ ] Tests cover the attribution and the field filtering against a small
      synthetic fixture committed in the repository.

milocaetano:
executor: opus — designs a measurement method and its attribution rules; open judgement, not retrieval. Tier: medium. Base: `origin/campaign/mission-velocity`. Campaign: #563.

milocaetano:
[...]

PR #569, opened 2026-09-21
feat(agentic): measure mission token and wall-clock cost from the sources that already exist

milocaetano:
Objective: Commit, under `tools/`, a harness that reports tokens and wall clock per mission from the three sources that already exist — the local session transcripts' `usage` and `timestamp` fields, pull-request and CI timings from `gh`, and the read-cost ledger — with its method registered in a commit that precedes any commit reporting a reading.
Tier: medium
Source: https://github.com/milocaetano/quantick/issues/564 (campaign parent https://github.com/milocaetano/quantick/issues/563, task key T1)
Criteria: A1–A8 and A10 delivered; A9 delivered with one bounded, declared exception (nine added lines in `ci.yml` that run the new offline suite); G1–G6 delivered; G-AI1–G-AI4 delivered; closing steps C1–C3 discharged, C4 on hand-back to the coordinator. A10 and G4–G6 were appended by the delivery review's completeness pass; nothing was renumbered. No criterion deferred, none open. Deliberately not done, because they are other issues: the baseline reading and the verdict on the seven merged optimizations (#565), the ranked attribution (#566), any change to the development loop (#567).
Validation: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets`, `cargo build --workspace` and `cargo test --workspace` each run on its own at the final head, exit codes captured separately; `ruff check --select F tools/ bridge/mt5/` clean; 110 offline Python tests passing; the harness run twice over all 652 real local transcripts with byte-identical output on both channels; the `gh` path smoke-tested against merged PR #547. Reviews: architecture PASS, AI review complete with zero open threads, delivery review PASS. Numbers below.

## What this is

The campaign cannot claim a saving without a number to compare against, and it
cannot trust a number whose method was chosen after the data was seen. This is
the instrument, and the method it implements, committed in that order.

**Eight commits, and the order is the point:**

```
01bec148  docs(agentic): register the mission-cost method before any reading
2f4dfeb6  test(agentic): synthetic fixture and tests for the mission-cost harness
b1890d1b  feat(agentic): measure mission tokens and wall clock from existing sources
762ac229  docs(quality): record this pull request's read-cost row
6052b1de  fix(agentic): repair the four review findings in the mission-cost harness
8da3927b  test(agentic): cover the two empty cases the attribution can meet
26f7fdf5  fix(agentic): close the three AI-review findings on the mission-cost harness
5eebc333  fix(agentic): a window needs both of its ends, and one loader serves the package
```

The method is the branch's first commit and it landed alone. **No commit on this
branch reports a number read from the trader's transcripts** — the baseline is
#565's job, and it will be the first reading, taken under thresholds fixed
before anyone could see which way they would fall.

## The method

[`docs/quality/mission-cost/method.md`](https://github.com/milocaetano/quantick/blob/feat/mission-cost-harness/docs/quality/mission-cost/method.md),
eight sections:

1. **What a mission is** — one branch and one pull request, with a work window.
2. **How a transcript is assigned** — declared sessions first, then a single
   window candidate. A session two missions could claim goes to a **shared**
   bucket with both names on it and is never divided, because timestamps cannot
   divide it fairly; a session no window claims goes to **unassigned**. Both
   buckets are reported with their full totals, per session.
3. **Which `usage` fields count** — `timestamp` plus exactly four counters,
   reported separately because they are priced differently and a change that
   trades one for another is invisible in a single total.
4. **How wall clock is bounded** — `IDLE_GAP_SECONDS = 300`, and three
   quantities: `agent_seconds` (concurrency counted), `elapsed_seconds` (the
   union, counted once) and `span_seconds` (the outer bound, not a cost).
5. **What counts as a reduction** — `MIN_GROUP_N = 5`,
   `MIN_RELATIVE_CHANGE = 0.10`, `MAX_UNPLACED_SHARE = 0.20`, and a five-way
   verdict where `cannot_be_attributed` outranks everything and `directional`
   is explicitly *not* a saving.
6. **The registry** — `docs/quality/mission-cost/missions.json`, schema 1,
   committed empty for #565 to populate.
7. **Determinism** — canonical JSON and an `inputs.digest`.
8. **Eleven error modes**, including the uncomfortable one:

> **E2 — The exact fix exists and is deliberately not used.** `cwd` and
> `gitBranch` are present on most transcript lines and would resolve E1
> outright. The reader does not touch them: #564's second acceptance criterion
> limits it to `usage` fields and timestamps, and a privacy boundary that bends
> for a convenient field is not a boundary. E1 is the price, paid openly.

**Two clarifications were made to the method inside this PR, both before any
reading exists, and neither moved a threshold:** the work window's start is read
from GitHub's commit list (a merged branch's merge base collapses onto its own
head, so the local computation returned nothing), and CI time is summed over the
branch's runs rather than one head's (a mission pays for every run it triggered).
Both are visible in `b1890d1b`'s diff of `method.md`.

## The harness

| File | What it owns |
| --- | --- |
| `tools/mission_cost/transcripts.py` | the privacy boundary: discovery, the five-field record, the idle bound |
| `tools/mission_cost/attribution.py` | the registry, the two assignment rules, the main/subagent split |
| `tools/mission_cost/dispersion.py` | quartiles and the registered reduction rule |
| `tools/mission_cost/delivery.py` | `gh` timings and the read-cost row, every call through an injectable runner |
| `tools/mission_cost/measure.py` | the command, the report, the comparison |

`record_from` returns a five-field namedtuple, and the transcript keeps no
records at all — it folds each line into four counters, two instants and a
merged span list and forgets it. The fixture plants bait strings —
`message.content`, `cwd`, `gitBranch`, `service_tier` — and the tests assert
that nothing a transcript still holds carries any of them. The read-cost ledger
is read through `tools/read_cost/ledger.py`'s own parser, not a second one.

## Reviews

**Architecture review: PASS.** Step 0 (`code-review` at `low`, the tier's level)
ran **twice** — once on the branch, once over the repairs — and the second round
found the worst finding of the whole review, introduced by the first round's
fixes.

| Round | Findings | Disposition |
| --- | --- | --- |
| 1, at `8da3927b` | 4 (1 Blocker, 2 Should fix, 1 Consider) | all confirmed, all fixed in `6052b1de` |
| repairs' own tests | 2 more, including a determinism bug | fixed in `6052b1de` |
| 2, at `26f7fdf5` | 3 (1 Blocker, 2 Should fix) | all confirmed, all fixed in `5eebc333` |

The two Blockers, both of which would have corrupted the campaign's numbers
rather than crashed:

- **`_unplaced_share` was all-or-nothing.** One aggregate span covered every
  unplaced session, so a single stray session brushing a group's window charged
  that group the entire unplaced total, pushing the share past
  `MAX_UNPLACED_SHARE` and collapsing every comparison to
  `cannot_be_attributed`. The buckets now carry each session's own totals and
  extent.
- **A half-resolved window was unbounded on the left.** `_overlaps` refused a
  window only when *both* ends were null, so a mission whose `started_at` never
  resolved claimed every session in the directory that ended before its
  `ended_at` — weeks of work done before the branch existed, charged to it in
  silence. A window is `[started_at, ended_at)` and now needs both of its ends.

Also fixed: a determinism bug where Windows turned LF into CRLF on the way to
standard output while `--out` wrote LF, so the one documented command hashed two
ways depending on how it was captured.

**AI review: complete, zero open threads.** Three WEAK findings published as
resolvable threads, all three fixed and resolved:
`notes` were free-form strings inside a *successful* report and are now typed
`{branch, kind, detail}` objects a consumer can branch on; every `gh` command is
built in one place instead of inline in three functions; and the reader's
retained state went from O(requests) to O(work blocks). Final verdicts: all six
PASS.

**Delivery review: PASS**, in the completeness-only mode tier `medium` buys —
stated as such in the report, because a `medium` marker that reads like a full
one is worse than no marker. It derived 22 distinct asks from the retained
request before opening the map and found **five unledgered**, two of them live
gaps rather than bookkeeping:

- *"If you find something tempting, write it in a comment on #563 and move on"*
  had no criterion and no comment. Discharged:
  [#563 comment 5754311754](https://github.com/milocaetano/quantick/issues/563#issuecomment-5754311754),
  four properties of the instrument that #565 and #566 would otherwise
  rediscover the hard way, with no baseline conclusion in it.
- `new-task`'s start obligations were only partly discharged — **the board card
  was still on `Todo`**. Moved to `In Progress` on both the Status and Campaign
  state fields and read back.

The other three had shipped and were traceability gaps. `R13`–`R16`, `A10` and
`G4`–`G6` were appended, nothing was renumbered, and the correction is recorded
in the goal under `## Map corrections`. Neither shape review can see either live
gap: they grade the diff, and neither gap was in it.

## Validation

**The four checks, each run on its own at the final head, exit codes captured
separately — no chained all-clear:**

| Check | Result |
| --- | --- |
| `cargo fmt --all -- --check` | `0` |
| `cargo clippy --workspace --all-targets` | `0`, 0 warnings |
| `cargo build --workspace` | `0` |
| `cargo test --workspace` | `0`, no failing result line |

**Python, which `cargo` cannot see:**

```
$ ruff check --select F tools/ bridge/mt5/
All checks passed!

$ python -m unittest discover -s tools/mission_cost -p 'test_*.py'
Ran 110 tests in 1.58s
OK
```

**One documented command, over the trader's real 652 transcripts, twice, on both
output channels:**

```
$ python tools/mission_cost/measure.py report --repo . --no-gh --out run1.json
$ python tools/mission_cost/measure.py report --repo . --no-gh > run2.json
$ sha256sum run1.json run2.json
5861821c3e7510e6eb1a4cdad8223589cfb116ce967a35b77ffc9ef5d095896f  run1.json
5861821c3e7510e6eb1a4cdad8223589cfb116ce967a35b77ffc9ef5d095896f  run2.json
```

652 transcripts, 49 sessions, **0 skipped, 0 unparsed lines, 0 undated usage
lines, 0 out-of-order lines**, 3.7 seconds. No token totals are quoted here:
reading them is #565's work, not this PR's.

**The `gh` path, smoke-tested against merged PR #547** with a throwaway registry
that was not committed. It resolved the window from GitHub
(`2026-09-19T13:31:37Z` → `2026-09-19T21:42:46Z`, the start taken from that pull
request's earliest commit), read 7 completed CI runs (4,402 s, listing not
truncated), pulled its read-cost row from the ledger, and split its transcript
cost into main thread and subagents. It also reported
`share_of_billable_tokens: 0.77` for the unplaced buckets — which is error mode
E1 measuring itself, and is exactly why `MAX_UNPLACED_SHARE = 0.20` exists. A
comparison on that basis would be graded `cannot_be_attributed`: the instrument
refusing to make a claim it cannot support.

## Performance impact

No Rust production path is touched. Per-trade, per-depth and per-frame rates are
unchanged, and `cargo run -p quantick-guards -- --report` is identical for every
crate. The harness is a local script, run by hand, rarely — and after the AI
review's third finding its retained memory is proportional to work blocks rather
than to every request ever made in a projects directory nobody prunes.

## Scope

Nothing here reads the data to draw a conclusion, grades any of #546, #547,
#550, #552, #553, #556 or #557, or changes the development loop. `.claude/`,
`AGENTS.md`, `CLAUDE.md` and `docs/workflow/` are untouched.

**The one declared exception**: nine added lines in `.github/workflows/ci.yml`,
a single step that runs the new offline suite, mirroring the `read-cost` and
`edit-loop` steps beside it. That is the ordinary obligation of adding tested
code — a suite CI never runs is not tested — and it changes no existing step,
tunes nothing and removes no gate. Tuning the loop is #567's work.

## What is not in this repository, and will not be

The session transcripts are the trader's own conversations. They are not
committed, they are not reproducible on a runner, and the harness is therefore
local by construction. Only derived aggregates are ever committed. The fixture
under `tools/mission_cost/fixtures/` was written by hand for these tests; not a
line of it came from a real session. `tools/mission_cost/README.md` says all of
this under *What is not in the repository*.

<details>
<summary>The mission goal (<code>.claude/GOAL.md</code>)</summary>

# Mission — a committed harness that measures what a mission costs

**Objective:** Commit, under `tools/`, a harness that reports tokens and wall
clock per mission from the three sources that already exist — the local session
transcripts' `usage` and `timestamp` fields, pull-request and CI timings from
`gh`, and the read-cost ledger — with its method registered in a commit that
precedes any commit reporting a reading.

**Why:** The campaign (#563) cannot claim a saving without a number to compare
against, and it cannot trust a number whose method was chosen after seeing the
data. This mission builds the instrument; #565 takes the reading.

**Tier:** medium. Justification: the coordinator set it. The work is a new
committed tool with its own tests and a registered method, in a campaign whose
whole credibility rests on the method being fixed before the data is seen, so
`delivery-review`'s completeness pass applies. It is not `small`: the diff adds
a new tool package plus a method document, and the ordering constraint is a
delivery claim that has to be graded.

**Campaign child.** Parent #563, task key T1, base `origin/campaign/mission-velocity`,
PR base exactly `campaign/mission-velocity`. Completion returns to the
coordinator; this mission merges nothing.

## Request ledger

- **R1** — A committed harness under `tools/` reads the three existing sources
  and reports tokens and wall clock per mission.
  *Source:* #564 Scope — "a committed harness under `tools/` that reads those
  three sources and reports tokens and wall clock per mission, plus the
  registered method it implements".
- **R2** — The method is registered in a commit that precedes the first commit
  that reports a reading, and it states: what a mission is for attribution
  purposes, how a transcript is assigned to a mission or pull request, which
  `usage` fields count, how wall clock is bounded, and what counts as a
  reduction. *Source:* #564 acceptance criterion 1.
- **R3** — The harness reads only `usage` fields and timestamps from
  transcripts. No transcript content is read into a report, committed, or
  published. *Source:* #564 acceptance criterion 2; #563 Constraints; D3.
- **R4** — Cost is attributed to main thread versus subagents separately, using
  the `subagents/` transcript layout. *Source:* #564 acceptance criterion 3.
- **R5** — The report carries dispersion, not only a mean, and marks a result
  inconclusive when n is too small to carry a claim. *Source:* #564 acceptance
  criterion 4; #563 Constraints — "Report dispersion and say when a result is
  inconclusive. Do not claim significance the data cannot carry."
- **R6** — It runs from the repository with one documented command and its
  output is deterministic for a fixed input set. *Source:* #564 acceptance
  criterion 5.
- **R7** — The README states that transcripts are not in the repository and not
  reproducible in CI, that the harness runs locally, and that only derived
  aggregates are committed. *Source:* #564 acceptance criterion 6; D3.
- **R8** — Tests cover the attribution and the field filtering against a small
  synthetic fixture committed in the repository. The fixture is written by this
  mission, never copied from a real transcript. *Source:* #564 acceptance
  criterion 7; D3.
- **R9** — Scope holds: no conclusion about the baseline, no verdict on the
  seven merged pull requests, no change to the development loop. *Source:*
  #564 Scope — "Out: reading the data to draw conclusions, which is T2. Out:
  any change to the development loop itself"; coordinator brief "SCOPE
  DISCIPLINE".
- **R10** — The registered method states the assignment rule's known error
  modes honestly. *Source:* D5 — "a time-window plus branch/PR mention
  heuristic is acceptable if you state its error modes honestly in the
  registered method".
- **R11** — Every repository artifact is in English. *Source:* coordinator
  brief; `CLAUDE.md`.
- **R12** — The mission hands its PR back to the coordinator and merges
  nothing, and does not mark the issue done. *Source:* D1; coordinator brief
  "Do not merge. Do not mark the issue done. Return control to the
  coordinator."
- **R13** — Anything tempting but out of scope is written as a comment on the
  campaign parent #563 rather than acted on. *Source:* coordinator brief —
  "If you find something tempting, write it in a comment on #563 and move on."
- **R14** — The campaign child's branch-bound context is recorded: the worktree
  is cut from `origin/campaign/mission-velocity`, the `mission-base` file
  carries the four required fields, and the work happens under
  `../quantick-worktrees/` and never in the main checkout. *Source:*
  coordinator brief, "BASE AND WORKTREE"; the integration contract's
  branch-bound task context.
- **R16** — The four token kinds are reported separately and never silently
  summed into one number. *Source:* coordinator brief, GROUNDING — "Do not
  treat all four token kinds as interchangeable - report them separately,
  because they are priced differently and they tell different stories."
- **R15** — The issue is started the way `new-task` requires: read in full,
  branch prefixed correctly, board card moved to In Progress. *Source:*
  coordinator brief — "follow ... the `new-task` ... skills as that mission
  requires".

## Decisions

Step 3 was discharged by the coordinator before dispatch; a subagent cannot ask
the trader. These are settled and are not reopened.

- **D1** — Agents may merge into `campaign/mission-velocity` only. Main merges
  are exclusively the trader's action. This mission merges nothing and hands
  its PR back to the coordinator.
- **D2** — 60% is the campaign's target, not a pass condition. This mission
  builds the instrument, not the verdict.
- **D3** — Session transcripts are the trader's own conversations. Read only
  `usage` fields and `timestamp`. Never read, quote, commit, log or publish
  transcript content. Commit derived aggregates only. The committed fixture is
  synthetic, written by this mission.
- **D4** — The harness lives under `tools/`, following the layout of
  `tools/read_cost/` and `tools/edit_loop/`: Python, tested the way those are.
  `python` on this machine is 3.12; `python3` is a broken shim.
- **D5** — A mission, for attribution, is one branch and one pull request. This
  mission chooses and registers how a transcript is assigned to one; the
  registered method states the error modes honestly.

## Assumptions

- **S1** — *Wanted to ask.* `cwd` and `gitBranch` are present on most
  transcript lines and would make assignment exact under concurrency. D3 and
  #564's second acceptance criterion say the harness reads only `usage` fields
  and timestamps, so the reader does not touch them. The reading taken: the
  criterion is literal, and the cost of honouring it — weaker attribution when
  two missions overlap in time — is paid openly in the method's error modes
  rather than quietly waived. The registry's explicit session-id form exists so
  an operator can restore exactness without the reader reading more.
- **S2** — The report's unit of comparison is the mission, because D5 defines
  one, and the campaign's target (#563 C6) is stated per mission.
- **S3** — Derived aggregates for the baseline itself belong to #565, so this
  mission commits no reading at all. The ordering R2 requires is then proved by
  the method being the branch's first commit and no later commit carrying a
  number read from the trader's transcripts.
- **S4** — `tools/mission_cost/` is the package name, matching
  `tools/read_cost/`'s convention of a directory per instrument. A conventional
  placement decision, not a question.

## Map corrections

The delivery review's completeness pass read the retained request against this
map and found four operational obligations nothing pointed at. Three had been
discharged and were traceability gaps only -- the branch-bound campaign context
(now `R14`/`G4`), the `new-task` start obligations (`R15`/`G5`, and the board
card was in fact still on Todo, so this one had an outcome gap too) and the
"do not mark the issue done" constraint (folded into `R12`/`C4`). The fourth,
"write it in a comment on #563 and move on" (`R13`/`G6`), was a real uncovered
ask and is now discharged. A fifth, in the brief's grounding notes rather than
its instructions -- "report them separately, because they are priced
differently" -- was delivered but unledgered, and is now `R16`/`A10`. The
numbers above them are unchanged.

## Acceptance criteria

- [ ] **A1** — `tools/mission_cost/` reads all three sources and prints one
      canonical-JSON report of tokens and wall clock per mission.
      *Evidence:* `python tools/mission_cost/measure.py report --help` and a
      real run over the local transcripts, output shape quoted in the PR body.
      → PR body, Validation. *(R1)*
- [ ] **A2** — `docs/quality/mission-cost/method.md` is the branch's first
      commit and defines the mission unit, the assignment rule, the counted
      `usage` fields, the wall-clock bound and what counts as a reduction.
      *Evidence:* `git log --reverse --oneline <base>..HEAD` shows the method
      commit first; the document's five sections. → PR body, Validation. *(R2, R10)*
- [ ] **A3** — The transcript reader returns a fixed-shape record built from
      `timestamp` and the four `message.usage` counters only, and a test proves
      that a line carrying message content yields a record with no trace of it.
      *Evidence:* `tools/mission_cost/test_transcripts.py`, the field-filtering
      tests. → PR body, Validation. *(R3, R8)*
- [ ] **A4** — Main-thread and subagent cost are separate columns in the
      report, keyed off the `subagents/` path segment, and a test proves the
      split over the synthetic fixture.
      *Evidence:* `tools/mission_cost/test_attribution.py`. → PR body,
      Validation. *(R4, R8)*
- [ ] **A5** — Every aggregated metric carries n, median, p25, p75, min and max
      alongside the mean, and a verdict field reads `inconclusive` when n is
      below the registered threshold.
      *Evidence:* `tools/mission_cost/test_statistics.py`. → PR body,
      Validation. *(R5)*
- [ ] **A6** — One documented command produces byte-identical output on two
      runs over the same input set.
      *Evidence:* the README's single command, run twice and hashed. → PR body,
      Validation. *(R6)*
- [ ] **A7** — `tools/mission_cost/README.md` states that transcripts are not
      in the repository, are not reproducible in CI, that the harness runs
      locally, and that only derived aggregates are committed.
      *Evidence:* the README's *What is not in the repository* section.
      → PR body, Validation. *(R7)*
- [ ] **A8** — A synthetic transcript fixture is committed under
      `tools/mission_cost/fixtures/`, written by this mission, and the tests run
      against it.
      *Evidence:* the fixture files and `python tools/mission_cost/test_*.py`
      passing. → PR body, Validation. *(R8, R3)*
- [ ] **A9** — The diff contains no baseline conclusion, no verdict on #546,
      #547, #550, #552, #553, #556 or #557, and no change to `.claude/`,
      `AGENTS.md`, `CLAUDE.md` or `docs/workflow/`. The one `.github/workflows/`
      edit is additive and bounded: a single step that runs the new offline
      suite, mirroring the `read-cost` and `edit-loop` steps beside it. It is
      the ordinary obligation of adding tested code, not a change to the
      development loop — that is #567's work, and this step optimizes nothing.
      *Evidence:* `git diff --stat <base>...HEAD`, and the CI diff read in the
      PR body. → PR body, Validation. *(R9)*
- [ ] **A10** — The report carries `input_tokens`,
      `cache_creation_input_tokens`, `cache_read_input_tokens` and
      `output_tokens` as separate fields for the mission total, the main thread
      and the subagents, with the derived `billable_tokens` alongside them and
      never instead of them.
      *Evidence:*
      `test_measure.Shape.test_a_mission_reports_all_four_token_kinds_and_not_one_total`
      and `test_transcripts.Totals.test_totals_sum_the_four_kinds_separately`;
      section 3 of the method. → PR body, Validation. *(R16)*
- [ ] **G1** — Every artifact in English, graded by `arch-review` dimension 8
      and the language guard.
      *Evidence:* `arch-review` report; `cargo test -p quantick-guards`.
      → PR body, Reviews. *(R11)*
- [ ] **G2** — The four checks green at the final head, each run on its own:
      `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets`,
      `cargo build --workspace`, `cargo test --workspace`; plus
      `ruff check --select F tools/` and the new Python tests under `python`.
      Performance impact declared: no Rust production path is touched, so the
      per-trade, per-depth and per-frame rates are unchanged; the harness is a
      rarely-run local script.
      *Evidence:* command output in the PR body; CI at the exact head.
      → PR body, Validation. *(R1)*
- [ ] **G3** — `arch-review` run with every Blocker and Should-fix resolved or
      deferred in the PR body.
      *Evidence:* the published `arch-review` report. → PR body, Reviews.
- [ ] **G4** — The branch-bound campaign context is in place: worktree cut from
      `origin/campaign/mission-velocity`, `mission-base` written with the four
      required fields, `mission-tier` recording `medium`, and no write to the
      main checkout.
      *Evidence:* `sh .claude/hooks/campaign_context.sh base <wt>` returns
      `origin/campaign/mission-velocity`; the `mission-base` and `mission-tier`
      files in the worktree's private git directory. → PR body, Validation.
      *(R14)*
- [ ] **G5** — The `new-task` start obligations are discharged: #564 read in
      full, branch prefixed `feat/`, board card moved to In Progress on both
      the Status and Campaign state fields.
      *Evidence:* the project item's field values read back after the move.
      → PR body, Validation. *(R15)*
- [ ] **G6** — Every out-of-scope observation is filed as a comment on #563 and
      not acted on in this diff.
      *Evidence:* https://github.com/milocaetano/quantick/issues/563#issuecomment-5754311754
      → #563. *(R13, R9)*
- [ ] **G-AI1** — AI review is executed for the current PR review.
- [ ] **G-AI2** — A durable AI-review report is published on the PR.
- [ ] **G-AI3** — `ai_review_threads.sh list` returns zero unresolved threads.
- [ ] **G-AI4** — `ai-review-complete` is valid for the current review key.

### Not applicable

- **Touches a hot path** — no Rust production path is touched; the harness is a
  local Python script run by hand.
- **Touches anything user-visible** — no UI surface changes, so `ui-harness`,
  `visual-qa` and `trader-ux-review` do not apply.
- **Adds a capability (feed, bar type, indicator, layer, panel, crate)** — the
  harness is a `tools/` script, not a workspace crate, and docks nothing into
  the application. `new-extension` does not apply.
- **Adds something a trader does** — nothing a trader operates changes. The
  harness is itself operable without a mouse: one documented command.
- **Engine / determinism territory** — the engine is untouched. The harness's
  own determinism is covered by A6.

## Closing steps

- **C1** — `delivery-review` PASS recorded for the current review key.
- **C2** — Draft PR open against `campaign/mission-velocity` carrying the
  mission summary block and this goal.
- **C3** — `sh .claude/hooks/mission_ship_gate.sh mission <pr>` PASS from this
  worktree, after exact-head green CI.
- **C4** — Control returned to the coordinator with the PR URL, head SHA, CI
  verdict, review outcomes, which acceptance criteria of #564 are proven and
  which are not, and anything deferred. No merge, and the issue is left open
  and not marked done. *(R12)*

## The verbatim request

> You are a campaign child mission for the quantick repository at C:\src\quantick.
> Read CLAUDE.md and AGENTS.md first, then docs/campaign/integration.md, then
> follow the `mission` skill at tier `medium` and the `new-task`, `arch-review`,
> `ai-review`, `delivery-review` and `ship` skills as that mission requires.
> Work in English for every repository artifact.
>
> TASK: GitHub issue #564 (task key T1). Read it in full with `gh issue view 564`.
> CAMPAIGN PARENT: issue #563. Read its charter body and the latest checkpoint
> comment before starting.
>
> BASE AND WORKTREE — this is a campaign child, so the integration contract
> supersedes the main-based examples in every skill:
> - Cut your branch and worktree from `origin/campaign/mission-velocity`, NOT
>   from origin/main.
> - Your PR base is exactly `campaign/mission-velocity`.
> - After creating the worktree, write one UTF-8/LF line to its private git
>   directory's `mission-base` file:
>   `<your-branch> origin/campaign/mission-velocity https://github.com/milocaetano/quantick/issues/563 https://github.com/milocaetano/quantick/issues/563#issuecomment-5754017735`
> - Work under ../quantick-worktrees/, never in the main checkout. The main
>   checkout is on a detached HEAD with unrelated staged files — do not touch
>   it, do not clean it.
>
> MISSION STEP 3 IS ALREADY DISCHARGED. A subagent cannot ask the trader. The
> coordinator ran it; these are your decisions D1..D5, record them in GOAL.md
> and do not re-open them:
> - D1: Agents may merge into `campaign/mission-velocity` only. Main merges are
>   exclusively the trader's action. Do not merge anything yourself; hand your
>   PR back to the coordinator.
> - D2: 60% is the campaign's target, not a pass condition. Your job is the
>   instrument, not the verdict.
> - D3: Session transcripts under C:\Users\User\.claude\projects\C--src-quantick\
>   are the trader's own conversations. Read ONLY `usage` fields and
>   `timestamp`. Never read, quote, commit, log or publish transcript content.
>   Commit derived aggregates only. The committed test fixture must be
>   synthetic, written by you, never a copy of a real transcript.
> - D4: The harness lives under `tools/` following the layout of the existing
>   `tools/read_cost/` and `tools/edit_loop/`. Python, matching how those are
>   written and tested. `python` on this machine is 3.12; `python3` is a broken
>   shim — use `python`.
> - D5: A "mission" for attribution purposes is one branch and one pull request.
>   You choose and register how a transcript is assigned to one; a time-window
>   plus branch/PR mention heuristic is acceptable if you state its error modes
>   honestly in the registered method.
>
> WHAT MAKES THIS MISSION UNUSUAL, AND THE ONE THING THAT MUST NOT SLIP:
> acceptance criterion one requires that the METHOD IS COMMITTED BEFORE THE
> FIRST READING IS COMMITTED. This is the whole point of the task — a method
> chosen after seeing the data proves nothing. So your first commit on the
> branch is the written method (what a mission is, how a transcript is assigned,
> which usage fields count, how wall clock is bounded, what counts as a
> reduction, and the known error modes). Only after that commit exists do you
> commit anything that reports a number. `delivery-review` reads the write order
> and the git log, so this must be visible in history, not just claimed.
>
> GROUNDING ALREADY DONE, so you do not have to rediscover it:
> - 650 .jsonl files exist under that projects directory, laid out as
>   `<session-uuid>.jsonl` and `<session-uuid>/subagents/agent-*.jsonl`, plus
>   `<session-uuid>/tool-results/`. The subagent layout is how you separate
>   main-thread from subagent cost.
> - Every usage-bearing line parses as JSON; a probe over all 650 files found 0
>   unparsed lines. Relevant fields seen:
>   `message.usage.{input_tokens,output_tokens,cache_read_input_tokens,cache_creation_input_tokens}`
>   and a top-level `timestamp`.
> - Cache reads are roughly 99% of all input tokens. A day's totals ranged from
>   4,500 to 13,196 requests, with cache-read input from 815M to 2.58B tokens.
>   Do not treat all four token kinds as interchangeable — report them
>   separately, because they are priced differently and they tell different
>   stories.
> - `gh` is authenticated. `gh run list --json` and `gh pr view --json` give the
>   CI and PR timings.
> - `docs/quality/read-cost/ledger.md` and `tools/read_cost/measure.py` already
>   exist; reuse rather than duplicate.
>
> SCOPE DISCIPLINE: deliver the harness and its registered method. Do NOT draw
> conclusions about the baseline, do NOT grade the seven merged pull requests,
> do NOT propose or make any change to the development loop. Those are issues
> #565, #566 and #567 and they are other people's work. If you find something
> tempting, write it in a comment on #563 and move on.
>
> VALIDATION: the four checks at your final head, run each on its own and never
> chain them with `|| echo` — `cargo fmt --all -- --check`,
> `cargo clippy --workspace --all-targets`, `cargo build --workspace`,
> `cargo test --workspace`. Plus `ruff check --select F` over anything you add
> under tools/, and run your own Python tests directly with `python`. A draft PR
> needs the `full-ci` label to get a real verdict.
>
> WHEN DONE: open the PR against `campaign/mission-velocity`, run the reviews
> the mission tier requires, and report back to me with: the PR URL, its head
> SHA, the CI verdict, the review outcomes, which acceptance criteria of #564
> are proven and which are not, and anything you deferred. Do not merge. Do not
> mark the issue done. Return control to the coordinator.

— the campaign coordinator for #563, dispatching task T1, 2026-09-20.

</details>

Head: 5eebc3339134ebabf90f2bb4b0f1ab5b64f947ce
Review key: 5e55c709b289d98d737d1c92972da97ee8603afc
Architecture report: https://github.com/milocaetano/quantick/pull/569#issuecomment-5754297847
Delivery report: https://github.com/milocaetano/quantick/pull/569#issuecomment-5754322996
AI report: https://github.com/milocaetano/quantick/pull/569#issuecomment-5754299089
CI: full ci and windows passed at the exact head; every other check that ran passed
AI review threads: zero returned by ai_review_threads.sh list


tools/mission_cost/measure.py
```diff
@@ -0,0 +1,433 @@
...
+                facts = pulls.facts(mission.pr)
+                if ended is None and facts["ended_at"]:
+                    ended = TRANSCRIPTS.parse_timestamp(facts["ended_at"])
+                if started is None:
+                    first = DELIVERY.first_commit_instant(mission.pr)
+                    started = TRANSCRIPTS.parse_timestamp(
+                        first or facts["created_at"] or ""
+                    )
+            except DELIVERY.DeliveryError as problem:
+                notes.append(f"{mission.branch}: {problem}")
```
milocaetano: **WEAK — AI-ready.** `notes` is a list of free-form strings inside a *successful* report, and it is the only place a failed `gh` lookup is recorded. An agent consuming this report cannot branch on it: `"feat/x: gh pr view 9 failed: ..."` and `"feat/x: no window, so only declared sessions reach it"` are two different conditions with two different remedies, and both arrive as prose.

It holds today because a person reads the notes. The variant that breaks it is #565: the baseline run has dozens of missions, and it needs to exclude programmatically the ones whose window could not be resolved — otherwise a mission with an unresolved window silently reports near-zero cost and drags the group median down, which is exactly the class of error the registered method's `cannot_be_attributed` verdict exists to prevent.

**Fix:** make each note a typed object rather than a sentence — `{"branch": ..., "kind": "gh_lookup_failed" | "no_window", "detail": "<the message>"}` — and keep the human sentence in `detail`. `kind` is then a closed, documented set an agent can switch on, and the report stays canonical JSON.

tools/mission_cost/delivery.py
```diff
@@ -0,0 +1,202 @@
...
+    first = TRANSCRIPTS.parse_timestamp(start)
+    last = TRANSCRIPTS.parse_timestamp(end)
+    if first is None or last is None:
+        return None
+    return (last - first).total_seconds()
+
+
+def pull_facts(number, runner=shell, as_of=None):
+    """The timings of one pull request, and nothing else it carries."""
+    raw = runner(["gh", "pr", "view", str(number), "--json", PULL_FIELDS])
```
milocaetano: **WEAK — decoupled.** Every call that reaches GitHub builds its own `gh` argument vector inline: `pull_facts` here, `first_commit_instant` at line 174, `ci_facts` at line 129. The `runner` injection makes all three *testable* — the fake runner in `test_measure.py` proves that — but it does not make them *swappable*: the command and its `--json` field list are welded into each function body.

It holds today because there is one forge and one CLI. Two variants break it. The first is a field `gh pr view` does not expose, which means dropping to `gh api` with a GraphQL or REST path — a fourth command shape edited into a fourth place. The second is measuring a mission whose pull request is not on GitHub at all, which this campaign does not need but #566's attribution work might, since it wants CI timings the CLI reports only in aggregate.

**Fix:** one `Forge` seam — a small class holding the runner and exposing `pull(number)`, `commits(number)` and `runs(branch)`, with the three argument vectors as module constants beside `PULL_FIELDS` and `RUN_FIELDS`. The three functions then take a forge instead of a runner, the fake in the tests becomes a fake forge, and a second implementation is a new class rather than three edited function bodies.

Not filed higher because the harness is a local script with one caller and the current shape matches `tools/read_cost/ledger.py`'s `gh_json`; a seam invented before its second implementation would be its own finding under standardisation.

tools/mission_cost/transcripts.py
```diff
@@ -0,0 +1,206 @@
...
+                continue
+            message = parsed.get("message")
+            usage = message.get("usage") if isinstance(message, dict) else None
+            if not isinstance(usage, dict):
+                continue
+            stamp = parse_timestamp(parsed.get("timestamp"))
+            if stamp is None:
+                item.undated += 1
+                continue
+            item.records.append(
```
milocaetano: **WEAK — scalable.** `Transcript.records` is an unbounded list, and `discover` holds every record of every transcript in memory at once before anything is aggregated. The rate is *rare* — a hand-run batch script — so this is not a hot path, but the per-input cost is not independent of history: memory grows linearly with the total number of requests ever made in that projects directory, and no cap is stated anywhere.

It holds today: 651 transcripts, roughly 42,000 usage-bearing lines, 3.7 seconds, and the records are five-field tuples. The variant that breaks it is the one this campaign is built to run — #565 reading a full baseline window, and later #568 re-reading it after the loop changes, across a projects directory that has not been pruned and, with the multi-root support added here, across several worktree directories at once. The failure is not a wrong number; it is a run that gets slower and heavier every week for no reason the reader can see.

**Fix:** nothing downstream needs the records themselves. `totals()` is a sum and `intervals()` is a fold over consecutive pairs, so `read` can accumulate the four counters, the first and last timestamps, and the merged interval list as it goes, and never keep the list. That makes the retained state O(work blocks) instead of O(requests), and the public surface — `totals()`, `first()`, `last()`, `intervals()`, `active_seconds()` — does not change, so no caller and no test moves.

Deliberately not fixed in this pull request: that is a rewrite of the reader, which is the privacy boundary, and it would land after the bait-string tests were written rather than before. Filed so #565 measures before it grows.

tools/mission_cost/measure.py
milocaetano: Fixed in `26f7fdf5`. Each note is now `{branch, kind, detail}` with `kind` from the closed set `NOTE_KINDS = ("gh_lookup_failed", "no_window")`, and the human sentence kept in `detail`. Proved by `test_measure.Notes.test_a_note_is_typed_so_a_consumer_can_branch_on_it`.

tools/mission_cost/delivery.py
milocaetano: Fixed in `26f7fdf5`, at the level the finding argued for rather than the class it also mentioned. Every command this module sends is now built by `pull_command`, `commits_command` and `runs_command` at the top of the file, beside `PULL_FIELDS`, `COMMIT_FIELDS` and `RUN_FIELDS`. A drop to `gh api` is one place to edit, and `test_measure.Commands` pins all three shapes. A `Forge` class is deliberately not introduced: there is one implementation, and a seam invented before its second implementation would be a standardisation finding of its own.

tools/mission_cost/transcripts.py
milocaetano: Fixed in `26f7fdf5`. `Transcript` keeps no records at all now: `read` folds each line into four counters, first and last instants, and a merged span list, then forgets it. Retained state is proportional to work blocks, not to requests. The public surface is unchanged, and the privacy test got stronger in the move — there are no records left to inspect, so it reads everything the object still holds (`repr(vars(item))`) and asserts no bait string is in any of it. Folding in file order trusts an append-only log; 0 out-of-order lines across the trader's 652 real transcripts, and one is now counted and reported rather than silently dropped (`test_a_line_out_of_order_is_counted_rather_than_folded`).

milocaetano:
# Architecture review — PR #569 @ `5eebc3339134ebabf90f2bb4b0f1ab5b64f947ce`

step 0: code-review at low (effort-first, no reuse notice), run twice — 4
findings at `8da3927b`, all 4 confirmed and fixed; 3 findings at `26f7fdf5`, all 3 confirmed and fixed in `5eebc333`; 0 open.

**Identity.** Branch `feat/mission-cost-harness`, worktree
`C:\src\quantick-worktrees\feat-mission-cost-harness` (clean), base
`origin/campaign/mission-velocity` via `campaign_context.sh`, base tip
`a4ba1cbc`, review key `5e55c709b289d98d737d1c92972da97ee8603afc`. Campaign
child of #563, task T1.

**Tier** `medium`, read from `mission-tier` (`feat/mission-cost-harness medium`),
matching the `**Tier:**` line in `GOAL.md`. All nine dimensions read.

**Diff.** 25 files, 3,169 insertions. One new Python package under `tools/`,
one method document under `docs/quality/`, one read-cost row, one line in the
docs index, nine added lines in `ci.yml`. **No Rust file changed and no existing
line modified anywhere in the diff.**

**Eight commits, and the order is the finding that matters most.** The method
(`01bec148`) is the branch's first commit and it landed alone. Nothing on this
branch reports a number read from the trader's transcripts. That is what #564's
first acceptance criterion protects and the git log is the evidence.

## Step 0 — correctness

Round one, at `8da3927b`: four findings, four confirmed by reading the code,
four fixed in `6052b1de`.

| Severity | Finding | Fix |
| --- | --- | --- |
| **Blocker** | `measure.py:_unplaced_share` — the shared and unassigned buckets were all-or-nothing. One aggregate span covered every unplaced session, so a single stray session brushing a group's window charged that group the whole unplaced total, pushing `unplaced_share` past `MAX_UNPLACED_SHARE` and collapsing every comparison to `cannot_be_attributed`. | The buckets carry each session's own totals and extent (`attribution.session_totals`); the share sums only the sessions whose time reaches the window. |
| **Should fix** | `measure.py:resolve_windows` / `delivery_for` — `gh pr view` ran twice per mission and the first answer was discarded. | `Pulls`, one cache. |
| **Should fix** | `transcripts.py:Transcript.span_seconds` — dead; attribution computes the mission-level span itself. | Removed. |
| **Consider** | `delivery.py:load_transcripts` — a loader of a different shape from its two siblings and without their guard, so a missing file raised a bare `AttributeError`. | One `load(name, path=None)` shape, guard included. |

Graded Blocker rather than Should-fix because the first one changes the output:
the threshold deciding whether the campaign may make a claim at all was computed
on the wrong denominator, and the failure mode is an instrument that refuses
every comparison. Correctness outranks the rest.

**Two more the repairs' own tests found**, both fixed in the same commit:

- `delivery.py:ci_facts` listed runs with `--limit 100` and could not tell a
  full branch from a truncated listing, so CI time could read low in silence.
  `CI_RUN_LIMIT = 300` with a `ci_listing_truncated` flag.
- `measure.py:emit` — **a determinism bug**. Through the text layer Windows
  turned every newline into CRLF on the way to standard output while `--out`
  wrote LF, so the one documented command hashed two ways depending on how it
  was captured. It writes bytes now, and both channels agree over the trader's
  652 real transcripts.

Round two, over the delta introduced by `8da3927b` and `26f7fdf5`. Three more,
all confirmed, all fixed in `5eebc333`. The first is the worst finding of this
whole review and it was introduced *by* the round-one repairs, which is the
argument for running the bug pass again over them rather than declaring victory:

| Severity | Finding | Fix |
| --- | --- | --- |
| **Blocker** | `attribution.py:_overlaps` — a window was refused only when **both** ends were null. A half-resolved one (`ended_at` read from GitHub, `started_at` still null because the commit list came back empty) was therefore unbounded on the left, and that mission claimed every session in the directory ending before its `ended_at` — weeks of work done before the branch existed, charged to it in silence, inflating exactly the number this campaign will quote. | `has_window` requires both ends, because a window is `[started_at, ended_at)` and half of one is not a narrower window but an unbounded one. Proved by `test_attribution.Windows.test_a_half_resolved_window_claims_nothing_by_time`. |
| **Should fix** | `measure.py:resolve_windows` — a half-resolved window produced no note at all, so a consumer excluding unresolved windows could not see the case that reads as a suspiciously cheap mission. | A third note kind, `partial_window`, distinct from `no_window`. |
| **Should fix** | `attribution.py:load` / `measure.py:load` — verbatim copies of the generalised `delivery.py:load` sitting in the same diff: three loaders to keep in step. | One implementation, in the leaf module, reached through a small bootstrap that is the irreducible part. It also stops handing back a `FileNotFoundError` naming a half-joined path. |

Two step-0 rounds is the budget. Round two found no finding that round one had
already passed on unchanged code; everything it returned was in the repairs
themselves. Nothing is open at this head.

## The nine dimensions

**1 — Docking.** A second consumer of the reader docks without opening a file:
`discover` returns a list and everything downstream takes it. A fourth external
*source* would mean editing `build_report`, which names each source explicitly —
the same shape `tools/read_cost/` and `tools/edit_loop/` already use, so
inventing a source registry for a hand-run local script would be a second way to
do a solved thing (dimension 5) rather than an improvement. Named, not filed.

**2 — Performance.** Rate: **rare**. A local script run by hand; no Rust
production path touched, so per-trade, per-depth and per-frame rates are
unchanged and `cargo run -p quantick-guards -- --report` is identical for every
crate. 3.7 s over 652 real transcripts; 110 tests in 1.6 s. The `26f7fdf5`
repair improved the retained-memory profile from O(requests) to O(work blocks),
which is the direction a harness meant to be re-run across a growing window all
campaign long needs to move.

**3 — Hardcoded values.** Every behaviour-deciding literal is a module-top
constant naming its unit and citing the method section it implements:
`IDLE_GAP_SECONDS`, `MIN_GROUP_N`, `MIN_RELATIVE_CHANGE`, `MAX_UNPLACED_SHARE`,
`CI_RUN_LIMIT`, `NOTE_KINDS`, `LEDGER`, `DEFAULT_REGISTRY`, `METHOD`, and the
three `gh` field lists.

The four thresholds are deliberately **not** user-tunable, and that is the right
tier rather than a missed one: the method states that moving one invalidates
every reading taken under the old value, so a `--min-group-n` flag would let a
future run tune its way to a result. `dispersion.py`'s docstring says so. What a
user does vary — which missions, which sessions, which window — lives in
`docs/quality/mission-cost/missions.json`, which is data, not code. No finding.

**4 — Tests.** 110, offline, against the committed synthetic fixture or a fake
`gh` runner. Written before the implementation: `2f4dfeb6` precedes `b1890d1b`.
Every new behaviour has a test that fails without it, each repair carries one,
and expectations are hand-computed from the fixture rather than read out of the
code under test.

The load-bearing half is negative. Bait strings planted in the fixture —
`message.content`, `cwd`, `gitBranch`, `service_tier` — must not survive, and
after `26f7fdf5` that test got *stronger*: there are no retained records left to
inspect, so it reads everything the transcript object still holds and asserts no
bait is anywhere in it. A reader widened by one field fails in CI rather than in
a published aggregate.

Edge cases the domain produces: a missing counter, an unparsable line, a usage
object with no timestamp, a line out of order, a transcript opened and
abandoned, a session directory with no root transcript, a registry with a
duplicate branch or a session declared twice, a mission with no window.

One gap, named and not filed: the `gh` path is proved by a fake runner and by a
manual smoke run against merged PR #547, not by a test that calls GitHub. That
is the correct trade — a test that calls GitHub is not a test — and the smoke
run is recorded in the PR body.

**5 — Standardisation.** Mirrors `tools/read_cost/`: a directory per instrument,
`measure.py` as the entry point, `test_*.py` beside it, `importlib` sibling
loading, canonical JSON out, an injectable runner for anything that shells out,
and the read-cost ledger read through `tools/read_cost/ledger.py`'s own parser
rather than a second one. `dispersion.py` is named that way precisely because
`unittest discover` puts the directory on `sys.path` and a `statistics.py` there
would shadow the stdlib module it uses.

**6 — Human-friendly.** Names carry their unit (`agent_seconds`,
`cache_read_input_tokens`, `IDLE_GAP_SECONDS`). Every non-obvious rule carries a
comment saying *why*: why a session is never split, why subagent time is counted
twice in one number and once in the other, why the exact fix for the weakest
attribution rule is deliberately not used, why the verdict chain is closed. No
dead code after the `span_seconds` and `now()` removals; nothing commented out.

**7 — The second operator.** **Act:** one documented command with `--help`,
taking data, with no UI anywhere near it. **Read:** canonical JSON with sorted
keys and a `schema` field — every number is a field, none of it pixels, and
after `26f7fdf5` even the failure notes are typed objects an agent can branch
on. **Discover:** `docs/README.md` indexes the method, the package README
documents the command, and the report names the method document it implements.
No authority surface: it reads, it writes nothing to the repository, it reaches
no market or safety action.

**8 — Language.** Two claims, separately. *The guard:*
`crates/guards/src/language.rs` runs inside `cargo test --workspace`, green at
this head. *Read by hand:* the branch name, all seven commit messages, the PR
title and body, 266 lines of `method.md`, two READMEs, and every identifier,
docstring, comment, test name and fixture string in the package. All English.
The fixture's invented strings are English. No exemption is claimed in this
diff, and none is needed.

**9 — The trunk.** **Trunk flat.** No `crates/` file moved, so the size ratchet
and `size-baseline.txt` are untouched; `tools/` and `docs/quality/` are outside
the context ratchet's roots, so no ceiling was consulted or raised. Nothing
landed on `QuantickApp` or any application root struct. The capability arrived
as a new directory plus one line in `docs/README.md` and one CI step — the shape
the trunk rule asks for.

The nine added lines in `ci.yml` get their own sentence, because this mission's
scope says the development loop is #567's work. The step runs the new offline
suite and mirrors the `read-cost` and `edit-loop` steps beside it. It is the
ordinary obligation of adding tested code — a suite CI never runs is not tested
— and it changes no existing step, tunes nothing and removes no gate. In scope,
and declared as an exception in the PR body.

No new execution artifacts and no `GOAL-archive-*` in the diff. The fixture is a
reusable test fixture, which the rule exempts explicitly.

## In passing, for `delivery-review`, with no severity

`method.md` was amended twice after its own first commit, both inside this pull
request and both before any reading exists: the work window's start now comes
from GitHub's commit list, because a merged branch's merge base collapses onto
its own head; and CI time is summed over the branch's runs rather than one
head's. Neither moved a threshold. Whether that satisfies criterion one belongs
in the delivery grade, not here.

## Validation, verified rather than taken on trust

Each run on its own at this head, exit codes captured separately — no chained
all-clear:

```
fmt=0        cargo fmt --all -- --check
clippy=0     cargo clippy --workspace --all-targets     (0 warnings)
build=0      cargo build --workspace
test=0       cargo test --workspace                     (no failing result line)
```

```
ruff check --select F tools/ bridge/mt5/        All checks passed!
python -m unittest discover -s tools/mission_cost -p 'test_*.py'
                                                Ran 110 tests ... OK
```

## Verdict

- **Correctness** — step 0 ran twice at `low`. Four findings in round one, all
  confirmed, all fixed in `6052b1de`, plus two more the repairs' tests found
  including a real determinism bug in the output path. Round two over the repairs returned three more, including a Blocker the repairs themselves introduced — a half-resolved window that was unbounded on the left — all fixed in `5eebc333`. Nothing open.
- **Docking** — yes. A new consumer of the reader attaches without opening a
  file; a fourth external source would mean editing `build_report`, which is the
  shape both neighbouring tools already use.
- **Performance** — flat, and better where it moved. No Rust path touched, rate
  is *rare*, 3.7 s over 652 real transcripts, guards report unchanged; retained
  memory went from O(requests) to O(work blocks).
- **Operability** — act, read and discover are all present: one documented
  command, canonical JSON, typed failure notes, indexed in `docs/README.md`. No
  authority surface.
- **Proof** — `test_transcripts.FieldFilter.test_nothing_a_transcript_retains_carries_a_bait_string`
  is the one that fails if the privacy boundary regresses, and
  `test_measure.Determinism.test_out_writes_the_same_bytes_it_would_have_printed`
  already caught the CRLF bug. Both are integration tests over the real read and
  the real command; the unit level is `record_from`'s own cases.
- **Accumulation** — trunk flat. No tracked file moved, no ceiling raised, no
  baseline consulted.
- **Language** — the guard passed, and I read the prose, the branch name and the
  commit messages myself. All English.

ARCH-REVIEW: PASS

milocaetano:
## AI review - PR #569 @ 5eebc3339134ebabf90f2bb4b0f1ab5b64f947ce

**Identity**
- HEAD: `5eebc3339134ebabf90f2bb4b0f1ab5b64f947ce`
- Branch: `feat/mission-cost-harness`
- Worktree: `C:\src\quantick-worktrees\feat-mission-cost-harness` (clean)
- Base ref: `origin/campaign/mission-velocity` (campaign child of #563, task T1)
- Base tip: `a4ba1cbc9971bef1a74944256b96346a2e920030`
- Review key: `5e55c709b289d98d737d1c92972da97ee8603afc`
- PR identity: #569, head `feat/mission-cost-harness`, base `campaign/mission-velocity`

**Scope** — full diff, round one, plus a narrow follow-up. Reviewed at
`8da3927b`; three findings published as resolvable threads; all three fixed in
`26f7fdf5`. The architecture review's second bug pass then found a Blocker in
those repairs, fixed in `5eebc333`, so the full diff was re-read at this head
for new FAILs against these six questions. None found, and no new WEAK is opened
against code already passed. Finding count 3, unresolved 0, accepted 0, fixed 3.

| ID | Thread | Dimension | Disposition |
| --- | --- | --- | --- |
| F1 | `PRRT_kwDOTfuoRs6kNnWA` | 3, AI-ready | fixed, resolved |
| F2 | `PRRT_kwDOTfuoRs6kNna8` | 2, decoupled | fixed, resolved |
| F3 | `PRRT_kwDOTfuoRs6kNnfn` | 6, scalable | fixed, resolved |

`ai_review_threads.sh count 569` returns `0`.

---

**1. Modular: PASS.** One responsibility per file and the name says which:
`transcripts.py` reads, `attribution.py` places, `dispersion.py` grades,
`delivery.py` asks the forge, `measure.py` assembles. Inside `transcripts.py`
the split went one level finer during repair —
`transcripts.py:78` `record_from` is the filter as a named function and
`transcripts.py:118` `Transcript.add` is the fold, so the privacy rule and the
wall-clock rule are two units instead of one read loop.

Blast radius: three pre-existing files edited, none of them behaviour.
`.github/workflows/ci.yml` has the most edited lines at **+9, 0 changed** — one
step appended beside the `read-cost` and `edit-loop` steps. Then
`docs/README.md` +1 and `docs/quality/read-cost/ledger.md` +1. Nothing existing
was modified anywhere in the diff.

**2. Decoupled: PASS** (was WEAK, F2). Every call reaching GitHub took its own
inline argument vector, so a field `gh pr view` does not expose meant editing
three function bodies. `delivery.py:32-51` now builds all three —
`pull_command`, `commits_command`, `runs_command` — beside the field lists they
use, and `test_measure.Commands` pins the shapes. The `runner` parameter remains
the injection seam and the tests answer from a table.

No consumer names a concrete producer where a bound would do: `attribution` and
`measure` take lists and dicts, and `Placement`, `Pulls`, `Session` and
`Transcript` each own their own state rather than having it passed in and
mutated by the caller. A `Forge` class is deliberately absent — one
implementation, and `tools/read_cost/ledger.py:179` `gh_json` is the same shape.

**3. AI-ready: PASS** (was WEAK, F1). `measure.py:127` `NOTE_KINDS` is a closed,
documented set and `measure.py:130` `_note` builds `{branch, kind, detail}`, so
a failed `gh` lookup inside a *successful* report is something a consumer can
branch on rather than prose it has to match. That matters concretely: #565 must
exclude missions whose window never resolved, because one of those reports
near-zero cost and drags a group median down.

The behaviour is a named call with a typed result: canonical JSON carrying
`schema`, `method` and the registered `thresholds`, with `--help` describing
every flag. Typed errors — `attribution.RegistryError`, `delivery.DeliveryError`
— rather than a `String` standing in for a failure; the CLI boundary collapses
them into an exit code and a message, which is the contract every other tool in
`tools/` uses. Calling it twice is safe and proved so: `test_measure.Determinism`
asserts byte-identical output across two runs and across both output channels.
No click handler exists anywhere; there is no UI to have one.

**4. Agent-tested: PASS.** 110 tests, all offline.
`python tools/mission_cost/test_transcripts.py FieldFilter.test_nothing_a_transcript_retains_carries_a_bait_string`
is the one that fails without the diff's privacy boundary, and it runs alone. It
is below `app` in the sense that matters here: nothing in the package imports a
crate, a runtime or a network.

Error-path fixtures: an unparsable line, a usage object with no timestamp, a
missing counter, a line out of order, an abandoned transcript, a registry with a
duplicate branch, a session declared twice, an unknown schema, an unknown group,
a mission with no window. Expectations are hand-computed from the fixture — the
460/230/1260/220 in `test_attribution.py` was added up by hand from the fixture
lines, not read out of the code under test.

Non-determinism hunt, and this one has teeth because the harness's whole job is
to produce comparable numbers: no `rand`; `os.walk` results are explicitly
sorted; every set is rendered through `sorted()`; the digest hashes sorted
lines; JSON is emitted with `sort_keys=True`. The package's only wall-clock read
was `delivery.now()`, which nothing called — removed in `26f7fdf5`, so there is
now no clock in it at all. The one environment read is `os.path.expanduser("~")`
in `default_transcripts`, and it cannot reach the output because `inputs.roots`
carries basenames only.

**5. Extensible: PASS.** `5eebc333` added a third note kind to `NOTE_KINDS`
by appending to the tuple and nothing else, which is the shape this dimension
asks for: a registry keyed by name, not by position. A new metric needs no code
at all:
`measure.py:metric_value` resolves by name with a dotted convention, so anything
added to `total`, `main_thread` or `subagents` is selectable the moment it
exists. The registry is keyed by branch name, so a new mission appends. A new
transcript layout is one branch in `transcripts._classify`.

`dispersion.compare`'s five-way verdict is a closed chain, and that is correct
rather than a finding: the verdicts *are* the domain, fixed by the registered
method, and a sixth one must be a method change with its own pull request —
exactly what a registry keyed for easy extension would quietly permit.

The one thing that is patched rather than added is a fourth external source,
which would mean editing `build_report`. Naming that variant honestly: nothing
in #563's decomposition needs one, and inventing a source registry for a
hand-run local script would be a standardisation finding against the two
neighbouring tools. PASS, with the variant named.

**6. Scalable: PASS** (was WEAK, F3). Rate: **rare** — a hand-run batch script,
no per-trade, per-depth or per-frame path touched anywhere, and no Rust changed.

Per-input cost is now independent of history, which is what the finding asked
for. `Transcript` retains no records: `read` folds each line into four counters,
two instants and a merged span list and forgets it, so the retained state is
proportional to work blocks rather than to every request ever made in a projects
directory nobody prunes. That is the difference between a harness that gets
heavier every week of the campaign and one that does not.

Buffers with a stated cap: `CI_RUN_LIMIT = 300` on the workflow listing, with
`ci_listing_truncated` in the report so a branch with more runs announces that
its CI time is a floor rather than reporting low in silence. Nothing is dropped
quietly — unparsed lines, undated usage lines, out-of-order lines and skipped
paths are all counted into `inputs`.

Measured at this head: 652 real transcripts, 49 sessions, 3.7 s, 0 unparsed, 0
undated, 0 out of order, byte-identical across two runs and both channels
(`5861821c…`).

---

**Top fix:** none outstanding. The three published findings are fixed and
resolved; F3 was the one that flipped a verdict on its own and it flipped
`6. Scalable` from WEAK to PASS.

AI-REVIEW: COMPLETE

milocaetano:
# Delivery review — PR #569 @ `5eebc3339134ebabf90f2bb4b0f1ab5b64f947ce`

**Mode: completeness-only**, which is what tier `medium` buys. This marker is
**not** the one a full review records: no independent criteria reviewer was
dispatched, no `A` or `G` line was graded against the diff by a fresh context,
and the criteria pass did not run. What follows is the completeness pass alone,
run inline, exactly as the skill's `medium` row prescribes.

**Identity.** Branch `feat/mission-cost-harness`, worktree
`C:\src\quantick-worktrees\feat-mission-cost-harness` (clean, `HEAD` unmoved
across the pass), base `origin/campaign/mission-velocity`, base tip `a4ba1cbc`,
review key `5e55c709b289d98d737d1c92972da97ee8603afc`. Campaign child of #563,
task T1.

## Step 1 — the checklist source

**Source 1**, the strong one: the ignored local `.claude/GOAL.md` in the task
worktree, in `mission`'s documented format — ledger `R1`…`R16`, decisions
`D1`…`D5`, assumptions `S1`…`S4`, criteria `A1`…`A10` and `G1`…`G6` plus the
four reserved `G-AI` lines, closing steps `C1`…`C4`, and the full verbatim
request as an attributed quotation. The PR body carries the same goal in
`<details>`.

The verbatim request section exists, so the completeness pass can run. The
retained request is the campaign coordinator's dispatch brief, which names
issue #564 as its task; both were read as the source.

## Completeness pass — the request against the map

Twenty-two distinct asks were derived from the source independently, before the
map was opened: nine from #564 (its scope in, its scope out, and its seven
acceptance criteria) and thirteen from the dispatch brief.

**Five were `UNLEDGERED` on the first reading.** Four were traceability gaps —
the outcome had shipped and nothing on the map pointed at it — and one was a
real uncovered ask with no outcome at all.

| # | Source span | Gap | Kind | Resolution |
| U1 | brief, SCOPE DISCIPLINE — "If you find something tempting, write it in a comment on #563 and move on." | no `R`, `G` or `C` covered it, and no comment existed | **outcome** | Discharged: [#563 comment 5754311754](https://github.com/milocaetano/quantick/issues/563#issuecomment-5754311754). Mapped as `R13`/`G6`. |
| U2 | brief, BASE AND WORKTREE — the `mission-base` record, the campaign base, never the main checkout | discharged, unmapped | traceability | `R14`/`G4` |
| U3 | brief — "follow the `new-task` … skills as that mission requires" | partly discharged: the branch prefix was right, **the board card was still on `Todo`** | **outcome** | Card moved to `In Progress` on both the Status and Campaign state fields and read back. Mapped as `R15`/`G5`. |
| U4 | brief, WHEN DONE — "Do not mark the issue done." | discharged, unmapped | traceability | folded into `R12`/`C4` |
| U5 | brief, GROUNDING — "Do not treat all four token kinds as interchangeable — report them separately" | delivered, unmapped | traceability | `R16`/`A10` |

U1 and U3 are why this pass exists. Both were *live* gaps, not bookkeeping: an
instruction to file out-of-scope observations rather than act on them had been
followed in spirit — nothing out of scope was built — but never discharged, and
the board still said this issue had not been started. Neither the architecture
review nor the AI review can see either one; they grade the diff, and neither
gap is in the diff.

A note on U1's resolution, since a comment posted to close a review finding is
the kind of thing that can be hollow: the comment carries four properties of the
instrument that #565 and #566 will otherwise rediscover the hard way — that
window-only attribution placed 77% of one merged pull request's window into the
unplaced buckets, that a session started in a worktree has its own projects
directory, that some sessions have no main-thread transcript, and that a ranking
which sums the four token kinds will mislead. It draws no baseline conclusion
and grades none of the seven merged optimizations, which is the scope line this
mission was told to hold.

**After the map corrections: `UNLEDGERED` = 0.** Every distinct outcome and
constraint in the retained request now maps to at least one `R`, and every
operational obligation to a `G` or `C` carrying its source and its evidence. No
number was renumbered; `R13`…`R16`, `A10` and `G4`…`G6` were appended, and the
correction itself is recorded in `GOAL.md` under `## Map corrections` so that a
corrected map does not read as one that was always right.

The map correction touched only the ignored `GOAL.md`, so the review key is
unchanged at `5e55c709b289d98d737d1c92972da97ee8603afc` and the architecture and
AI-review projections recorded against it remain current. `git status
--porcelain` reports no tracked change.

## What this pass did not do

Named plainly, because a `medium` marker that reads like a `high` one is worse
than no marker:

- No dossier was assembled and no independent reviewer was dispatched.
- No `A` or `G` line was graded `DELIVERED` / `PARTIAL` / `MISSING` /
  `UNPROVEN` against the diff by a fresh context.
- No `R` received a `COVERED` / `PARTLY COVERED` / `DROPPED` grade.

The evidence each criterion names is in the PR body and in the two published
review reports; what has *not* happened is an independent reading of that
evidence by a context that did not write it.

## Two things this pass noticed and is not grading

Passed to the coordinator rather than scored, because they belong to a criteria
pass this tier does not buy:

1. **`A2` is the mission's whole point and it is the one a full review should
   read hardest.** The claim is that the method commit precedes any commit
   reporting a reading. `git log --reverse` shows `01bec148` first and alone,
   and no commit on the branch carries a number read from the transcripts. But
   `method.md` was amended twice afterwards, inside `b1890d1b` — the window's
   start now comes from GitHub's commit list, and CI time is summed over the
   branch rather than one head. Both landed before any reading exists and
   neither moved a threshold, which is what the criterion protects. A full
   criteria pass should confirm that reading rather than take it from here.
2. **`S1` is an assumption that drove the design**, and rule 7 says such a thing
   is reported as a question the interrogation should have asked. It was not
   asked because it could not be: step 3 was discharged by the coordinator
   before dispatch and a subagent cannot ask the trader. The assumption — that
   #564's second acceptance criterion is literal, so `cwd` and `gitBranch` are
   not read even though they would make the attribution exact — is recorded in
   `GOAL.md` marked *wanted to ask*, and its cost is written into the method as
   error mode E2 rather than quietly absorbed. That is the right handling of an
   unaskable question, but the trader should know the call was made.

## Verdict

Checklist source: **1** (the local `GOAL.md`, documented format, verbatim
request present). The completeness pass **could** run and **did** run.

- Distinct asks derived from the source: **22**
- `UNLEDGERED` on first reading: **5** — two with outcome gaps, three
  traceability only
- `UNLEDGERED` after resolution: **0**
- Deferrals: **none**, granted or requested

Nothing is `UNLEDGERED`. Under the `medium` row — step 1, then the completeness
pass alone — that is the pass condition, and the mode is stated above so this
marker cannot be mistaken for a full review's.

DELIVERY-REVIEW: PASS

milocaetano:
## Mission completion - PR #569 @ 5eebc3339134ebabf90f2bb4b0f1ab5b64f947ce

Caller: mission
PR: https://github.com/milocaetano/quantick/pull/569
Branch: feat/mission-cost-harness
Base: origin/campaign/mission-velocity
Base tip: a4ba1cbc9971bef1a74944256b96346a2e920030
Review key: 5e55c709b289d98d737d1c92972da97ee8603afc

Mission source stayed local; the PR carries its concise durable summary.

- [x] - **D1** — The open PR is non-draft and matches branch, head and base.
  Evidence: open non-draft PR https://github.com/milocaetano/quantick/pull/569 matches branch feat/mission-cost-harness at 5eebc3339134ebabf90f2bb4b0f1ab5b64f947ce and base origin/campaign/mission-velocity
- [x] - **D2** — Every registered CI check is green at that head.
  Evidence: full ci and windows passed at the exact head; every other registered check that ran is pass
- [x] - **D3** — Architecture has a current projection and durable PASS report.
  Evidence: https://github.com/milocaetano/quantick/pull/569#issuecomment-5754297847
- [x] - **D4** — Delivery has both when applicable; only bounded `small` is exempt.
  Evidence: https://github.com/milocaetano/quantick/pull/569#issuecomment-5754322996
- [x] - **D5** — AI has current completion, a durable report and zero listed threads.
  Evidence: https://github.com/milocaetano/quantick/pull/569#issuecomment-5754299089; ai_review_threads.sh list returned zero
- [x] - **D6** — The PR body carries the concise mission summary, current-head evidence and report URLs.
  Evidence: the PR body contains the concise mission summary and current quantick-delivery-evidence:v1 block
- [x] - **D7** — The final verifier publishes and verifies this literal reconciliation.
  Evidence: this literal report is published and verified before the command returns PASS
- [x] - **D8** — Only the user merges to `main`.
  Evidence: this verifier performs no merge; main remains a human action

MISSION-COMPLETION: PASS

